PR #34, opened 2026-07-27
Quit apps from the launcher: per-app Quit Application and a Quit All Applications command

### LINKED ISSUE #32: [Feature Request] Quitting apps
Something i often use in Alfred is the option to quit certain open apps + the option to quit all apps

suiramdev:
## Summary

A running app can now be quit straight from the launcher instead of switching to it and pressing ⌘Q, and the whole desktop can be cleared with one command — the two things the issue asks for.

- **Quit Application** — the last row of an app's ⌘K Actions menu, and **⌘↵** on the selected row. It only appears while that app is actually running (same `RunningAppsMonitor` that drives the row's running dot), so it is never a dead action, and the palette dismisses only when something really was running.
- **Quit All Applications** — a new palette command. It targets every Dock-level (`.regular`) app except Finder and Tinycast itself; background agents, accessories and helpers are left alone.

Both are graceful `NSRunningApplication.terminate()`, so an app with unsaved work still puts up its own save sheet — nothing is force-killed.

Three decisions worth a look:

- Finder is excluded from Quit All because `terminate()` only makes it relaunch.
- Tinycast excludes itself by PID rather than by activation policy — About/Settings temporarily flips the app to `.regular`, so a policy-only filter could make it quit itself.
- ⌘↵ was unused in launcher mode and is the "secondary action" slot the other modes already use (clipboard copy, history copy-expression); the menu row advertises it.

Fixes #32

## Validation

- Compiled the real `AppLauncher` into a standalone harness (the repo's test idiom): `quit(bundleID:)` terminated a live TextEdit and returned `true`, and returned `false` both before the launch and after the termination. `quitAllTargets()` on a working desktop selected 10 of 162 running processes — Finder excluded, self excluded, only Dock-level apps.
- Drove the built Debug app: with TextEdit running, ⌘K on its row showed **Quit Application** (destructive red, `⌘↵` keycaps); activating it quit TextEdit and dismissed the palette. The row is absent for apps that are not running.
- `xcodebuild -project Tinycast.xcodeproj -scheme Tinycast -configuration Debug` builds clean.
- `quitAll()` itself was deliberately not executed — it would have terminated the apps hosting the session. What was verified is the target list it iterates; the executor is a `terminate()` loop over exactly that list.

## Notes

Claude Code generated it.

### DIFF
--- a/Tinycast/Core/AppCore.swift
+++ b/Tinycast/Core/AppCore.swift
@@ -261,6 +261,14 @@ final class AppCore: ObservableObject {
         }
     }
 
+    /// Quits the app behind an entry; a no-op (palette stays put) when it isn't running.
+    func quit(_ app: AppEntry) {
+        guard app.kind == .application, let bundleID = app.bundleID,
+            AppLauncher.quit(bundleID: bundleID)
+        else { return }
+        hidePalette(restoreFocus: false)
+    }
+
     private func runCommand(_ entry: AppEntry) {
         switch CommandRegistry.command(for: entry) {
         case .calculatorHistory:
@@ -284,6 +292,9 @@ final class AppCore: ObservableObject {
         case .about:
             hidePalette(restoreFocus: false)
             showAbout()
+        case .quitAllApps:
+            hidePalette(restoreFocus: false)
+            AppLauncher.quitAll()
         case .quit:
             NSApp.terminate(nil)
         case nil:
--- a/Tinycast/Core/AppLauncher.swift
+++ b/Tinycast/Core/AppLauncher.swift
@@ -40,4 +40,32 @@ enum AppLauncher {
             running.activate()
         }
     }
+
+    /// Asks every running instance of a bundle ID to quit — graceful, so an app with unsaved work still gets to put its own sheet up. False when nothing was running.
+    @MainActor
+    @discardableResult
+    static func quit(bundleID: String) -> Bool {
+        let running = NSRunningApplication.runningApplications(withBundleIdentifier: bundleID)
+        for app in running { app.terminate() }
+        return !running.isEmpty
+    }
+
+    /// Finder is never a Quit All target: `terminate()` only makes it relaunch, and nobody means the desktop when they say "quit everything".
+    private static let quitAllExclusions: Set<String> = ["com.apple.finder"]
+
+    /// What Quit All acts on: every app with a Dock presence, minus Finder and Tinycast itself. Accessories and background agents are left alone.
+    @MainActor
+    static func quitAllTargets() -> [NSRunningApplication] {
+        let ownPID = NSRunningApplication.current.processIdentifier
+        return NSWorkspace.shared.runningApplications.filter { app in
+            app.activationPolicy == .regular
+                && app.processIdentifier != ownPID
+                && !quitAllExclusions.contains(app.bundleIdentifier ?? "")
+        }
+    }
+
+    @MainActor
+    static func quitAll() {
+        for app in quitAllTargets() { app.terminate() }
+    }
 }
--- a/Tinycast/Core/RunningApps.swift
+++ b/Tinycast/Core/RunningApps.swift
@@ -21,6 +21,12 @@ final class RunningAppsMonitor: ObservableObject {
         }
     }
 
+    /// True when the entry's bundle is currently running — drives the row's running dot and the Quit action.
+    func isRunning(_ app: AppEntry) -> Bool {
+        guard let bundleID = app.bundleID else { return false }
+        return runningBundleIDs.contains(bundleID)
+    }
+
     private func refresh() {
         runningBundleIDs = Set(
             NSWorkspace.shared.runningApplications.compactMap(\.bundleIdentifier))
--- a/Tinycast/Features/RootPaletteView.swift
+++ b/Tinycast/Features/RootPaletteView.swift
@@ -10,6 +10,8 @@ struct RootPaletteView: View {
     @EnvironmentObject private var calcHistory: CalculatorHistoryStore
     @EnvironmentObject private var emojiIndex: EmojiIndex
     @EnvironmentObject private var frequentEmoji: FrequentEmojiStore
+    /// Observed so an app launching / quitting re-renders the Quit action's availability.
+    @EnvironmentObject private var runningApps: RunningAppsMonitor
     /// Observed so a skin tone changed in Settings re-renders the grid glyphs immediately.
     @ObservedObject private var settings = AppCore.shared.settings
     @FocusState private var searchFocused: Bool
@@ -105,7 +107,9 @@ struct RootPaletteView: View {
                 return CalcActionsMenu.content(result: calc, core: core)
             }
             if let app = selectedAppEntry {
-                return AppActionsMenu.content(app: app, core: core, favorites: favorites)
+                return AppActionsMenu.content(
+                    app: app, core: core, favorites: favorites,
+                    running: runningApps.isRunning(app))
             }
             return nil
         case .clipboard:
@@ -344,7 +348,11 @@ struct RootPaletteView: View {
                 guard command, histResults.indices.contains(index) else { return .ignored }
                 core.copyHistoryExpression(histResults[index])
             case .launcher:
-                return .ignored
+                // ⌘↵ quits the selected app when it's running (the Actions menu advertises it); nothing else in the launcher takes a modified ↵.
+                guard command, let app = selectedAppEntry, runningApps.isRunning(app) else {
+                    return .ignored
+                }
+                core.quit(app)
             }
             return .handled
         }
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ CPU churn. Just SwiftUI + AppKit with zero dependencies. It's fast because there
 
 ## Features
 
-- **App launcher** — fuzzy-search and launch anything, pin favorites, see what's running.
+- **App launcher** — fuzzy-search and launch anything, pin favorites, see what's running, quit an app
+  or every app at once.
 - **Calculator** — do math and unit conversions inline, right in the palette.
 - **Clipboard history** — text and images, searchable, pasted back into the app you were using.
 - **Global hotkey** — one shortcut summons the palette from anywhere.
--- a/Tinycast/Core/CommandRegistry.swift
+++ b/Tinycast/Core/CommandRegistry.swift
@@ -10,6 +10,7 @@ enum CommandID: String, CaseIterable, Sendable {
     case importFromRaycast = "command:import-from-raycast"
     case settings = "command:settings"
     case about = "command:about"
+    case quitAllApps = "command:quit-all-apps"
     case quit = "command:quit"
 
     var name: String {
@@ -22,6 +23,7 @@ enum CommandID: String, CaseIterable, Sendable {
         case .importFromRaycast: return "Import from Raycast"
         case .settings: return "Settings"
         case .about: return "About Tinycast"
+        case .quitAllApps: return "Quit All Applications"
         case .quit: return "Quit Tinycast"
         }
     }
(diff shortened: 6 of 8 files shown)


abue-ammar (commented):
Reviewed against the branch, built it, and probed the runtime behaviour locally.

## Verification

- `xcodebuild -project Tinycast.xcodeproj -scheme Tinycast -configuration Debug` on a clean worktree of `feat/quit-action` — **BUILD SUCCEEDED**.
- Compiled the `quitAllTargets()` predicate standalone against this machine's live process list: **3 targets out of 77 running apps** (VS Code, Helium, System Settings). Finder excluded, self excluded, and **zero `.regular` apps with a nil bundle ID** — so the `?? ""` fallback in the exclusion check never misfires.
- Traced the fuzzy ranking for the `quit` collision (see the inline note on `AppCore.runCommand`).

## What's right

- Everything lands where it belongs: dispatch on `AppCore`, a `CommandID` case, an `AppLauncher` static. No new singletons, no new long-lived state, `docs/launcher.md` updated alongside.
- Excluding Tinycast **by PID rather than activation policy** is the correct call, and the comment says why — About/Settings flips the app to `.regular`. Easy thing to get wrong.
- The Quit row is **appended last**, so when it appears or disappears the other rows keep their indices and `activateMenuItem`'s bounds check holds. No misfire.
- Folding the bundle-ID lookup into `RunningAppsMonitor.isRunning(_:)` is a genuine cleanup of the call site it replaces.

## Summary of findings

Six inline comments below. The two I'd fix before merge are the **menu mutating while open** (`RootPaletteView`) and the **⌘↵ kind guard drifting from the menu's** — both are on `RootPaletteView.swift`. The equality guard in `RunningAppsMonitor.refresh()` is a one-line win worth taking while you're in the file. The rest are judgment calls and nits.

Nothing here blocks merge.

Tinycast/Features/RootPaletteView.swift
```diff
@@ -105,7 +107,9 @@ struct RootPaletteView: View {
                 return CalcActionsMenu.content(result: calc, core: core)
             }
             if let app = selectedAppEntry {
-                return AppActionsMenu.content(app: app, core: core, favorites: favorites)
+                return AppActionsMenu.content(
+                    app: app, core: core, favorites: favorites,
+                    running: runningApps.isRunning(app))
```
abue-ammar, line 112: **The open ⌘K menu can now mutate underneath the user.**

Because `RootPaletteView` observes `RunningAppsMonitor`, an app launching or quitting *externally* re-renders `actionsContent` while the menu is up: the Quit row appears or disappears, the popover resizes under the cursor, and a highlight parked on that row silently becomes a no-op.

It isn't a crash — the row is appended last so the other indices are stable, and `activateMenuItem` bounds-checks. But [`docs/palette.md:25`](https://github.com/abue-ammar/tinycast/blob/main/docs/palette.md#L25) is explicit that the palette freezes while a footer menu is open, and this is the one thing in the menu that can now change on its own.

Suggest snapshotting the running state into `@State` when `showActions` flips true, and reading the snapshot here, rather than reading the monitor live.

Tinycast/Features/RootPaletteView.swift
```diff
@@ -10,6 +10,8 @@ struct RootPaletteView: View {
     @EnvironmentObject private var calcHistory: CalculatorHistoryStore
     @EnvironmentObject private var emojiIndex: EmojiIndex
     @EnvironmentObject private var frequentEmoji: FrequentEmojiStore
+    /// Observed so an app launching / quitting re-renders the Quit action's availability.
+    @EnvironmentObject private var runningApps: RunningAppsMonitor
```
abue-ammar, line 14: **Invalidation scope widens from `LauncherList` to the whole root view.**

Every workspace launch/terminate now re-runs the entire `RootPaletteView` body, including `appResults` → `favorites.ordered(...)` — a partition plus sort that, unlike `appIndex.matches`, is not memoized.

In practice the palette is usually hidden when apps launch, so the real cost is small. But the fix is cheap: move this `@EnvironmentObject` into a small wrapper view around the bottom-trailing overlay, so only that subview invalidates. Pairs with the equality guard suggested on `RunningApps.swift`.

Tinycast/Features/RootPaletteView.swift
```diff
@@ -344,7 +348,11 @@ struct RootPaletteView: View {
                 guard command, histResults.indices.contains(index) else { return .ignored }
                 core.copyHistoryExpression(histResults[index])
             case .launcher:
-                return .ignored
+                // ⌘↵ quits the selected app when it's running (the Actions menu advertises it); nothing else in the launcher takes a modified ↵.
+                guard command, let app = selectedAppEntry, runningApps.isRunning(app) else {
+                    return .ignored
+                }
+                core.quit(app)
```
abue-ammar, line 355: **This guard has drifted from the menu's.**

Here the condition is `runningApps.isRunning(app)`; in `AppActionsMenu.content` it's `running, app.kind == .application`. `core.quit` re-guards the kind, so nothing bad happens — but a running non-`.application` entry (a `.systemSettings` pane whose bundle ID happens to be live) returns `.handled` and does nothing, swallowing the key.

Mirror the menu's condition exactly so the two paths can't drift further:

```swift
guard command, let app = selectedAppEntry, app.kind == .application,
    runningApps.isRunning(app)
else { return .ignored }
```

Tinycast/Core/RunningApps.swift
```diff
@@ -21,6 +21,12 @@ final class RunningAppsMonitor: ObservableObject {
         }
     }
 
+    /// True when the entry's bundle is currently running — drives the row's running dot and the Quit action.
+    func isRunning(_ app: AppEntry) -> Bool {
+        guard let bundleID = app.bundleID else { return false }
+        return runningBundleIDs.contains(bundleID)
+    }
+
     private func refresh() {
```
abue-ammar, line 30: **Unconditional republish, now that more of the tree observes this.**

`runningBundleIDs` is assigned on every launch/terminate notification whether or not the set actually changed, so `@Published` fires regardless. That was cheap when only `LauncherList` observed it; it's less cheap now that `RootPaletteView` does too.

One line:

```swift
private func refresh() {
    let next = Set(NSWorkspace.shared.runningApplications.compactMap(\.bundleIdentifier))
    guard next != runningBundleIDs else { return }
    runningBundleIDs = next
}
```

Worth doing independently of the rest of this PR.

Tinycast/Core/AppCore.swift
```diff
@@ -284,6 +292,9 @@ final class AppCore: ObservableObject {
         case .about:
             hidePalette(restoreFocus: false)
             showAbout()
+        case .quitAllApps:
+            hidePalette(restoreFocus: false)
```
abue-ammar, line 296: **Quit All is one ↵ away, with no confirmation.**

I checked the ranking, and the common case is safe: `FuzzyMatch`'s prefix tier is `90_000 - name.count`, which puts "Quit Tinycast" (89 987) above "Quit All Applications" (89 979) for the query `quit`. Good.

But `quit a`, `all`, or empty-query alphabetical browsing all land straight on it, and it sits one row from the command a user hits by muscle memory. It's the only action in the app whose blast radius reaches outside Tinycast.

Graceful `terminate()` makes this recoverable rather than destructive, so this is a judgment call, not a bug. Still: I'd want either a confirm step, or the same deliberate-gesture treatment (⌘↵) the per-app quit gets, rather than plain ↵.

Tinycast/Core/AppCore.swift
```diff
@@ -261,6 +261,14 @@ final class AppCore: ObservableObject {
         }
     }
 
+    /// Quits the app behind an entry; a no-op (palette stays put) when it isn't running.
+    func quit(_ app: AppEntry) {
+        guard app.kind == .application, let bundleID = app.bundleID,
+            AppLauncher.quit(bundleID: bundleID)
+        else { return }
+        hidePalette(restoreFocus: false)
+    }
```
abue-ammar, line 270: **`restoreFocus: false` after a quit — worth a second thought.**

Consistent with every other command here, so defensible. But the semantics differ: after `launch`, the launched app takes focus, so there's nothing to restore. After quitting a *background* app, the user is dropped wherever macOS decides to land them, rather than back in the app they were actually working in.

`restoreFocus: true` is the more accurate behaviour — except when the app being quit *is* `previousApp`, where re-fronting a terminating app would be wrong. Not a blocker either way.

Tinycast/Core/AppLauncher.swift
```diff
@@ -40,4 +40,32 @@ enum AppLauncher {
...
+        for app in running { app.terminate() }
+        return !running.isEmpty
+    }
+
+    /// Finder is never a Quit All target: `terminate()` only makes it relaunch, and nobody means the desktop when they say "quit everything".
+    private static let quitAllExclusions: Set<String> = ["com.apple.finder"]
+
+    /// What Quit All acts on: every app with a Dock presence, minus Finder and Tinycast itself. Accessories and background agents are left alone.
+    @MainActor
+    static func quitAllTargets() -> [NSRunningApplication] {
```
abue-ammar, line 58: Nit: this is split out from `quitAll()` purely for testability, but the harness described in the PR body wasn't committed — so as it stands it's an extra internal entry point with one caller.

Either commit the harness under `Tools/`, or add a line noting it can't be made deterministic (it reads live system state). I ran the predicate standalone myself and it behaves exactly as described, so this is bookkeeping, not doubt about the logic.

Also cosmetic: `xmark.app` for the command vs `power` for the menu row is two glyphs for the same verb, and `power` is already `Quit Tinycast`'s symbol.

abue-ammar: `xmark.circle` is a great choice

Tinycast/Features/RootPaletteView.swift
abue-ammar: Fixed in #44 — `RootPaletteView.openActions()` samples the running state once, on open, into `selectionIsRunning`, so the rows can no longer change while the menu is up. All eight menu-open sites funnel through that one method, which is what keeps it to a single sample point.

Tinycast/Features/RootPaletteView.swift
abue-ammar: Fixed in #44 — the `@EnvironmentObject RunningAppsMonitor` is gone from `RootPaletteView` entirely (the snapshot above removed the need for it), so invalidation is back to `LauncherList` alone.

Tinycast/Features/RootPaletteView.swift
abue-ammar: Fixed in #44 — the handler now reads `app.kind == .application, core.runningApps.isRunning(app)`, identical to the menu row's condition.

Tinycast/Core/RunningApps.swift
abue-ammar: Fixed in #44 — `refresh()` compares against the current set and returns early when unchanged.

Tinycast/Core/AppCore.swift
abue-ammar: Addressed in #44 — an `NSAlert` confirmation in the existing `BackupActions` idiom. Two things fell out of it: the target list is resolved once and both counted and terminated (resolving twice would race), and `hasDestructiveAction` turned out to only tint the button — "Quit All" kept `keyEquivalent == "\r"`, so ↵ is explicitly moved to Cancel. Given this command sits one ↵ away in the palette, leaving ↵ on the destructive button would have made the guard nearly worthless.

Tinycast/Core/AppCore.swift
abue-ammar: Addressed in #44 — now `hidePalette(restoreFocus: !quittingPreviousApp)`, so focus returns to where the user was unless that app is the one being quit.

Tinycast/Core/AppLauncher.swift
abue-ammar: Addressed in #44 — `AppLauncher.quitAll()` is deleted. With `AppCore` resolving the list to count it for the alert, the terminate loop lives there and `quitAll()` had no callers left, so `quitAllTargets()` is now a genuinely-used seam rather than a test-only one. Glyph is `xmark.circle` per your call.

PR #35, opened 2026-07-27
Currency conversion in the inline calculator

### LINKED ISSUE #33: [Feature Request] Currency conversion in the inline calculator
The inline calculator already converts units (`10 km to mi`), number bases (`255 to hex`) and dates. It should also convert **currencies** from natural language, so the palette answers `1 euro to dollars` the same way it answers `1 m to ft`.

## Behaviour

Typing a currency request in the launcher (or in Calculator History search) shows the usual inline answer card:

```
1 EUR                    1.16 USD
Euro                     US Dollar
```

- The card shows the original amount, the converted amount and a clear label for each currency.
- Enter copies the answer and records it to Calculator History, exactly like every other calculator answer.

## Recognised input

- **Names and codes** — `euro`/`euros`/`EUR`, `dollar`/`dollars`/`USD`, `pound`/`pounds`/`GBP`, `yen`/`JPY`, plus the other widely used currencies. Regional dollars need their code or a qualifier (`CAD`, `AUD`) so a bare `dollars` stays USD.
- **Symbols** — `€20 to GBP`, `£50 in dollars`, `$100 to yen`, `¥1000 in euros`.
- **Connectors** — `to`, `in`, `->`, `→`, matching the existing unit-conversion phrasings.
- **Implied amount of 1** — `eur to usd` reads as `1 eur to usd`, like `m to ft`.
- **Expression on the left** — `2*50 usd to eur`.

## Disambiguation with units

`pound` is both a weight and a currency. The rule: a conversion where **both** sides resolve to compatible measurement units stays a unit conversion; currency only takes over when the unit path cannot answer.

- `10 pounds to kilograms` → weight (unchanged)
- `10 pounds` → weight (unchanged bare auto-conversion)
- `10 pounds to euros` → currency
- `10 usd to kg` → friendly `Cannot convert Currency to Weight.` error card, matching the existing category-mismatch message

## Rates

- Fetched from a free, key-less endpoint and cached on disk next to the other calculator data (`~/Library/Caches/<bundle-id>/`).
- Refreshed periodically so answers stay current, with a shorter retry after a failure.
- **Offline:** keep using the last downloaded snapshot. With no snapshot at all, the card says so rather than showing a wrong number, and an unknown currency reports that no rate is available.

## Constraints

- `Tinycast/Core/Calculator/` must stay Foundation-only and pure — the rate snapshot is passed **into** the engine, never fetched by it, so `Tools/calc-test.swift` keeps compiling the real sources against a fixed table.
- Networking and persistence live in a new manager owned by `AppCore`, wired in `start()`.

suiramdev:
Closes #33.

The inline calculator answers currency requests in natural language, on the same two-column answer card as unit conversions. **Off by default** — it needs the network, so it stays disabled until you enable it in Settings and accept a consent dialog.

```
100 dollars to yen   ->  100 USD (US Dollar)      = 16,374.00 JPY (Japanese Yen)
€20 to GBP           ->  20 EUR (Euro)            = 17.10 GBP (British Pound)
1 zloty to eur       ->  1 PLN (Polish Zloty)     = 0.23 EUR (Euro)
eur to usd           ->  1 EUR (Euro)             = 1.14 USD (US Dollar)
2*50 usd to cad      ->  100 USD (US Dollar)      = 140.82 CAD (Canadian Dollar)
1 idr to usd         ->  1 IDR (Indonesian Rupiah) = 0.00005565 USD (US Dollar)
10 usd to kg         ->  ⚠︎ Cannot convert Currency to Weight.
```

Enter copies the answer and records it to Calculator History like any other result.

## Consent

Currency conversion reaches the network, so it ships off. A new **Miscellaneous** settings tab holds a **Currency Conversion** toggle; flipping it on opens a dialog naming the provider, the request cadence and what does and doesn't leave the machine, with a link to the provider. The switch only moves after you accept — declining leaves it exactly where it was.

Without consent the feature doesn't exist: no request is made, no cached rates are read, and a currency query produces **no card at all** — not even the category-mismatch error, which would otherwise advertise a feature you never turned on. Turning it back off cancels the refresh loop, drops the snapshot and deletes the cached file.

The gate is a type rather than a boolean threaded through call sites:

```swift
enum CurrencySource { case off; case on(CurrencyRates?) }

static func evaluate(_ raw: String, currency: CurrencySource = .off) -> CalcResult?
```

Defaulting to `.off` means a caller that forgets to pass a source gets the feature **disabled**, not silently enabled. `CurrencyRateStore` re-checks consent at seven entry points rather than trusting its caller — reading the cache at init, the source handed to the engine, `start()`, each turn of the refresh loop, `refreshNow()`, and twice around the request itself, since consent can be withdrawn while a response is in flight.

The flag lives on the store, deliberately **not** in `AppSettings`: `SettingsBackup` mirrors that type field-for-field, so putting it there would let an imported config — or a Raycast import — grant network access without anyone agreeing to it.

## What's recognised

- **165 currencies**, by ISO code or name — `EUR`/`euro`/`euros`, `USD`/`dollar`/`dollars`, `GBP`/`pound`/`pounds`, `zloty`, `forint`, `taka`, `krónur`…
- **Signs** — `€ £ $ ¥ ₹ ₩ ₽ ₺ ₪ ₫ ฿ ₴ ₦ ₱` and 12 more. A leading sign is swapped back into amount-first order, so `€20 to GBP` and `20€ to GBP` parse identically.
- **Connectors** — `to`, `in`, `->`, `→`, plus an implied amount of 1 (`eur to usd`), matching the unit-conversion phrasings.
- **Expressions on the value side** — `2*50 usd to eur`.

## Not confusing money with measurements

`pound` is both. The rule is pipeline order: currency runs **after** the unit path, so a query both sides of which resolve to compatible units stays a measurement.

| Query | Result |
| --- | --- |
| `10 pounds to kilograms` | `10 lb` → `4.5359237 kg` (weight, unchanged) |
| `10 pounds` | `4.5359237 kg` (bare auto-conversion, unchanged) |
| `10 pounds to euros` | `10 GBP` → `11.70 EUR` (currency) |
| `1 cup to ml` | volume, even though `CUP` is the Cuban peso |
| `10 usd to kg` | `Cannot convert Currency to Weight.` |

## Rates

`CurrencyRateStore` (`Core/`, owned by `AppCore`) reads **[Frankfurter](https://frankfurter.dev)** — open source, no API key, no account, no quota, rates blended from 84 central banks.

- One `GET api.frankfurter.dev/v2/rates?base=USD`, ~1.4 KB gzipped, **at most once every 6h and only while enabled**. A launch with a cache under 6h old makes zero requests; a failed fetch retries in 15 min. Nothing user-triggered ever hits the network.
- Cached at `~/Library/Caches/<bundle-id>/currency-rates.json`, restored before any network exists, so a cold offline launch still answers from the last snapshot.
- With no snapshot at all the card says so rather than guessing; a currency the feed doesn't quote reports `No exchange rate for <CODE>.`

## Half the table is generated

`node Tools/gen-currencies.js` joins two sources on the ISO code and emits `CurrencyData.generated.swift` (165 names, 26 signs, 129 nouns), following the existing `gen-emoji.js` pattern:

- **Frankfurter** decides which currencies exist — the same feed the rates come from, so the table can't list something the app can't price.
- **CLDR** (`en`, from the pinned `cldr-json` checkout, not the host's `Intl`, whose output tracks the local ICU version) decides what humans call them. CLDR also supplies the sign tie-break for free: it writes every dollar but USD as `CA$`/`A$`/`NT$`, so plain `$` has exactly one claimant.

Only unambiguous data is emitted. What stays hand-written is one table of **21 contested nouns** — `dollars` is claimed by 22 currencies, `francs` 10, `pounds` 9 — where CLDR correctly refuses to choose and the calculator must. Genuinely ambiguous words are assigned to nobody: `krona` is both SEK and ISK, so it produces no card.

## Keeping the engine pure

`Core/Calculator/` stays Foundation-only **and** pure: the store hands `CalcEngine.evaluate` a finished `CurrencySource`, exactly as the clock is injected as `now`/`calendar`. `CalcMemo` keys its memo on the consent flag plus the snapshot's `fetchedAt`, so flipping the setting or landing a fresh table invalidates it without diffing every rate.

## Verification

- `swiftc Tinycast/Core/Calculator/*.swift Tools/calc-test.swift` → **244 passed, 0 failed** (73 new, including the consent gate and the generated-table wiring).
- `xcodebuild -scheme Tinycast -configuration Debug` → clean.
- Driven through the real app end to end: with consent withheld, a 10-second socket poll showed **no outbound connections**, no cache file, and no card for any currency query while units and arithmetic were unaffected. Accepting the dialog wrote the cache and answered `1 EUR = 1.14 USD`; declining changed nothing; toggling back off deleted the cache and silenced the feature again.


suiramdev:
Closing per @abue-ammar's call in https://github.com/abue-ammar/tinycast/issues/33#issuecomment-5091061812 — Tinycast is intentionally fully offline, and this feature needs an external API for live rates, so it doesn't fit the project's current design.

Fair enough. Nothing here is lost: the branch `suiramdev:feat/unit-conversion` stays as-is, so **this PR can be re-opened at any time if @abue-ammar decides to accept it** (or if an offline-friendly approach turns up — the engine side is already rate-source agnostic: `CalcEngine.evaluate` takes a `CurrencyRates` snapshot and never fetches, so swapping `CurrencyRateStore` for a bundled or user-supplied table would be a contained change).

abue-ammar:
@suiramdev hey thanks for opening PR! i haven't fully looked at the code yet. but are you using any api? if yes when and how often are we calling the api and if you used any caching.

suiramdev:
> @suiramdev hey thanks for opening PR! i haven't fully looked at the code yet. but are you using any api? if yes when and how often are we calling the api and if you used any caching.

Yes, it uses one API: https://open.er-api.com/.

It makes at most one request every six hours. There is a single call site in `AppCore.start()`, which fetches new rates only when the cached snapshot has expired. If the app launches with a cache less than six hours old, it makes no network request.

The cache is stored at `~/Library/Caches/<bundle-id>/currency-rates.json`.

That said, the feature is not fully offline. Apple does not provide a built-in exchange-rate API, and current rates necessarily come from an external source.

We could place the feature behind an opt-in toggle, so the app makes zero external connections unless the user explicitly enables it. However, that may not align with the audience you are targeting?

abue-ammar:
> > @suiramdev hey thanks for opening PR! i haven't fully looked at the code yet. but are you using any api? if yes when and how often are we calling the api and if you used any caching.
> 
> Yes, it uses one API: https://open.er-api.com/.
> 
> It makes at most one request every six hours. There is a single call site in `AppCore.start()`, which fetches new rates only when the cached snapshot has expired. If the app launches with a cache less than six hours old, it makes no network request.
> 
> The cache is stored at `~/Library/Caches/<bundle-id>/currency-rates.json`.
> 
> That said, the feature is not fully offline. Apple does not provide a built-in exchange-rate API, and current rates necessarily come from an external source.
> 
> We could place the feature behind an opt-in toggle, so the app makes zero external connections unless the user explicitly enables it. However, that may not align with the audience you are targeting?

1st https://open.er-api.com does this requires apikey? if yes then it has limit  or need money, why aren't we choosing non apikey based apis like https://api.frankfurter.app (idk if it's capable) for our usecase. maybe we can have a new settings called " Misc" there we can put a toggle with description showing app will be connected to network and bla bla. and about caching since app starts at login we can fetch from api and cache it for whole day. in general i would like less aggressive than 6hrs. lemme know what do you think 

edit: a refresh button wherever suits better either in calc or in the misc

aiexz:
It would be awesome to see crypto support as well (e.g. BTC, ETH, SOL)

abue-ammar:
> It would be awesome to see crypto support as well (e.g. BTC, ETH, SOL)

if a fully free api is capable then sure it has to be single api

suiramdev:
Obviously, free APIs may have some latency when updating exchange rates, so they’re not entirely reliable, especially for trading. However, I used this feature in Raycast when planning trips, where an approximate conversion is usually enough to estimate how much things will cost.

I didn’t choose this API deliberately. You're free to change to whatever is more reliable.

nank1ro:
Raycast uses its own backend for currencies:
- https://backend.raycast.com/api/v1/currencies
- https://backend.raycast.com/api/v1/currencies/crypto?symbols=ADA,AVAX,BCH,BNB,BSV,BTC,DASH,DOGE,DOT,EOS,ETC,ETH,LTC,LUNA,NEO,POL,SHIB,SOL,TRX,USDT,XLM,XMR,XRP

abue-ammar:
> Raycast uses its own backend for currencies:
> 
> * https://backend.raycast.com/api/v1/currencies
> * https://backend.raycast.com/api/v1/currencies/crypto?symbols=ADA,AVAX,BCH,BNB,BSV,BTC,DASH,DOGE,DOT,EOS,ETC,ETH,LTC,LUNA,NEO,POL,SHIB,SOL,TRX,USDT,XLM,XMR,XRP

@suiramdev maybe we can use this? it's free no fast

suiramdev:
I’ll proceed with updating the PR. If this gets merged, how about adding a toggle in Settings to enable the feature? Rather than introducing a separate toggle for each feature, we could use a single setting for all features that make online requests, ensuring the user provides explicit consent.

abue-ammar:
@suiramdev found one thing about https://open.er-api.com for free api their results are sane for 24 hrs. they reset after 24 hrs for free api key holders. so 6 hrs cache isn't redundant

suiramdev:
> > Raycast uses its own backend for currencies:
> > 
> > * https://backend.raycast.com/api/v1/currencies
> > * https://backend.raycast.com/api/v1/currencies/crypto?symbols=ADA,AVAX,BCH,BNB,BSV,BTC,DASH,DOGE,DOT,EOS,ETC,ETH,LTC,LUNA,NEO,POL,SHIB,SOL,TRX,USDT,XLM,XMR,XRP
> 
> @suiramdev maybe we can use this? it's free no fast

Actually, I'm afraid for you that this may break their TOS. I don't know if you're legally able to use their API route on a tier app...

abue-ammar:
> I’ll proceed with updating the PR. If this gets merged, how about adding a toggle in Settings to enable the feature? Rather than introducing a separate toggle for each feature, we could use a single setting for all features that make online requests, ensuring the user provides explicit consent.

yes definitely we gotta add a new settings and  concent is needed. but where are we putting that tho. btw do you have discord you can reach out to me so we can think of where can we put it. because some other extras are coming too like quicklink, snippets

suiramdev:
This seems more likely to be aligned with tinytask's principles : https://frankfurter.dev/

suiramdev:
Pushed three commits answering your question about the API, plus the settings toggle.

**Provider swapped.** Now [Frankfurter](https://frankfurter.dev) instead of exchangerate-api — open source, no key, no account, no quota, rates blended from 84 central banks rather than one vendor's feed. One `GET /v2/rates?base=USD`, ~1.4 KB gzipped, at most once every 6h.

**The currency table is generated, not hand-typed.** `node Tools/gen-currencies.js` joins Frankfurter's currency list with Unicode CLDR and emits `CurrencyData.generated.swift` — 165 codes, names, 26 signs and 129 nouns — following the same pattern as `gen-emoji.js`. Only 21 hand-written entries remain, the nouns several currencies genuinely share (`dollars` is claimed by 22 currencies, so CLDR rightly refuses to pick one and we have to).

We may want to generate it not on build time, but when the user toggles the features on.

**Consent gate, off by default.** New **Miscellaneous** tab with a **Currency Conversion** toggle. Enabling opens a dialog naming the provider, the cadence and what leaves the machine, with a link out; the switch only moves after you accept. Without consent no request is made, no cached rates are read, and a currency query shows no card at all. Turning it off deletes the cache. The flag deliberately lives on the store rather than in `AppSettings`, since `SettingsBackup` mirrors that type and an import shouldn't be able to grant network access.

I verified the gate at runtime rather than just in tests: with consent withheld, a 10-second socket poll on the running app showed no outbound connections at all, while units and arithmetic kept working.

One thing I'd like to revisit: **the Settings UI is wordier than it needs to be.** The consent sheet has four bullets and the pane carries a callout on top of that — thorough, but probably more detail than belongs in a settings pane at rest. I'll take a pass at tightening it. Happy to hear how much you want surfaced there versus tucked behind the link, since that's a taste call on your side of the project.

PR #51, opened 2026-07-28
feat: Add Snippets support with Markdown file persistence and keyword expansion

elalecs:
## Summary
This PR introduces full **Snippets** support to Tinycast, enabling quick text expansion, dynamic template variables, launcher search integration, and file-based Markdown persistence.

### Markdown File-Based Persistence
- Snippets are stored individually as `.md` files in `~/.config/tinycast/snippets/`.
- YAML Frontmatter headers store metadata (`name`, `keyword`, `category`, `enabled`, `show_in_launcher`).
- Automatically updates in real-time when files are modified, created, or deleted externally via `DispatchSource` directory watching.

### Keyword Auto-Expansion
- Global background event tap (`CGEventTap`) listens for keyword triggers (e.g., `:sh`, `!notes`) across any macOS app.
- Pastes expanded templates into the active application cleanly.

### Dynamic Template Engine & Variable Support
- Supports dynamic placeholders: `{cursor}`, `{clipboard}`, `{date}`, `{time}`, `{argument name="..."}`, and `{snippet:Name}`.
- Native interactive prompt modal for required template argument inputs.

### Launcher Integration & Searchability
- Snippets appear directly in Tinycast's main launcher search under a dedicated **"Snippets"** section.
- Searchable by name, keyword, or category.
- `show_in_launcher` toggle allows choosing whether a snippet appears in launcher search.

### Settings UI & Permissions
- Dedicated Snippets tab in Settings with uniform macOS switch controls, list sidebar, and template variable helper buttons.
- Updated Onboarding and Permissions Settings to check and request **Input Monitoring** permission alongside **Accessibility**.

### Demo

https://github.com/user-attachments/assets/5e3fdec2-a516-4b45-8090-a7c864a07e44


abue-ammar:
hey thanks a lot! i will review it and let you know. and could you keep the CLAUDE.md because it holds lots of quirks which are important. 

edit: adding a demo video would be great

elalecs:
Here a demo

https://github.com/user-attachments/assets/d9aa3dfb-b2ad-47b6-a4c3-ff85b9946513

abue-ammar:
> Here a demo
> 
>  tinycast-snippets.mp4

great work! functionality seems working nice. but in settings UI need to be better something that matches our design aesthetic we will get to that soon after i review actual code. and it is not mandatory but i would encourage you to join our discord community we can polish this up more https://discord.gg/v2Eeb4QQy3

also if there is any functionality gaps between raycast snippets and ours you can mention them so that our users will know what we have and what we don't. so far great work!

abue-ammar:
@elalecs updated contributing guidelines please read, update your repo make sure you followed and accepted https://github.com/abue-ammar/tinycast/blob/main/CONTRIBUTING.md

elalecs:
Please let me know which parts I should focus on. According to the CONTRIBUTING.md, there could be several elements. I would appreciate it if you could explain your concerns.

abue-ammar:
@elalecs lemme rebase it. main branch is updated. and please reach out to me on discord. lets polish it further

abue-ammar:
Thanks for this, @elalecs — the demo looks great and file-based Markdown storage is the right instinct for snippets.

I've gone through it closely and I'm going to close it. Sorry to do that after this much work; the blockers are structural rather than a list of nits:

- **Snippet identity.** `Snippet.id` is re-minted on every parse and never persisted, so `selectedSnippetID` goes stale on every reload: clicking **+** leaves the detail pane on the placeholder, typing one character blanks it, renaming writes a new `.md` per keystroke (I verified — 3 characters typed left 4 files on disk), and `delete(id:)` targets `slug(name).md` so it silently no-ops on any hand-authored file whose filename doesn't match its `name:`. Frontmatter is unescaped as well, so the Name field can inject `keyword:`/`enabled:`. The file URL needs to be the identity.
- **Settings pane.** It bypasses `SettingsPane`/`SettingsCard`/`SettingsRow` and nests a 220pt sidebar inside a window that already has one — at 720×550 minus the 184pt sidebar that leaves ~295pt for the form, less than the toggle row alone needs — and every keystroke does a synchronous full-directory read + parse on the main actor.
- **Permissions.** This doesn't follow `HyperKeyTap`, which solves the same problem in this repo. The tap is created unconditionally at launch, `CGRequestListenEventAccess()` fires with no user gesture (and once per second forever while Accessibility is ungranted, since the prompt sits above the trust guard), and `stop()` is never called, so there's no way to disable it. A global keystroke listener has to ship off, behind a Settings toggle.
- **Tap lifecycle.** The run loop source is added to the spawned thread's run loop but removed from `CFRunLoopGetMain()`, so the thread and tap leak for the process lifetime; `eventTap` is `nonisolated(unsafe)` and written cross-thread; and `isTapStarting` is never reset, which makes the health-timer retry permanently dead.
- **Shared paste path.** `postCommandV`'s fallback still changes `.cghidEventTap` → `.cgAnnotatedSessionEventTap`, which affects clipboard, emoji and calculator pasting and isn't needed by this feature. The clipboard injection branch also replaces the user's pasteboard without restoring it.
- **Conventions.** No `Tools/` harness (the serializer bugs above would all have been caught), the template engine isn't Foundation-pure so it can't be harnessed, storage isn't keyed by bundle id so `Tinycast Dev` and an installed build share the same files, and there's no `docs/snippets.md`.

Reworking identity, the consent gate and the tap lifecycle is closer to a rewrite than a review round, so I'd rather close this than leave it open indefinitely.

If you'd like another run at it, I'm happy to agree the design up front — snippet identity and storage location, the consent gate, and reusing the `HyperKeyTap` shape for the tap. Thanks again for the time you put in, and sorry for the outcome.

PR #53, opened 2026-07-28
Open the palette on the display under the cursor

drshnk3:
## Problem

On a multi-display setup the palette always opens on the display that holds the menu bar, regardless of
where the pointer is. With an external monitor set as the main display, summoning the launcher from the
built-in screen still puts the window on the external one — you have to look away from where you were
working to use it.

## Cause

`PaletteWindowController.resolveAnchor()` resolved its screen from `NSScreen.main`:

```swift
guard let screen = NSScreen.main else { return nil }
let visible = screen.visibleFrame
```

`NSScreen.main` is documented as the screen containing the window with keyboard focus. Tinycast is an
accessory app (`.accessory` activation policy) and the palette is a `.nonactivatingPanel`, so at the
moment the anchor is resolved there is no key window on the display the user is looking at — `main`
falls back to the display with the menu bar. The anchor is correct for a single-display machine, which
is why this only shows up with more than one screen attached.

## Fix

A new `targetScreen()` picks the display before the anchor is computed:

```swift
private func targetScreen() -> NSScreen? {
    guard core.settings.openOnCursorScreen else { return NSScreen.main }
    let mouse = NSEvent.mouseLocation
    return NSScreen.screens.first { NSMouseInRect(mouse, $0.frame, false) } ?? NSScreen.main
}
```

`NSEvent.mouseLocation` and `NSScreen.frame` share the same bottom-left screen coordinate space, so the
containment test needs no conversion — but it does need the right interval, which is why this is
`NSMouseInRect` and not `CGRect.contains`. A mouse location is the CoreGraphics cursor position flipped
about the primary display's height, so a screen's rows land in the half-open interval `(minY, maxY]`:
the topmost row is exactly `maxY`, which `contains` excludes, while that same value is the `minY` of the
display stacked above. `contains` would hand a pointer parked at the top of one display to its
neighbour. `NSMouseInRect(_:_:false)` is exactly this predicate and already ships in AppKit for this
purpose. The `?? NSScreen.main` is a defensive fallback for a location that matches no frame at all.

Everything else about placement is unchanged — the anchor is still resolved **once per summon** and held
for the session, so a compact↔expanded resize still only changes the height and can never move the
window, and dragging the pointer to another display while the palette is open does not teleport it.

## The setting

Exposed as `AppSettings.openOnCursorScreen`, with a toggle in **Settings → General → Appearance**:

> **Follow the cursor across displays** — Open the launcher on whichever display the pointer is on,
> rather than the one with the menu bar.

It is mirrored in `SettingsBackup` (`SettingsData.openOnCursorScreen`, gather + apply) since that type
mirrors `AppSettings` field-for-field. It is an ordinary UI preference — no network, no consent gate —
so it belongs in `AppSettings` rather than on a store of its own. The backup field is `Bool?`, so
importing a pre-#53 backup leaves the current value alone rather than forcing `false`.

### Defaulting it to `true` (resolved: keeping it on)

Absence is treated as `true`, following the same idiom as `showFavoritesInCompactMode` and
`hyperKeyIncludesShift`:

```swift
openOnCursorScreen =
    defaults.object(forKey: Key.openOnCursorScreen) == nil
    || defaults.bool(forKey: Key.openOnCursorScreen)
```

The reasoning: the current behavior reads as a bug rather than a preference, and defaulting on fixes it
for existing multi-display users without them having to find the switch. The trade-off, accepted
knowingly, is that a multi-display user who had adapted to the menu-bar placement gets moved once on
upgrade, with a discoverable switch to undo it.

## Known trade-off

The cursor is not always where attention is: summon by hotkey with the pointer parked on the external
display while typing on the built-in, and the palette follows the pointer, not your eyes. Reading the
focused window of the recorded `previousApp` would be more accurate but needs AX geometry on the summon
path for a marginal gain. The setting name says "follow the cursor", which is honest about what it does.

## Testing

Verified on a two-display setup (3440x1440 external set as main + built-in Retina), Debug channel:

- pointer on the external main display, toggle on — palette opens on the external display
- pointer on the built-in display, toggle on — palette opens on the built-in display (the reported bug)
- pointer on the built-in display, toggle off — palette opens on the external display (old behavior)
- palette open, pointer dragged to the other display, then typing to trigger compact→expanded — window
  stays put, top edge does not drift

`xcodebuild -scheme Tinycast -configuration Debug` succeeds with no new warnings.

## Files

| File | Change |
|---|---|
| `Tinycast/Core/PaletteWindowController.swift` | `targetScreen()`; `resolveAnchor()` calls it instead of reading `NSScreen.main` |
| `Tinycast/Core/AppSettings.swift` | `openOnCursorScreen` + its defaults key, defaulting to `true` |
| `Tinycast/Features/Settings/GeneralSettingsView.swift` | Toggle row in the Appearance card |
| `Tinycast/Core/Backup/SettingsBackup.swift` | Mirrored field in `SettingsData`, gather + apply |
| `docs/palette.md` | New **Window placement** section covering the anchor and the `NSScreen.main` gotcha |

No generated files touched, and no changes to `project.yml` (so no `xcodegen generate` needed).

### DIFF
--- a/Tinycast/Core/AppSettings.swift
+++ b/Tinycast/Core/AppSettings.swift
@@ -138,7 +138,6 @@ final class AppSettings: ObservableObject {
             || defaults.bool(forKey: Key.showFavoritesInCompactMode)
         // An unset key means "never configured" and seeds the defaults; a stored empty array is a user who deliberately cleared the list.
         searchScopes = defaults.stringArray(forKey: Key.searchScopes) ?? SearchScopes.defaults
-        // Also defaults to true, so absence must be distinguished from a stored `false`.
         openOnCursorScreen =
             defaults.object(forKey: Key.openOnCursorScreen) == nil
             || defaults.bool(forKey: Key.openOnCursorScreen)
--- a/Tinycast/Core/AppSettings.swift
+++ b/Tinycast/Core/AppSettings.swift
@@ -39,6 +39,7 @@ final class AppSettings: ObservableObject {
         static let compactMode = "compactMode"
         static let showFavoritesInCompactMode = "showFavoritesInCompactMode"
         static let searchScopes = "launcherSearchScopes"
+        static let openOnCursorScreen = "openOnCursorScreen"
     }
 
     /// Folders (and individual `.app` bundles) `AppIndex` scans, in scan order. Editing this re-indexes — `AppIndex.start(settings:)` observes it.
@@ -98,6 +99,11 @@ final class AppSettings: ObservableObject {
         didSet { defaults.set(showFavoritesInCompactMode, forKey: Key.showFavoritesInCompactMode) }
     }
 
+    /// Summon the palette on the display under the pointer instead of the one holding the menu bar.
+    @Published var openOnCursorScreen: Bool {
+        didSet { defaults.set(openOnCursorScreen, forKey: Key.openOnCursorScreen) }
+    }
+
     init() {
         // integer(forKey:) returns 0 when unset, which no case matches — falls through to 3 Months.
         clipboardRetention =
@@ -132,5 +138,10 @@ final class AppSettings: ObservableObject {
             || defaults.bool(forKey: Key.showFavoritesInCompactMode)
         // An unset key means "never configured" and seeds the defaults; a stored empty array is a user who deliberately cleared the list.
         searchScopes = defaults.stringArray(forKey: Key.searchScopes) ?? SearchScopes.defaults
+        // Also defaults to true: following the pointer is what a multi-display user expects, and the
+        // alternative pins the palette to the menu-bar display no matter where they're working.
+        openOnCursorScreen =
+            defaults.object(forKey: Key.openOnCursorScreen) == nil
+            || defaults.bool(forKey: Key.openOnCursorScreen)
     }
 }
--- a/Tinycast/Core/AppSettings.swift
+++ b/Tinycast/Core/AppSettings.swift
@@ -138,8 +138,7 @@ final class AppSettings: ObservableObject {
             || defaults.bool(forKey: Key.showFavoritesInCompactMode)
         // An unset key means "never configured" and seeds the defaults; a stored empty array is a user who deliberately cleared the list.
         searchScopes = defaults.stringArray(forKey: Key.searchScopes) ?? SearchScopes.defaults
-        // Also defaults to true: following the pointer is what a multi-display user expects, and the
-        // alternative pins the palette to the menu-bar display no matter where they're working.
+        // Also defaults to true, so absence must be distinguished from a stored `false`.
         openOnCursorScreen =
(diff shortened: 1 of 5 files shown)


abue-ammar:
could you add a demo video in the pr summary? also how it behaves in compact mode.  i didn't plan on adding unnecessary features making it bloated like raycast.

abue-ammar:
@dk5761 updated contributing guidelines please read update your repo rebasing it, make sure you followed and accepted https://github.com/abue-ammar/tinycast/blob/main/CONTRIBUTING.md

suiramdev:
@abue-ammar It's hard to record a dual screen, but I can confirm you I tested it out and it works. It opens the launcher on the screen where the cursor is. It's mergeable if the code suits you, I haven't looked to the code myself

suiramdev:
> @abue-ammar It's hard to record a dual screen, but I can confirm you I tested it out and it works. It opens the launcher on the screen where the cursor is. It's mergeable if the code suits you, I haven't looked to the code myself

Screenshot 2026-07-29 at 11 50 54

https://github.com/user-attachments/assets/8e069520-17d2-4cb8-924f-301c21f836f2

drshnk3:
@abue-ammar  My apologies, I was not able to respond promptly. I will read the guidelines, rebase the repo and update the pr.

@suiramdev  thanks for testing the code. I will also look if there is any way to provide a video which can show the functionality.

abue-ammar:
> @abue-ammar My apologies, I was not able to respond promptly. I will read the guidelines, rebase the repo and update the pr.
> 
> @suiramdev thanks for testing the code. I will also look if there is any way to provide a video which can show the functionality.

great! since it's working i will review the code and  merge it. btw one change request i want. could you check if raycast by default  does that or they ha a settings to toggle. if by default  they does that we don't  need a toggle settings

drshnk3:
> > @abue-ammar My apologies, I was not able to respond promptly. I will read the guidelines, rebase the repo and update the pr.
> > @suiramdev thanks for testing the code. I will also look if there is any way to provide a video which can show the functionality.
> 
> great! since it's working i will review the code and merge it. btw one change request i want. could you check if raycast by default does that or they ha a settings to toggle. if by default they does that we don't need a toggle settings

Raycast has a toggle.

image

abue-ammar:
> > > @abue-ammar My apologies, I was not able to respond promptly. I will read the guidelines, rebase the repo and update the pr.
> > > @suiramdev thanks for testing the code. I will also look if there is any way to provide a video which can show the functionality.
> > 
> > 
> > great! since it's working i will review the code and merge it. btw one change request i want. could you check if raycast by default does that or they ha a settings to toggle. if by default they does that we don't need a toggle settings
> 
> Raycast has a toggle.
> 
> image

cool then!. lemme quick review the actual code. then i will merge it

PR #55, opened 2026-07-28
Improve keyboard selection scrolling

### LINKED ISSUE #100: [Bug]: Keyboard selection scrolls before reaching the visible edge
### What happened?

When navigating launcher results with the Up and Down arrow keys, the list can begin scrolling while the selected row is still inside the visible viewport. In some transitions the selection advances first and the list catches up on the next frame. After scrolling, changing the query can also briefly preserve the previous offset, causing the first result to overlap the search header.

Expected behavior:

- Keep the list stationary while the selected row remains visible.
- Once the selection reaches an edge, scroll only enough to reveal the next row.
- Returning to the first result restores the complete top inset and section header.
- Changing or clearing the query consistently resets the filtered list to its natural top position.
- The same behavior remains consistent in compact mode and shared list screens.

Current main · Tinycast Dev

### macOS version

26.5.2

matijrn:
## Related issue

Closes #100

## What changed

- Keep launcher, clipboard, and calculator-history lists stationary while keyboard selection remains visible.
- Minimally reveal the next row when selection reaches a viewport edge.
- Restore the exact inset-aware top origin when navigating to the first row or changing/resetting a query.
- Keep filtered scroll containers mounted so lazy list updates cannot restore a stale offset for one frame.
- Reassert the top origin after compact mode expands.

The implementation uses an explicit `ListScrollIntent`: follow requests use unanchored `ScrollViewReader.scrollTo`, while top requests reset the backing `NSClipView` to its constrained native origin.

## Why the additional scroll handling is needed

- A simple `scrollTo` left the previous offset active when typing after keyboard navigation, causing filtered results to appear partially displaced.
- Recreating the filtered scroll view could briefly render the stale offset, producing a one-frame jump or flicker.
- Expanding from compact mode changes the layout after the initial reset, so the top position must be reasserted once the expanded layout is available.
- Scrolling only to the first result does not restore the search/header inset precisely, which is why the native clip-view origin is reset for top requests.
- The same behavior is shared by launcher, clipboard, and calculator history to keep keyboard navigation consistent.

## Memory footprint

Measured with `footprint` after 300 Down presses, 300 Up presses, and 10 repeated query-change cycles.

| Idle after launch       | 21 MB  |
| Palette open            | 35 MB  |
| Feature in use (peak)   | 80.7 MB |
| Palette closed, settled | 80 MB  |

Leak-tested: yes. `leaks` reports 287 allocations / 14,320 bytes in the existing AppIntents `NSXPCConnection` cycle; the count and allocation tree are identical on current `main`. No new app-attributable leak or cumulative growth was observed. `xcodebuild analyze` succeeds.

## Demo

Side-by-side video uses the official v0.7.5 stable build on the left and this PR on the right. Both start in compact mode and use the same 750 × 475 expanded window and actions: type `a`, press Down eight times, then Up eight times. v0.7.5 recenters the selection and moves the search/header area while the selected row is still visible; the PR keeps the viewport stationary until an edge is reached.

https://github.com/user-attachments/assets/f8cf67fe-fcf4-4a27-a736-23e4328a94e3

## Drawbacks

After a full icon-list traversal, macOS image and Core Animation caches remain warm, so the process does not return to its cold-launch 21 MB footprint. Current `main` has the same behavior. A second stress cycle remains at the same warm baseline with no cumulative growth.

## Tests & validation

- Clean unsigned Debug build: `xcodebuild ... clean build`
- Static analysis: `xcodebuild ... analyze`
- `swift Tools/fuzz-test.swift`
- `Tools/ranking-test.swift`
- `Tools/calc-test.swift` — 426 passed
- `Tools/clipboard-test.swift` — 23/23 passed
- `Tools/scopes-test.swift`
- `Tools/emoji-test.swift` — 2,054 records
- `Tools/custom-command-test.swift`
- Manual launcher navigation in expanded and compact mode, including edge scrolling, returning to the first result, repeated `w`/`wh` query changes, and clearing the query after scrolling.
- Manual Clipboard History navigation/filter reset and Calculator History launch.

I have read and accept the Contributor License and Feedback Agreement.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved edge dissolve behavior so scroll boundaries transition smoothly as content moves behind them.
  * Fixed stale scroll offsets and restored correct top positioning when switching queries, modes, or selections.
  * Reduced unwanted scrolling during mouse selection while preserving minimal row visibility during keyboard navigation.

* **Documentation**
  * Updated UI documentation to describe the revised edge masking and selection scroll behavior.

### DIFF
--- a/Tinycast/Core/EdgeDissolve.swift
+++ b/Tinycast/Core/EdgeDissolve.swift
@@ -50,16 +50,18 @@ struct EdgeDissolveMask: ViewModifier {
 
     private func stops(height: CGFloat) -> [Gradient.Stop] {
         guard canScroll, height > 0 else { return [.init(color: .black, location: 0)] }
-        // Midpoint alpha eases from 1 toward the floor as a full band of content scrolls past (Raycast: opacity = 1 − (1 − min) · clamp(scrollDistance / fadeHeight, 0, 1)).
-        let topAlpha = 1 - (1 - Self.topMinAlpha) * min(topDistance / topFade, 1)
-        let bottomAlpha = 1 - (1 - Self.bottomMinAlpha) * min(bottomDistance / bottomFade, 1)
+        // Each edge stays opaque at its resting boundary, then dissolves as content passes behind it.
+        let topProgress = min(topDistance / topFade, 1)
+        let bottomProgress = min(bottomDistance / bottomFade, 1)
+        let topAlpha = 1 - (1 - Self.topMinAlpha) * topProgress
+        let bottomAlpha = 1 - (1 - Self.bottomMinAlpha) * bottomProgress
         return [
-            .init(color: .black.opacity(0), location: 0),
+            .init(color: .black.opacity(1 - topProgress), location: 0),
             .init(color: .black.opacity(topAlpha), location: topFade / 2 / height),
             .init(color: .black, location: topFade / height),
             .init(color: .black, location: 1 - bottomFade / height),
             .init(color: .black.opacity(bottomAlpha), location: 1 - bottomFade / 2 / height),
-            .init(color: .black.opacity(0), location: 1),
+            .init(color: .black.opacity(1 - bottomProgress), location: 1),
         ]
     }
 }
--- a/Tinycast/Core/ThinScrollbar.swift
+++ b/Tinycast/Core/ThinScrollbar.swift
@@ -178,6 +178,73 @@ extension View {
     func hideNativeScrollers() -> some View {
         background(NativeScrollerHider().frame(width: 0, height: 0))
     }
+
+    /// Attach inside a `ScrollView` to restore its backing clip view to AppKit's exact inset-aware top edge when lazy filtering would otherwise preserve a stale offset.
+    func resetNativeScrollToTop(id: UUID?) -> some View {
+        background(NativeScrollTopResetter(resetID: id).frame(width: 0, height: 0))
+    }
+}
+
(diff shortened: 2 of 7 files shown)


abue-ammar:
please add demo videos side by comparison what benifits we got

abue-ammar:
@matijrn updated contributing guidelines please read update your repo make sure you followed and accepted https://github.com/abue-ammar/tinycast/blob/main/CONTRIBUTING.md

abue-ammar:
@matijrn great work!  we have 2 more viewer
1. emoji panel
2. calculator history ( you can find it in app launcher)

IMPORTANT: in emoji panel we have vertical + horizontal navigation. try opening emoji panel with hotkeys last time i found there was an issue that when you launch emoji panel it renders with already scrolled up a bit. try in compact mode as well as in expanded mode (you can find the toggle in general settings). make sure no placeholder text and icon left to it has no layout shift while launching emoji panel, clipboard history, app launcher one by one (it did a quirky behavior when i was implementing compact mode, some how managed to fix). a full video would be great.

abue-ammar:
i have manually checked everything is working

abue-ammar:
@matijrn  your implementation actually breaks our edge dissolve effect and makes it soo transparent. so i have implemented in a different way without touching those. if you want to check #109 . thanks for your PR it helped

matijrn:
Thanks for checking it and for sharing the alternative implementation, glad #55 helped

PR #70, opened 2026-07-28
Add typed calculator expressions

### LINKED ISSUE #64: [Feature]: Powerful Calculator expressions
### What should it do?

Be able to parse things like `10k`, `10$`, `10kg + 10g`, etc.

Nothing here:
Image

Only works with the most basic expressions for now
Image

This was the most useful thing in the Raycast by far - instant calculator at hand
https://www.raycast.com/core-features/calculator
https://manual.raycast.com/calculator

kynnyhsap:
> **Development disclosure**
>
> This PR was created with Codex using GPT-5.6-sol. The calculator behavior was developed test-first: the desired cases and failure boundaries were added to the standalone harness before the engine implementation. The finished Debug build was then exercised end to end in the real Tinycast UI using Computer Use, including compact numbers, mixed-unit arithmetic, percentages, composite units, and incomplete expressions.

Closes #64.

## Summary

The inline calculator can now evaluate typed expressions across measurements and consent-gated currencies instead of stopping at single-value conversions.

| Query | Result |
| --- | --- |
| `10k` | `10,000` |
| `10kg + 10g` | `10.01 kg` |
| `500g + 1kg` | `1,500 g` |
| `(10kg + 500g) * 3%` | `0.315 kg` |
| `5kg / 500g` | `10` |
| `5 feet 3 inches to cm` | `160.02 cm` |
| `$10 + €5` | `15.43 USD` against the injected rate table |
| `10kg * 3% +` | keeps `0.3 kg` visible while the next operand is typed |

Compatible addition and subtraction convert the right operand into the leftmost unit, so the query decides the output unit. Multiplication and division by scalars preserve it; dividing compatible quantities cancels the dimension and returns a scalar. Standard precedence, parentheses, unary signs, factorials, percentages, adjacent composite units, and a final `to` / `in` conversion all work in the same expression.

## Ambiguity and safety

The attached `k` from the issue is a thousands suffix: `10k` is `10,000`, and `1kUSD` is `1,000 USD`. Whitespace keeps Kelvin explicit (`10 k to c`), while the established attached conversion form remains valid when the target makes the meaning unambiguous (`273.15K to C`).

Percentages follow calculator semantics:

- `10kg + 20%` → `12 kg`
- `10kg - 20%` → `8 kg`
- `10kg * 3%` → `0.3 kg`
- `10kg / 25%` → `40 kg`

The parser deliberately does not invent derived dimensions. `1kg * 1m` returns a clear error instead of guessing a unit, and affine temperatures can only be added or subtracted on the same scale. Malformed or genuinely incomplete input remains silent.

## Incomplete expressions

A trailing binary operator now preserves the last complete, actionable result:

- `10 +` → `10` (`Result`)
- `10kg +` → `10 kg` (`Kilograms`)
- `10kg + 500g +` → `10.5 kg` (`Kilograms`)

The existing calculator card and badge layout are unchanged; this only keeps feeding it the last valid result and its metadata while the user types.

## Implementation

`CalcQuantity` is a Foundation-only typed precedence parser beside the existing scalar `CalcParser`. Keeping the two separate avoids teaching the scalar evaluator about dimensions, while still reusing `CalcToken`, unit definitions, formatter behavior, and the injected `CurrencySource`.

Currency arithmetic preserves the existing safety boundary: no new requests, settings, cache behavior, or dependencies. With consent off the path does not exist; with consent on it uses only the already-injected immutable rate snapshot.

`CalcEngine` runs the typed path before currency conversion so unit aliases such as `pounds` continue to win when both sides are measurements. The Xcode project was regenerated with XcodeGen after adding the source file, and the calculator/UI documentation now reflects the expanded pipeline and result badges.

## Verification

- `swift Tools/fuzz-test.swift` → all passed.
- `swiftc -swift-version 6 Tinycast/Core/LauncherRankingStore.swift Tools/ranking-test.swift ...` → all passed.
- `swiftc -swift-version 6 Tinycast/Core/Calculator/*.swift Tools/calc-test.swift ...` → **397 passed, 0 failed**.
- `swiftc -swift-version 6 Tinycast/Core/ClipboardStore.swift Tools/clipboard-test.swift ...` → **23/23 passed**.
- Clean Xcode 26 Debug build on macOS 26.4 → **BUILD SUCCEEDED**, no new warnings.
- Driven through `Tinycast Dev.app`: compact numbers, mixed-unit addition, parentheses + percentages, composite units, and a trailing operator all produced the expected card, copy value, and target-unit badge.
- Apple `footprint` over an open → evaluate `10kg * 3% +` → close cycle: **35 MB idle, 35 MB peak**.
- Apple `leaks` after the same cycle: **0 leaks, 0 bytes**.

This changes calculator behavior only—no card layout, styling, window sizing, or motion changed.


abue-ammar:
@kynnyhsap genuine good code. 1 small test failed and all passed. i am doing some minor polish and pushing it here in a bit

abue-ammar:
@kynnyhsap imma add more stuffs and make it more polished. thanks a lot

abue-ammar:
Pushed `0951471` — the two unit semantics are corrected, so #70 now merges self-consistent.

## Answer in the last unit typed

`+` / `-` convert the **left** side into the right operand's unit instead of the reverse, matching Raycast:

| Query | Before | Now |
| --- | --- | --- |
| `5feet + 1m` | `8.2808 ft` | **`2.524 m`** |
| `10kg + 500g` | `10.5 kg` | `10,500 g` |
| `500g + 1kg` | `1,500 g` | `1.5 kg` |
| `1kg + 500g + 2lb` | — | `5.306933933 lb` (left-associative, so the last unit wins) |

A `to` / `in` suffix still overrides it, so `10kg + 500g to lb` is unchanged.

**Composite notation is exempt.** `5 feet 3 inches` and `1hr 30min` are one quantity, not a sum, so they keep answering in the leading unit (`5.25 ft`, `1.5 hr`). `peekBinary` already separated the two cases — it reports `consumesToken: false` for the invisible `+` between adjacent quantities — so `addOrSubtract` keys the unit choice off that existing flag rather than inventing a second signal.

## A bare number takes the unit it's written against

`5kg+5` → `10 kg`, `10kg + 5` → `15 kg`, `$10 + 5` → `15.00 USD`. This also resolves the review finding about half-typed units from the other direction: instead of choosing between an error card and a blank one, the intermediate state is simply *right*.

Under adjacency it still returns nil, because there a bare trailing number is a unit mid-typing — `1hr 30` is one keystroke short of `1hr 30min`, and `31 hr` is a worse answer than none. Typing progressions for `10kg + 5kg`, `5feet + 1m`, `5 feet 3 inches` and `$10 + €5` are now card-flicker-free end to end.

## One more consistency fix

An expression containing an operator no longer falls through to the bare auto-conversion path, so `2 * 5kg` is `10 kg` and matches `5kg * 2`. Previously it returned `22.05 lb`. Only a bare quantity (`50cm`, `1m`) auto-converts now. This flips the pre-existing `2*3 kg` assertion from `13.22773573 lb` to `6 kg`, deliberately.

## Verification

- **426 assertions pass** (up from 415). 28 existing expectations were flipped by the semantics change; every one was verified by running the harness, not by hand-deriving.
- Differential probe of ~120 queries before/after: the only changes are the intended ones.
- Consent `.off` across every currency shape: still no card.
- fuzz / ranking / clipboard / emoji harnesses green; clean Debug build, no warnings; `xcodegen generate` leaves `project.pbxproj` untouched.

## One thing to sanity-check in the app

I applied last-unit-wins to **currency too**, so there's one rule rather than two: `$10 + €5` is now `14.20 EUR` (was `15.43 USD`). If "I'm adding to my dollars" reads better for money, it's a one-line revert in `addOrSubtract` — say the word.

PR #79, opened 2026-07-29
fix: support macOS 15 deployment and invalidate stale clipboard captures

zuston:
Thank you for your great work! I love Tinycast, but it didn’t support my Mac running macOS 15. This PR adds macOS 15 support with the help of Codex. I’m new to Swift, so any feedback or suggestions would be greatly appreciated.


suiramdev:
Nice PR! I think older macOS versions should remain supported, and this expectation could be clarified in `AGENTS.md` or `CONTRIBUTING.md`. It may be difficult to verify that every feature works on older versions, but AI should be able to use the relevant documentation to account for compatibility.

Ideally, compatibility would be confirmed in the PR description, though this could remain optional. Where necessary, we could gate certain features based on the macOS version.

abue-ammar:
thanks for the pr! your  PR helped. i did a different  approach. we will support macos 15 in a complete  separate  branch via patch file. 1 agent skill can generate  necessary patch verify it and merge branch and generate  release. here's  the branch . @suiramdev you can  review  it https://github.com/abue-ammar/tinycast/tree/compat/macos15

zuston:
thanks @abue-ammar for your quick reply. Do you mind release this in the download page?

PR #89, opened 2026-07-29
Support implicit multiplication before parens, constants, and functions

### LINKED ISSUE #84: [Feature]: Support implicit multiplication before parens, constants and functions
### What should it do?

Ordinary maths notation drops the `×` between a number and a bracket, a constant, or a function call. Tinycast rejects all of those outright:

| input | today | with an explicit `*` |
|---|---|---|
| `4(2+3)` | no card | `4*(2+3)` → `20` |
| `(2+3)(2+3)` | no card | `(2+3)*(2+3)` → `25` |
| `2pi` / `2π` | no card | `2*pi` → `6.283185307` |
| `2sqrt(9)` | no card | `2*sqrt(9)` → `6` |

Implicit multiplication should be treated as a `*` at the same binding power, so `2(3+1)+1` → `9`.

Worth scoping narrowly: this must not disturb the bare-unit path, where a number followed by a word already means something else entirely (`10km to mi` is a conversion, not `10 * km`). Restricting it to `(`, the known constants (`pi`, `π`, `e`) and the known function names keeps the two features from ever meeting — none of those names overlap a unit or currency name.

This is how the expression is written on paper and in every textbook, so it's what people type first — and the failure mode is silence, not an error, which reads as the calculator being broken rather than picky. Raycast's calculator accepts it.

It's also the cheapest kind of win: no new syntax to learn or document, no new results — every input it unlocks already has a working spelling today, just a more verbose one.

mathieusouflis:
Closes #84.

## Bug

Ordinary maths notation drops the `×` between a number and a bracket. None of these produce a card today:

| input | today | with explicit `*` |
|---|---|---|
| `4(2+3)` | no card | `4*(2+3)` → 20 |
| `(2+3)(2+3)` | no card | `(2+3)*(2+3)` → 25 |
| `2pi` / `2π` | no card | `2*pi` → 6.283185307 |
| `2sqrt(9)` | no card | `2*sqrt(9)` → 6 |

`CalcParser`'s precedence-climbing loop only continues past an operand when it recognizes an explicit binary operator next (`peekBinary()`); it returns nil for `(` or an identifier, so the loop stops, `isAtEnd` is false, and the whole expression is rejected.

## Fix

`CalcParser.swift` — `parseExpression`'s loop gets a second check alongside the explicit-operator one: if the next token can *only* be the start of another operand here (`(`, or a name in `constants`/`functions`), treat it as an implicit `*` at the same binding power (20/21, same as explicit `*`) — but, unlike a real operator, don't consume a token first, since there isn't one to consume.

**Why this can't collide with units or currency:** the check is narrowly scoped to `CalcParser.constants` (`pi`, `π`, `e`) and `CalcParser.functions` (`sqrt`, `log`, `ln`, `sin`, `cos`, `tan`, `abs`, `floor`, `ceil`, `round`) — none of which overlap any unit or currency name in `CalcUnits`/`CalcCurrency`. So `10km` still resolves through the existing bare-unit-conversion path untouched; implicit multiplication never even gets a chance to look at it.

## Tests

Added to `Tools/calc-test.swift`: the four reported cases, a precedence check (`2(3+1)+1` → 9, confirming implicit `*` binds tighter than `+`), and a regression pin on `10km to mi` to document the "never touches units" guarantee.

- `Tools/calc-test.swift` — **255 passed, 0 failed**

## Memory

Debug build, idle RSS sampled over ~15 s after launch: **79.0 MB**, against **79.6 MB** for `main` built from the same base — unchanged within noise. No separate peak figure: the change is a branch inside an existing parse loop, allocating nothing and retaining nothing.

## Notes

Non-visual change — no UI. The one thing worth a second look in review is the scoping argument above: the guarantee that implicit `*` never swallows a unit rests entirely on `constants`/`functions` being disjoint from the unit and currency tables. If a future unit is ever named `e` or `abs`, that assumption needs revisiting — the `10km to mi` test case exists to fail loudly if it ever stops holding.

### DIFF
--- a/Tinycast/Core/Calculator/CalcParser.swift
+++ b/Tinycast/Core/Calculator/CalcParser.swift
@@ -2,7 +2,7 @@ import Foundation
 
 enum CalcToken: Equatable, Sendable {
     case number(Double)
-    /// An attached thousands suffix (`10k` / `2.5K`), kept distinct so a lone shorthand still earns a calculator card.
+    /// A number written in shorthand (`10k` / `2.5K`, `1e5`), kept distinct so a lone one still earns a calculator card.
     case compactNumber(Double)
     /// Radix-prefixed integer literal (0xff / 0b1010 / 0o777), kept exact for base conversion.
     case intLiteral(UInt64, radix: Int)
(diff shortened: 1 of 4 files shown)


suiramdev:
LGTM! Keep up with these bug fixes, good work :)

PR #90, opened 2026-07-29
Add a modulo operator on "%"

### LINKED ISSUE #85: [Feature]: Modulo operator in the inline calculator
### What should it do?

There's no way to take a remainder in the calculator. `%` only ever means percent, so a modulo expression produces no card at all — not `1`, not `10/100 * 3`, nothing:

| input | today | expected |
|---|---|---|
| `10 % 3` | no card | `1` |
| `17 % 5` | no card | `2` |
| `10 mod 3` | no card (`mod` isn't recognized) | `1` |
| `2 + 10 % 3` | no card | `3` (binds tighter than `+`) |

The two meanings of `%` can be told apart by what follows it: if the next token can only start another operand (a number, `(`, a constant, a function), it's modulo; otherwise it's the existing percent postfix. That keeps every percent case working unchanged, since `20% of 450`, `450 + 20%` and a bare `20%` are all followed by something that *can't* start a value — a keyword, an operator, or end of input.

Remainder is a standard calculator operation, and `%` is its spelling in C, JavaScript, Go, Rust and friends — the same audience that already types `**` and `0xff` into this box. Accepting `mod` as a keyword alias covers the people who'd expect the spreadsheet/Python-ish spelling instead.

It's one operator on an existing precedence level, with no new UI and no new result card shape.

mathieusouflis:
Closes #85.

## Bug

`%` only ever means percent in `CalcParser`. `10 % 3` produces no card at all — not `1`, not `10/100 * 3`, nothing — because the postfix loop unconditionally consumes `%` as "this value is a percent," leaving the trailing `3` with nowhere to attach:

```
10 % 3   -> no card
17 % 5   -> no card
10 mod 3 -> no card ("mod" isn't recognized at all)
```

## Fix

Disambiguate by what comes right after `%`. `parseOperand`'s postfix loop now peeks at the next token before consuming `%`: if it can only be the start of another operand (a number, a signed value, `(...)`, a constant, or a function — i.e. a *value* follows), the loop leaves `%` untouched instead of eating it as percent. `peekBinary` picks it up from there as a binary modulo operator at the same precedence as `*`/`/` (a common convention). `apply` computes it as `truncatingRemainder` (C `fmod` semantics — sign follows the dividend). `mod` is accepted as a keyword alias for the same operator.

`10 % 3` → `1`, `17 % 5` → `2`, `10 mod 3` → `1`, and `2 + 10 % 3` → `3` (modulo binds tighter than `+`, same as `*`/`/`).

**Why the existing percent cases are untouched:** `20% of 450`, `450 + 20%`, and bare `20%` all have `%` followed by something that *can't* start an operand — `of` (a keyword, not a value), an operator, or end-of-input — so the disambiguation check falls through to the existing percent path unchanged.

## Tests

Added to `Tools/calc-test.swift`: the three reported cases plus a precedence check. I didn't touch the existing percent assertions since they're the regression proof that the two meanings don't collide.

- `Tools/calc-test.swift` — **252 passed, 0 failed**

## Memory

Debug build, idle RSS sampled over ~15 s after launch: **79.2 MB**, against **79.6 MB** for `main` built from the same base — unchanged within noise. No separate peak figure: the change is a lookahead in an existing parse loop, allocating nothing and retaining nothing.

## Tradeoff, flagged deliberately

`%` is now context-dependent, which is the part worth arguing about in review. I picked "a value follows ⇒ modulo" because it's the only reading that makes both meanings reachable without a new symbol, and because every existing percent form happens to be followed by a non-value. But it does mean `%` can't be read purely locally any more. The alternative — modulo only under `mod`, leaving `%` as percent forever — is simpler to explain and I'm happy to cut it back to that if you'd rather not have the overload.

### DIFF
--- a/Tinycast/Core/Calculator/CalcParser.swift
+++ b/Tinycast/Core/Calculator/CalcParser.swift
@@ -277,9 +277,8 @@ private struct Parser {
         switch current {
         case .op(let op) where op == "+" || op == "-": return (op, 10, 11)
         case .op(let op) where op == "*" || op == "/": return (op, Self.mulBP, Self.mulBP + 1)
-        // Reached only when the postfix loop declined "%" as percent; "*"/"/" precedence is the common convention.
-        case .op("%"): return ("%", Self.mulBP, Self.mulBP + 1)
         case .ident("of"): return ("*", Self.mulBP, Self.mulBP + 1)
+        // Spelled-out only: "%" is already percent, and "20% - 5" gives no local signal to tell the two apart.
         case .ident("mod"): return ("%", Self.mulBP, Self.mulBP + 1)
         case .op("^"): return ("^", 30, 30)  // right-associative: 2^3^2 = 512
         default: return nil
@@ -319,8 +318,6 @@ private struct Parser {
                 value = Value(value: fact)
             case .op("%"):
                 guard !value.isPercent else { return nil }
-                // Leave it for peekBinary rather than consuming it as percent.
-                if percentIsModulo() { break loop }
                 value.isPercent = true
             case .ident("deg"):
                 guard !value.isPercent else { return nil }
@@ -331,17 +328,6 @@ private struct Parser {
             pos += 1
         }
         return value
-    }
-
-    /// Only a value can follow modulo, so `of`, an operator, `)` and end-of-input all keep the percent reading.
-    private func percentIsModulo() -> Bool {
-        switch pos + 1 < tokens.count ? tokens[pos + 1] : nil {
-        case .number, .compactNumber, .intLiteral: return true
-        case .op("-"), .op("+"), .op("("): return true
-        case .ident(let name):
-            return CalcParser.constants[name] != nil || CalcParser.functions[name] != nil
-        default: return false
-        }
     }
 
     private mutating func parsePrefix() -> Value? {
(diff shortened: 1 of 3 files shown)


suiramdev:
LGTM!

abue-ammar:
Rebased onto `main` (conflicted with #89's `Self.mulBP` constants) and pushed a follow-up that cuts `%` back to percent-only, keeping `mod` as the operator — the alternative you flagged in the description.

The lookahead premise didn't hold: `+`, `-`, `(` and constants all legitimately follow a *percent*, so "a value follows ⇒ modulo" caught existing expressions. Verified against `b079de6`:

| input | main | with `%` overloaded |
|---|---|---|
| `450 + 20% - 5` | 535 | 450 |
| `50% + 50%` | 0.75 | 0 |
| `7 % pi` | 0.2199 | 0.7168 |

`20% - 5` has no local signal separating the two readings, so it isn't fixable by narrowing the lookahead. `mod` costs one keyword and stays unambiguous; `truncatingRemainder` and the `*`/`/` precedence are yours, unchanged. Added `-10 mod 3`, `10 mod 0` and a `450 + 20% - 5` regression test, plus a docs note.

465 passed / 0 failed, Debug build clean. Thanks for flagging the tradeoff explicitly in the description — made this a quick call.

PR #82, opened 2026-07-29
Add custom commands

junderhill:
This PR add's custom commands that can be configured in settings and then run from the launcher. https://github.com/abue-ammar/tinycast/issues/41

## Related issue

Closes #41

## Memory footprint

Measured on Apple Silicon running macOS 26.5.1 with Xcode 26.6. Measurements use `footprint`’s `phys_footprint` after one warm-up cycle and five repeated palette open/search/close cycles.

| Step                    | Memory   |
| ----------------------- | -------- |
| Idle after launch       | 42.37 MB |
| Palette open            | 42.62 MB |
| Feature in use (peak)   | 43.04 MB |
| Palette closed, settled | 42.60 MB |

Settled memory was within 0.23 MB of the warmed baseline, with no continuing growth across repeated cycles. The initial cold footprint was 39.54 MB.

Leak-tested: Yes. Apple’s `leaks` tool found no Tinycast-owned leaked types or retain cycles. It reported 19 KB of system Accessibility/XPC allocations (`AXObserverCookie` and `NSXPCConnection`), which were also present in an independently launched control process.

## Drawbacks

- Commands run in a non-interactive login shell. Aliases defined only in `.zshrc`, interactive prompts and `sudo` are therefore not available by default. Users can explicitly invoke an interactive shell when required, for example `/bin/zsh -ic 'alias-name'`.
- Commands execute with the user’s account and permissions. Imported backups therefore warn before accepting executable command content.
- Standard input and output are disconnected deliberately, prioritising predictable execution over support for interactive commands.
- Only the user-facing command name is searchable; command text is intentionally not indexed.

## Manual Testing/Validation
* Created and ran a “Sleep Displays” command using /usr/bin/pmset displaysleepnow.
* Verified the command appears as a Custom Command in launcher results.
* Verified the launcher action is labelled “Run Command”.
* Repeated palette open, custom-command search, clear and close five times while measuring memory.
* Verified success, non-zero exit handling, bounded stderr capture, persistence and home-directory execution through the custom-command harness.

**Example**
Add Custom Command named "Sleep Displays", Command `/usr/bin/pmset displaysleepnow`

image

image

PR #97, opened 2026-07-29
Add sleep and lock screen commands

### LINKED ISSUE #96: [Feature]: Add Lock Screen and Sleep commands
## What should it do?

Add **Lock Screen** and **Sleep** as built-in launcher commands. They should behave like existing commands: searchable, hideable, favoritable, and assignable to global shortcuts.

The implementation should stay native and local: use public macOS APIs, hide the palette before dispatch, surface permission/system failures, and include command hotkeys in settings backup and restore.

## Why does it belong in Tinycast?

These are common launcher actions, and modern Mac keyboards with Touch ID no longer provide a reliable keyboard shortcut for full system sleep. Raycast users expect both commands, while the implementation is small and adds no dependency, network access, account, or background service.

chlins:
## Related issue

Closes #96

## What changed

- Adds `Lock Screen` and `Sleep` as native launcher commands.
- Makes command entries assignable to global shortcuts, including launch-time re-registration and settings backup/restore.
- Uses public macOS APIs: CoreGraphics posts the standard `⌃⌘Q` lock shortcut, while IOKit requests system sleep.
- Hides the palette before dispatch and distinguishes permission denial from system-action failure.
- Updates Accessibility copy in README, onboarding, and Permissions settings.

## Memory footprint

Measured with `footprint` on macOS 26 after an optimized arm64 build.

| Idle after launch       | 21 MB  |
| Palette open            | 21 MB  |
| Feature in use (peak)   | 21 MB  |
| Palette closed, settled | 20 MB  |

Leak-tested: 20 palette open/close cycles; settled back to 20 MB with a 21 MB peak.

## Demo

[Watch the side-by-side demo](https://github.com/chlins/tinycast/releases/download/pr-97-demo/pr-demo-before-after.mp4): existing stable build on the left, this branch on the right. Both search `lock`, then `sleep`, at the same window size.

## Drawbacks

Lock Screen needs Accessibility permission because macOS gates synthetic keyboard events. Tinycast requests access and reports permission denial separately from event-construction failure.

## Tests & validation

- `xcodegen generate` reproduces the committed project with no diff.
- Swift 6 strict-concurrency typecheck with warnings as errors passes for all app sources.
- All documented standalone harnesses pass: fuzzy matching, ranking, calculator (426), clipboard (23), and search scopes.
- New command harness passes 18 cases without invoking real lock or sleep actions.
- Search/display paths, command icons, and palette open/close behavior exercised by hand.
- Full optimized arm64 app compiled, ad-hoc signed, launched, and memory-tested.
- Full `xcodebuild` was not available because this machine has Command Line Tools but no full Xcode app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ CPU churn. Just SwiftUI + AppKit with zero dependencies. It's fast because there
 - **Clipboard history** — text and images, searchable, pasted back into the app you were using.
 - **Global hotkey** — one shortcut summons the palette from anywhere.
 - **Per-app hotkeys** — bind a key to an app; press it to toggle (focus/hide).
+- **System commands** — lock the screen or put the Mac to sleep from the launcher or a hotkey.
 
 ## Install
 
@@ -50,9 +51,9 @@ directly from Releases instead, clear it once: `xattr -dr com.apple.quarantine
 
 ## Permissions
 
-**Accessibility** — needed only so Tinycast can paste a clipboard item back into the app you
-came from. You're prompted the first time you paste; grant it in **System Settings → Privacy &
-Security → Accessibility**.
+**Accessibility** — needed so Tinycast can paste a clipboard item back into the app you came from
+and invoke **Lock Screen**. You're prompted the first time either action needs it; grant access in
+**System Settings → Privacy & Security → Accessibility**.
 
 ## Using it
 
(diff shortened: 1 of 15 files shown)


abue-ammar:
we probably do not need this. custom commands are being implemented here #82

abue-ammar:
closing it because  you  can achieve these using custom  command which we have  now as i mentioned  above. thanks for the pr

PR #102, opened 2026-07-30
Add Ctrl-N/P palette navigation

### LINKED ISSUE #101: [Feature]: Ctrl-N/Ctrl-P to move the selection
### What should it do?

Ctrl-N / Ctrl-P move the palette selection down / up, exactly like ↓ / ↑ — in the results list, the emoji grid, and an open Actions menu, with Ctrl-N also expanding the compact bar. Holding the chord repeats, like holding an arrow.

It's the standard Emacs-style navigation pair, and Raycast ships it by default (its manual lists Ctrl-N/P under list navigation). Keeps hands on the home row while filtering. Zero UI, zero settings, a strict alias of existing arrow behavior.

ahonn:
## Related issue

Closes #101

## What changed

⌃N / ⌃P now alias ↓ / ↑ everywhere the arrows already work — the results list, the emoji grid, an open Actions menu, and ⌃N expanding the compact bar — with held-chord repeat.

`PalettePanel.sendEvent` rewrites a bare-⌃ N/P keyDown into a synthetic arrow keyDown and dispatches that instead. The always-focused search field maps ⌃N/⌃P to `moveDown:`/`moveUp:` via the standard key bindings, and nothing guarantees a SwiftUI `onKeyPress` would see them first — rewriting at the window's single event entry point sidesteps that and keeps the arrows the only navigation path. Matched on the typed character so non-QWERTY layouts follow the letter; ⇧/⌘/⌥ variants stay with the field editor.

## Memory footprint

Measured on the Debug build with `footprint(1)` (`phys_footprint`), macOS 26, Apple Silicon.

| Idle after launch       | 20 MB  |
| Palette open            | 35 MB  |
| Feature in use (peak)   | 37 MB  |
| Palette closed, settled | 36 MB  |

Settled stays at the palette-open level by design — the panel is kept resident after first summon (`isReleasedWhenClosed = false`); a second open/close cycle returned to 36 MB with no growth.

Leak-tested: `leaks` on the live process after summon → navigate → dismiss — 0 leaks for 0 bytes.

## Drawbacks

⌃N/⌃P are Emacs caret bindings inside text fields; in the single-line search field `moveDown:`/`moveUp:` are no-ops, so claiming them loses nothing. ⌃F/⌃B (the other Emacs pair) are deliberately not taken — those really move the caret.

## Tests & validation

- Manual pass on the running app: ⌃N/⌃P navigation in the list / emoji grid / open Actions menu / compact bar, held-chord repeat; arrows, plain typing of n/p, and ⌃A/⌃E editing unaffected.
- Debug build clean, no new warnings. No `Tools/` harness covers window event routing (AppKit/view layer); no engine sources touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Security → Accessibility**.
 
 1. Open **Settings → General** and record a global shortcut to summon Tinycast.
 2. Press it anywhere → the palette floats in. Type to filter, **↵** to launch.
-3. **Tab** switches between Apps and Clipboard; **↑/↓** move, **Esc** dismisses.
+3. **Tab** switches between Apps and Clipboard; **↑/↓** (or **⌃P/⌃N**) move, **Esc** dismisses.
 4. **Settings → Shortcuts** — search an app or custom command and record a global shortcut.
 
 ## Building from source
--- a/Tinycast/Core/PalettePanel.swift
+++ b/Tinycast/Core/PalettePanel.swift
@@ -33,6 +33,16 @@ final class PalettePanel: NSPanel {
         case .keyDown: paletteViewModel?.hoverHighlightArmed = false
         default: break
         }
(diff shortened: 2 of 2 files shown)


abue-ammar:
it is now available in latest stable. 4 ecmas

PR #107, opened 2026-07-30
System commands: macOS actions in the launcher

### LINKED ISSUE #106: [Feature]: System commands, macOS actions in the launcher
### What should it do?

Add a **System Commands** category to the launcher, covering the macOS actions Raycast ships as System Actions, so locking the screen, emptying the Trash or toggling Bluetooth is one search away instead of a menu dive or a system shortcut you have to remember.

31 commands, grouped the way Raycast groups them:

| Group | Commands |
| --- | --- |
| Power and session | Lock Screen, Sleep, Sleep Displays, Restart, Shut Down, Log Out, Show Screen Saver |
| Audio and media | Play / Pause, Next Track, Previous Track, Toggle Mute, Turn Volume Up, Turn Volume Down, Set Volume, and presets at 0 / 25 / 50 / 75 / 100% |
| Display and appearance | Show Desktop, Toggle System Appearance, Toggle Stage Manager |
| Files and storage | Open Trash, Empty Trash, Eject All Disks, Toggle Hidden Files |
| Apps and notifications | Hide All Apps Except Frontmost, Unhide All Hidden Apps, Quit All Applications, Dismiss Notifications |
| Connectivity | Toggle Bluetooth |

How they behave:

- They get their own launcher section and their own Shortcuts settings tab, but ride the normal `AppEntry` path, so fuzzy search, favorites, visibility and learned ranking all work with no special cases.
- Restart, Shut Down, Log Out, Empty Trash and Quit All Applications confirm first, with Return bound to Cancel so a reflexive second press cannot fire them.
- Volume and mute show a small HUD, since macOS only draws its own for real media keys. Commands whose effect is invisible (Empty Trash, the toggles) confirm with a short toast naming the state they landed in.
- Nothing-to-do is reported as information rather than an error. Empty Trash on an empty Trash says so instead of surfacing Finder's error.
- No new global hotkey surface. Built-in commands do not have per-command hotkeys today and this does not change that.

**Quit All Applications already exists** as a built-in command. It moves into this category and keeps its persisted favorite, visibility and ranking key, so nobody loses a favorite.

Deliberately left out, to keep the set to things that earn their weight:

- Windows-only Raycast actions (Hibernate, Toggle HDR, Toggle Night Light).
- Raycast's three preference rows (warn before emptying Trash, warn before quitting, excluded applications). Those are settings, not commands.
- Quit All Applications Except Frontmost.

Tinycast already ships one of these (Quit All Applications), so the shape is established. This generalises an idea that is already in the app rather than bolting on a new subsystem, and it is the part of Raycast people reach for most after the app launcher itself.

The weight is small and mostly data: one Foundation-only catalog of names, symbols and confirmation policy, plus one runner holding the platform calls. It adds no background work, no timers, no network, and nothing runs unless the user activates a command.

Two things I want to flag honestly before any of this is worth your time.

**The process is inverted here, and that is on me.** A full implementation already exists and I am opening a draft PR alongside this issue so you can see what is actually being proposed rather than judge it from a description. It stays a draft until you green-light the idea. If the answer is no, or the set is too big, nothing is lost.

**Implementation risk.** Most commands use public API (AppKit, CoreAudio, `NSWorkspace`, `pmset`). Five cannot, because macOS exposes no public control:

| Command | Route | Risk |
| --- | --- | --- |
| Toggle Bluetooth | Private `IOBluetoothPreference*` symbols, resolved with `dlsym` | Highest. A private symbol can vanish in any macOS release. Fails closed if absent. |
| Dismiss Notifications | Accessibility traversal of Notification Center | High. Matches Accessibility subroles, not English labels, and refuses to press an arbitrary button. |
| Show Desktop | Apple's own bundled Mission Control helper | Medium |
| Toggle Stage Manager, Toggle Hidden Files | Undocumented `defaults` keys, written and read back | Medium. Refuses to write when the current value cannot be read. |

If you would rather not carry that risk, **Toggle Bluetooth is the last commit in the branch** and can be dropped without touching anything else. Same for Dismiss Notifications. I would rather cut them than have them rot.

mathieusouflis:
## Related issue

Closes #106

Opening as a **draft**: the issue is not agreed yet, so this is here to be looked at, not merged. It leaves draft only if you green-light the idea, and after the memory numbers below are filled in.

## What changed

Two things, in that order in the history.

**1. Tinycast presents its own dialogs.** `NSAlert` and `NSSlider` are gone from the app. A new `ModalWindowController` (owned by `AppCore`, like every other long-lived manager) is the single presenter for confirmations, failure reports, value prompts and the transient HUD, all drawn with `Theme` tokens on the palette's own surface recipe.

This part touches features outside system commands, which is worth calling out: the custom command confirmation and failure dialogs, the Quit All confirmation, and the settings import and export notices all move onto it. Behaviour is preserved, including Return being bound to Cancel on destructive dialogs.

Two consequences that are improvements rather than side effects:

- Presentation is `async` (`withCheckedContinuation`) instead of `runModal`, so there is no nested run loop. The presenter refuses a second dialog while one is up, which is what now stops a held hotkey stacking dialogs. The `isConfirmingCommand` re-entrancy flag is deleted.
- The panel is non-activating like the palette, so a dialog no longer pulls app focus off whatever you were in.

**2. 31 system commands.** `SystemCommand.swift` is a Foundation-only catalog (stable IDs, names, symbols, confirmation policy) covered by a new `Tools/system-command-test.swift` harness. `SystemCommandRunner` holds every platform call. `AppCore.perform` is the single execution funnel, so the confirmation gate cannot be bypassed from the palette, a favorite slot or the compact bar.

Non-obvious things in the diff:

- **Quit All Applications keeps its old entry ID** (`command:quit-all-apps`) even though it now lives in the new catalog, so existing favorites, visibility and learned ranking survive the move. There is a test for it.
- **Launcher order is load-bearing.** `AppIndex` publishes applications, System Settings, system commands, then commands, and `LauncherList` filters in that same order, so the visible rows stay identical to the flat selection index.
- **Empty Trash asks Finder for `count items of trash` before emptying.** Finder raises an error when told to empty an empty Trash. The count goes through Finder rather than reading `~/.Trash` directly because that folder is TCC-protected, so an unprivileged read fails in a way indistinguishable from "empty", which would silently skip a real empty.
- **Volume and mute fall back to the output's preferred stereo channels** when the device exposes no master element, which is common on HDMI, and Toggle Mute parks the level at zero when the device has no mute control at all.
- **Eject All Disks treats a sibling volume that the same physical eject already unmounted as done, not failed**, since one disk can publish several volumes.
- Automation and Bluetooth usage strings are added to `Info.plist`, each in the commit that first needs it.

### Commits

32 commits, each building clean and passing the harness: the modal surface, then the catalog and launcher category (which carries Lock Screen, because Swift will not allow an empty enum to declare a raw type), then one commit per remaining command. Bluetooth is deliberately last so it can be dropped on its own.

## Memory footprint

Not measured yet. This is a draft and the numbers land before it is marked ready.

## Demo

> **Required before this leaves draft:** side-by-side before/after video, same window size, same actions, per CONTRIBUTING.md. The tables below are supplementary and do not replace it.

**Video (before / after):**

### New surfaces

| Surface | Screenshot |
| --- | --- |
| System Commands section in the launcher | image |
| Shortcuts settings, System Commands tab | image |
| Set Volume modal with the slider | image |
| Volume HUD after Turn Volume Up | Screenshot 2026-07-30 at 10 49 44 AM |
| Success toast, Trash Emptied |  Screenshot 2026-07-30 at 10 50 02 AM |
| Destructive confirmation, Shut Down | image |

### Migrated dialogs

These already existed as `NSAlert`, so this is where before and after is a real comparison.

| Dialog | Before (NSAlert) | After (Tinycast modal) |
| --- | --- | --- |
| Custom command confirmation |  | image |
| Custom command failure, with the 127 hint |  | image |
| Quit All Applications confirmation |  | image |
| Settings Import fail |  | image |
| Executable import warning |  |  |

## Drawbacks

**Five commands do not ride public API, because macOS exposes no public control.** Each fails safe rather than guessing:

| Command | Route | How it fails |
| --- | --- | --- |
| Toggle Bluetooth | Private `IOBluetoothPreference*` symbols via `dlsym` | Resolved dynamically, fails closed if the symbols are gone. The setter returns `void`, so the getter is polled to confirm. |
| Dismiss Notifications | Accessibility traversal of Notification Center | Matches Accessibility subroles, not English labels, so it does not silently no-op on a localised system. Never presses an arbitrary button, because a notification's own rows (Reply, Open) also press. |
| Show Desktop | Apple's bundled Mission Control helper | Path existence and exit status are both checked. |
| Toggle Stage Manager, Toggle Hidden Files | Undocumented `defaults` keys | Refuses to write when the current value cannot be read, and reads back to confirm. |

Toggle Bluetooth is the highest risk and is the **last commit**, so it can be dropped without touching anything else. Dismiss Notifications is second and is self-contained too. I would rather cut either than have it rot.

**Other tradeoffs made on purpose:**

- No per-command global hotkeys. Built-in commands do not have them today, and adding them means touching hotkey persistence, settings, backup and conflict detection. Better as its own change.
- The modal refactor's blast radius reaches custom commands and backup. Splitting it out would have left one PR with a component nothing uses and another that could not build alone.
- Toggle System Appearance changes macOS but not Tinycast, which stays locked to dark per the invariant. That is intentional but could read as a bug.
- The dialog is centred on the display under the cursor, not on the palette. Worth a second opinion.

**Unsure about:** whether 31 commands is the right size for this app, or whether a smaller core set is truer to "no bloat". Happy to trim.

## Tests & validation

**Harnesses** (`docs/development.md` § Tests), all run on the final commit:

| Harness | Result |
| --- | --- |
| `system-command-test.swift` (new) | 72 passed, 0 failed |
| `fuzz-test.swift` | all passed |
| `ranking-test.swift` | all passed |
| `calc-test.swift` | 426 passed, 0 failed |
| `clipboard-test.swift` | 23/23 passed |
| `scopes-test.swift` | all passed |
| `emoji-test.swift` | passed, 2054 records |
| `custom-command-test.swift` | first 15 cases pass, then it stalls on my machine. It spawns a real interactive `zsh`, so this looks environment-specific and reproduces on `main` too. Flagging rather than hiding it. |

The new harness covers the catalog only, which is why `SystemCommand.swift` is Foundation-only: exact inventory, unique and namespaced entry IDs, entry-ID round trip, the Quit All key preservation, and that only the agreed commands require confirmation. It deliberately performs no side effects.

**Every one of the 32 commits builds clean** and passes the harness, verified commit by commit, not just at the tip.

**Exercised by hand:** launcher search and section ordering, the Shortcuts tab, favoriting a system command, the confirmation and cancel paths, Set Volume, the volume HUD, and Empty Trash on both an empty and a non-empty Trash.

# Memory
|Step | Memory |
|---|---|
| Idle after launch | 23mo |
| Palette open | 37.6mo |
| Feature in use (peak) | 60mo |
| Palette closed, settled | 37.6mo |

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,6 +110,13 @@ Never break these without an explicit task to do so.
 - **Clipboard writes stamp a private `internalType` marker** so the poller skips Tinycast's own writes.
 - **Hotkeys persist under legacy `KeyboardShortcuts_<name>` UserDefaults keys** (from the removed
   KeyboardShortcuts package) so old bindings survive. See [hotkeys.md](docs/hotkeys.md).
+- **Tinycast presents its own dialogs, never `NSAlert` / `NSSlider` / system popovers.** Every
+  confirmation, failure report, value prompt and transient readout goes through
+  `ModalWindowController` (owned by `AppCore`; reachable elsewhere via `AppCore.showNotice` /
+  `askConfirmation`). Presentation is `async`, so there is no nested run loop, and the presenter
+  refuses a second dialog while one is up that, not a flag, is what stops a held hotkey stacking
+  dialogs. **↵ belongs to Cancel on every destructive dialog.** See
+  [ui.md](docs/ui.md#modals--hud).
 - **Read [`docs/ui.md`](docs/ui.md) before any restyle or new view.** `Core/Theme.swift` is the single
   design-token source.
 - **`Core/EdgeDissolve.swift` and `Core/ThinScrollbar.swift` are off-limits.** Both are tuned by eye
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -76,7 +76,9 @@ Never break these without an explicit task to do so.
   confirmation gate lives in `AppCore` and not in the runner. All of `Core/Snippets/` compiles into
   `Tools/snippets-test.swift` (the harness globs the directory), so the model, Markdown serializer,
   template engine, repository and keyword policies stay Foundation-only, and the AppKit files there
-  keep their dependencies to what the harness can stub.
+  keep their dependencies to what the harness can stub. `Core/SystemCommand.swift` is also
+  Foundation-only for `Tools/system-command-test.swift`; platform effects belong in
+  `SystemCommandRunner`, while confirmation and failure UI remain in `AppCore`.
 - **`Tools/fuzz-test.swift` holds a COPY of `FuzzyMatch`** from `Core/AppIndex.swift`. Change the
   scoring in one, mirror it in the other, or the test is meaningless.
 - **`EmojiData.generated.swift` is emitted by `node Tools/gen-emoji.js` and
--- a/Tinycast.xcodeproj/project.pbxproj
+++ b/Tinycast.xcodeproj/project.pbxproj
@@ -38,6 +38,7 @@
 		5568D02AE095293A15A7F30D /* RightClick.swift in Sources */ = {isa = PBXBuildFile; fileRef = 93E9ACD504E016DFF6FE6C1B /* RightClick.swift */; };
 		57AE40420A6A5349AD092878 /* ShortcutRecorder.swift in Sources */ = {isa = PBXBuildFile; fileRef = 596ADD747B4546B1E3340094 /* ShortcutRecorder.swift */; };
 		5E49E070BDBB91A2829F5321 /* SnippetTextInjector.swift in Sources */ = {isa = PBXBuildFile; fileRef = 309908BAB8C19B67F4608DA9 /* SnippetTextInjector.swift */; };
+		62002F021B21EC91AB77ED4D /* ModalWindowController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B4C27209F650DAB6262F0D9 /* ModalWindowController.swift */; };
 		649961AA51BA4D87626C27D5 /* CalcFormatter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 09093BF10EA97BDCA45A241F /* CalcFormatter.swift */; };
 		664D2BD485F3C5DB3F90B3C6 /* CustomCommandsSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = DCBF70B1C646861FDF20ED2D /* CustomCommandsSettingsView.swift */; };
 		68FCC30764DD44ED72EFB6BA /* TinycastApp.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2B629EC3B716FEE881BB0075 /* TinycastApp.swift */; };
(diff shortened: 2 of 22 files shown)


abue-ammar:
@mathieusouflis this thing is very well cooked. soo damn loved it.

PR #109, opened 2026-07-30
Keep palette lists stationary while the selection moves

### LINKED ISSUE #100: [Bug]: Keyboard selection scrolls before reaching the visible edge

abue-ammar:
## Related issue

Closes #100. Alternative to #55.

## What changed

Arrow keys recentered every palette list on a `scrollToken` UUID, so the viewport jumped even when the
selected row was already visible. The emoji grid already solved this on `main` with a reset/follow
scroll intent, so this generalises that one mechanism to all four modes rather than adding a second,
parallel one.

- **`ScrollIntent`** (`Core/ScrollIntent.swift`) replaces both `scrollToken` and `EmojiScrollIntent`.
  `.follow` is a minimal scroll-to-visible (nil anchor), so a list stays stationary while the
  selection walks across it and only advances a row at the viewport edges; `.top` restores the content
  origin. Modes never coexist, so `RootPaletteView` keeps one intent state instead of two.
- **`scrollOriginAnchor()`** marks content offset 0 with a zero-height overlay, which makes `.top`
  exact. Scrolling to the first row instead leaves the content's top padding hidden under the header —
  the emoji grid's reset was off by `Theme.Spacing.xs` for that reason, and this fixes it too.
- Landing on flat index 0 restores the origin so that row's section header comes back into view.
- The clipboard follow-the-moved-row handler (fresh capture, promote-on-paste, pin) now follows the
  selection instead of recentering it.
- `AGENTS.md`: `Core/EdgeDissolve.swift` and `Core/ThinScrollbar.swift` are now marked off-limits.
  Both are tuned by eye against the floating bars, so any edit there is a visual regression.

## Notes on approach

The anchor is an overlay applied *after* the content padding, not an extra child of the `ScrollView` —
a second child picks up the implicit stack's default spacing and pushes the whole list down. Nothing
about layout or laziness changes: no wrapper container, no reindentation of the row builders.

No AppKit involved. `NSClipView` origin math, run-loop retry passes and post-layout `Task.yield()`
hops were all considered and rejected: the scroll has to go through SwiftUI's own scroll state, or
`onScrollGeometryChange` reports a stale offset and the edge dissolve keeps painting its scrolled
gradient while the list is actually at rest.

`EdgeDissolve.swift` and `ThinScrollbar.swift` are untouched — zero lines. The dissolve is unchanged
by design.

Net +99/−60 across 8 files, and it deletes more mechanism than it adds.

## Tests & validation

- Clean unsigned Debug build via `xcodebuild`, no warnings.
- `swift Tools/fuzz-test.swift` — ALL PASSED
- `Tools/ranking-test.swift` — ALL PASSED
- `Tools/calc-test.swift` — 426 passed, 0 failed
- `Tools/clipboard-test.swift` — 23/23 passed
- `Tools/scopes-test.swift` — ALL PASSED
- `Tools/emoji-test.swift` — all checks passed (2,054 records)
- `Tools/custom-command-test.swift` — ALL PASSED

Manual sign-off still wanted on: launcher/clipboard/calculator-history arrow navigation to both ends,
query change after scrolling (origin restored with the header inset intact), and the emoji grid, whose
handler shape is unchanged apart from the exact origin.

## Summary by CodeRabbit

* **New Features**
  * Improved keyboard navigation across all palette modes with consistent, minimal “scroll-to-visible” following behavior.
  * Added reliable “scroll to true top/origin” when resetting or returning to the first row.
  * Preserved the origin section header when following selection at the first item (without unnecessary repositioning).

* **Documentation**
  * Updated the UI documentation to reflect the unified scrolling behavior and the clarified rules for `.follow` vs `.top`.


matijrn:
Thanks — I tested the merged #109 build at `bc633f6` and found one compact-mode regression:

1. Enable compact mode.
2. Open the launcher.
3. Type the first character (for example, `w`).
4. As the palette expands, the results list starts underneath the header, so the first row overlaps the search field.

This appears to happen because the `.top` intent is emitted on the query change while the results list is still unmounted, and the compact-to-expanded change only resizes the panel without reasserting the origin after the expanded layout mounts. During #55 testing, this exact case required a post-expansion top reassertion to remain stable.

Could you take a look? I have a screenshot of the reproduced frame if useful.

PR #110, opened 2026-07-30
Add app uninstall with leftover file selection

### LINKED ISSUE #50: [Feature Request] Quick uninstall
I mean this raycast feature where it kinda cleans whatever the app would left behind :)

Image

cpreston321:
## Related issue

Closes #50

## What changed

An application result's ⌘K Actions menu ends with **Uninstall Application…** (also **⌃⌫** on the row). It opens a new palette sub-screen listing the app bundle plus the support files it left behind — each row a checkbox with its size, everything checked, filterable by the header field and sortable by path / size / name.

Two removal paths, pairing the way Finder's do:

- **Uninstall Application** (`↵`, the footer default) moves the checked paths to the **Trash**, so it needs no second confirmation: the list *is* the confirmation.
- **Permanently Delete…** (`⇧⌘⌫`, Finder's own chord) unlinks instead, and is the one action that puts up an alert first, with `↵` on **Cancel** so a reflexive second Return can't erase anything.

The palette stays open across the whole flow (`UninstallPhase`): the list, a progress screen naming what is being removed, then a summary — how many items went, how much was reclaimed, whether it's recoverable from the Trash, and any paths that couldn't be removed with the two reasons why. Finishing returns to the root search rather than closing the palette. A dismissal mid-flow keeps the session (pop-to-root exempts it, and the generic summon restores it), so a click-away never discards a scan or a set of checkboxes.

**Non-obvious bits in the diff:**

- `AppLeftovers` (`Core/AppLeftovers.swift`) is Foundation-only and pure — home directory and the sibling-install flag are injected — so `Tools/leftovers-test.swift` compiles it standalone and drives it against a throwaway home. The LaunchServices lookup and the trashing live in `UninstallSession` for exactly that reason.
- Discovery is deliberately narrow: only the `~/Library` roots in `allowedRoots`, keyed on an exact bundle id (plus dotted-boundary prefixes for helper/extension ids, and the team-id prefix on group containers) or on the app name and its no-space / hyphen / underscore variants. Every result is re-verified to be a *proper* descendant of a root — never a root itself, never `$HOME`, never reachable via `..`.
- A bundle id that isn't plain reverse-DNS is refused rather than sanitized, and a name under three characters matches nothing.
- When another registered install shares the bundle id (`Xcode.app` beside `Xcode-beta.app`), *every* keyed path is ambiguous, so all of them are dropped and only the bundle is offered.
- `canUninstall` refuses `/System`, `/usr`, every `com.apple.*` id (which also covers the cryptex-delivered apps) and every Tinycast channel. It is the one rule both the menu row and the ⌃⌫ chord read, so the advertised chord can't drift from the menu.
- **Nothing system-level is discovered or removed** — no `/Library/LaunchDaemons`, privileged helpers or system extensions. Those need an administrator, and leaving them is the fail-safe direction.
- The scan is **two-stage**: discovery lands at once, then sizes fill in per row. A size is a full tree walk — cold, a 3.5 GB bundle measured **32 s** here against 116 ms warm — so making the list wait on it read as a hang.
- Symlinks are never sized through or resolved for measurement: a cask's `/Applications` alias points at the bundle that already has its own row. The alias gets its own row so it doesn't dangle, and removal targets the resolved bundle.
- Removal quits the app first (graceful, bounded 3 s wait) **only when the bundle itself is checked** — unchecking it to clear caches alone shouldn't force-quit an app you're using.
- `filtered(_:)`, not `items`, is what the selection and every activation address. New code resolving "the highlighted row" must go through the filtered list.
- Two new `Theme` tokens (`Size.checkbox` 16, `Radius.checkbox` 4) — justified in `docs/ui.md` along with why the checkbox is drawn from tokens instead of `checkmark.square` (an SF checkbox glyph reads as a near-invisible hairline on this surface).
- The sort control adds a third popover menu, and the only one anchored to a **top** corner. Mutual exclusion between the three now lives in the `toggle…`/`open…` helpers rather than three `onChange` handlers, and `openActions()` refuses to open when a screen has no actions (an empty panel would swallow the keyboard invisibly).

## Demo

Caveat on the clip: it shows the **after** state only. CONTRIBUTING asks for side-by-side before/after, same window size, same actions — that's still owed.

https://github.com/user-attachments/assets/9e4cd9cf-fdba-4a1b-8a92-990394298176

## Drawbacks

- **A literal space can't be typed into the filter field.** Bare space is intercepted as the row toggle before the field editor sees it. Substring matching makes it a non-issue in practice, but it is a real asymmetry with the other screens.
- **User-level only.** System paths and privileged helpers are never touched, so a few apps will leave a `/Library` remnant behind. Deliberate: removing those needs an administrator.
- **Homebrew casks aren't special-cased.** Trashing a cask's payload leaves brew's metadata thinking it's installed. The alias is handled, the `brew uninstall` handoff isn't.
- **Sizes appear progressively**, so the status-line total climbs for a moment on a cold cache instead of being right immediately.
- **Permanent delete has no undo** — hence the alert, the `↵`-on-Cancel default, and the Trash being the default path.
- Row order comes from the sort control rather than sections, which makes the uninstall list the one palette list without section headers.

## Tests & validation

New harness `Tools/leftovers-test.swift` (55 checks) compiles the real `AppLeftovers.swift` and drives it against a throwaway home, so a run can never see — let alone return — a path in the developer's own Library. It covers eligibility, bundle-id and name-keyed discovery, the non-matches, every guard (refused bundle ids, short names, the surviving-sibling case, no result ever being a `~/Library` root), sizing, the symlink rules, and removal — the permanent path erases fixtures for real, one failure doesn't strand the items after it, and the trash path runs on a UUID-named fixture the harness then deletes. The one assertion needing Full Disk Access reports `SKIP`, not `FAIL`.

```sh
swiftc -swift-version 6 Tinycast/Core/AppLeftovers.swift Tools/leftovers-test.swift \
    -o /tmp/leftovers-test && /tmp/leftovers-test
```

Every existing harness still passes: fuzz, ranking, calc (426), clipboard (23/23), scopes, emoji (2054 records), custom-command.

By hand: discovery probed read-only against all of `/Applications` on this machine — for Affinity it returns exactly the set in the issue's reference screenshot plus two QuickLook containers, and apps that were never launched correctly return nothing. A disposable fixture app (fake bundle id, 4 MB payload, five real support paths) was used for the end-to-end run so no real app was at risk. Build is clean and warning-free; `xcodegen generate` re-run and the project committed.

Docs: new `docs/uninstall.md`, plus `ui.md` (new tokens, the row grammar, the third menu anchor), `launcher.md`, `palette.md`, `development.md`, `AGENTS.md` (a new Critical Invariant for the removal rules) and the README feature list.

## Summary by CodeRabbit

* **New Features**
  * Added an app uninstall workflow from the launcher, including selectable leftover files and folders.
  * Added sorting, size estimates, progress feedback, removal summaries, and Finder reveal actions.
  * Supports sending selected items to the Trash or permanently deleting them with confirmation.
  * Added safety checks to restrict removal to eligible user-level paths.

* **Documentation**
  * Added comprehensive uninstall guidance and updated feature, launcher, palette, UI, and development documentation.

* **Tests**
  * Added coverage for leftover discovery, safety rules, sizing, and removal behavior.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -40,8 +40,8 @@ Full detail: [`docs/architecture.md`](docs/architecture.md).
   `AuxWindowController`. SwiftUI `Settings` / `Window` scenes are deliberately avoided (unreliable for
   accessory apps).
 - **Subsystems:** [palette](docs/palette.md) · [launcher & fuzzy match](docs/launcher.md) ·
-  [calculator](docs/calculator.md) · [clipboard](docs/clipboard.md) · [emoji](docs/emoji.md) ·
-  [snippets](docs/snippets.md) · [window management](docs/window-management.md) ·
+  [uninstall](docs/uninstall.md) · [calculator](docs/calculator.md) · [clipboard](docs/clipboard.md) ·
+  [emoji](docs/emoji.md) · [snippets](docs/snippets.md) · [window management](docs/window-management.md) ·
   [hotkeys](docs/hotkeys.md) · [UI & design system](docs/ui.md).
 
 ## Critical Invariants
@@ -74,7 +74,11 @@ Never break these without an explicit task to do so.
   `Core/CustomCommand.swift` and `Core/ShellCommandRunner.swift` must likewise stay free of AppKit /
   SwiftUI (Foundation plus Combine for `ObservableObject` and Darwin for `mkstemp`) so
   `Tools/custom-command-test.swift` can compile them standalone — which is why the custom-command
-  confirmation gate lives in `AppCore` and not in the runner. All of `Core/Snippets/` compiles into
+  confirmation gate lives in `AppCore` and not in the runner. `Core/AppLeftovers.swift` is the same
+  deal for `Tools/leftovers-test.swift` — Foundation only and pure, with the home directory and the
+  sibling-install flag injected, which is why the LaunchServices lookup and the trashing orchestration
+  live in `UninstallSession` (the `trashItem` / `removeItem` primitive itself stays in `AppLeftovers`,
+  so the harness can drive it). All of `Core/Snippets/` compiles into
   `Tools/snippets-test.swift` (the harness globs the directory), so the model, Markdown serializer,
   template engine, repository and keyword policies stay Foundation-only, and the AppKit files there
   keep their dependencies to what the harness can stub. `Core/SystemCommand.swift` is also
@@ -122,6 +126,18 @@ Never break these without an explicit task to do so.
   via `Task.detached` / `nonisolated`. Keep that boundary. House idioms: `NotificationToken` (RAII) for
   block observers, `isolated deinit` for `ClipboardStore`'s SQLite teardown, decode raw Carbon / C
   pointers to plain values before crossing into actor code.
+- **Uninstall only ever moves user-level paths to the Trash.** `AppLeftovers` may return nothing that
+  isn't a proper descendant of one of its `allowedRoots` under `~/Library`, never a root itself; a
+  bundle id that isn't plain reverse-DNS is refused rather than sanitized; and a shared bundle id (a
+  surviving sibling install) drops *all* keyed paths. `canUninstall` refuses `/System`, `/usr`, every
+  `com.apple.*` id (`isProtectedVendor`) and every Tinycast channel — the action is not offered for those
+  at all. Within an eligible app, removability is decided **per row**: `isRemovable` checks that the
+  parent permits deletion and that no immutable / SIP-restricted flag is set, and a locked row can never
+  be checked or toggled. The default removal is
+  `trashItem` — the list ships with every removable row checked, so ↵ has to be recoverable; `removeItem`
+  is reachable only from the explicit Permanently Delete row (⇧⌘⌫), the one action in the flow that
+  confirms first. Nothing system-level (LaunchDaemons, privileged helpers, system extensions) is
+  discovered or removed. See [uninstall.md](docs/uninstall.md).
 - **Clipboard writes stamp a private `internalType` marker** so the poller skips Tinycast's own writes.
 - **Hotkeys persist under legacy `KeyboardShortcuts_<name>` UserDefaults keys** (from the removed
   KeyboardShortcuts package) so old bindings survive. See [hotkeys.md](docs/hotkeys.md).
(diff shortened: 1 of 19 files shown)


abue-ammar:
@cpreston321 hey nice work. i have some change requests. 

i dont want this:
image

instead we will consistent with all the views we have eg: app launch, clipboard, emoji, calculator history. 
it will always be this :
image

nextup: 
image
this is jumping up. it must be same aligned with : 
image
these category. we will strictly not allow any ui inconsistency and layout shift

abue-ammar:
@cpreston321 i have edit comment i have added more

cpreston321:
@abue-ammar my first initially thought was the issue for a 1-1 ui but I will make those adjustments now those make sense.

abue-ammar:
> @abue-ammar my first initially thought was the issue for a 1-1 ui but I will make those adjustments now those make sense.

take your time nothing to hurry. i want well optimized efficient codes. and also how are you handling protected apps and its data? make sure it behaves just like raycast. eg:
image

cpreston321:
@abue-ammar I will double check my logic here and compare what Raycast does thanks for the heads up.

abue-ammar:
> @abue-ammar I will double check my logic here and compare what Raycast does thanks for the heads up.

please rebase with main i can see lots of conflicts. make sure it doesn't break main

cpreston321:
@abue-ammar those changes should be good to go.

abue-ammar:
> @abue-ammar those changes should be good to go.

an updated video would be much appreciated bro. update PR summary if needed then i will start review

abue-ammar:
@cpreston321 quit app menu item disappear for running app. it is acting quirky 
image

it appears when i right click on 
image
another issue: ghostty is not running it appears quit menu
image

abue-ammar:
lots of thingss broken
image
it does not restore scrolled position
image
+ scroll freezes

abue-ammar:
also this view is missing:
image

abue-ammar:
@cpreston321 we dont ship half baked quirky stuffs. already lots of things broken. when ur are implementing a feature it MUST not break existing stuffs. at these rate this PR might not get merged.

cpreston321:
@abue-ammar i compared to Raycast and any protected apps don't even show uninstall in the action menu so that's what I did. It does not show for protected apps.

cpreston321:
> @abue-ammar i compared to Raycast and any protected apps don't even show uninstall in the action menu so that's what I did. It does not show for protected apps.

So I thought.. that is why it's not showing for protected apps. I will re-work my PR.

abue-ammar:
> @abue-ammar i compared to Raycast and any protected apps don't even show uninstall in the action menu so that's what I did. It does not show for protected apps.

already lots of things broken. your code must follow SOLID principal and there are apps where u can partially delete stuffs.  also you are implementing uninstaller. why would you need chnages in this file 
image
like this you/your AI have removed lots of stuffs that is important code!

cpreston321:
> > @abue-ammar i compared to Raycast and any protected apps don't even show uninstall in the action menu so that's what I did. It does not show for protected apps.
> 
> already lots of things broken. your code must follow SOLID principal and there are apps where u can partially delete stuffs. also you are implementing uninstaller. why would you need chnages in this file image like this you/your AI have removed lots of stuffs that is important code!

I reverted code locally, I ran into a weird issue when I went to the un-install screen then went back but I thought it was because I added the new screen which was not the case.

abue-ammar:
i am implementing it

PR #112, opened 2026-07-31
Support both Raycast v1 and v2 .rayconfig formats

### LINKED ISSUE #38: [BUG] Raycast import from .rayconfig doesn't work
Image

## Raycast Info
Image

thedavidweng:
Closes #38

## Summary
Raycast has two `.rayconfig` export formats in the wild:

- **v1 (current stable / Raycast 1.104.x):** a single AES-256-CBC ciphertext blob, key/IV derived from the passphrase with OpenSSL's EVP_BytesToKey (SHA-256 / MD5, no salt), then a short leading prefix and a gzip stream containing a JSON object under `builtin_package_*` providers.
- **v2 (upcoming / beta):** gzip → JSON envelope → hex ciphertext decrypted with AES-256-GCM under a scrypt(N=16384,r=8,p=1) key → gunzip → settings JSON.

`RaycastImport` now tries v1 first, then v2, and `parse` dispatches to the appropriate mapper based on the decrypted JSON shape. `Scrypt.swift` is preserved so v2 still works.

## Test plan
- [x] Build passes (`xcrun xcodebuild -project Tinycast.xcodeproj -scheme Tinycast -configuration Debug build CODE_SIGNING_ALLOWED=NO`).
- [x] Standalone `swiftc` script successfully decrypted the user-supplied v1 `.rayconfig` with the new AES-CBC flow and located the gzip stream at offset 16.
- [ ] v2 `.rayconfig` not available for regression testing.

### DIFF
--- a/Tinycast/Core/Backup/RaycastImport.swift
+++ b/Tinycast/Core/Backup/RaycastImport.swift
@@ -9,7 +9,7 @@ enum RaycastImportError: LocalizedError {
 
     var errorDescription: String? {
         switch self {
-        case .notRaycastFile: return "This doesn't look like a Raycast export (.rayconfig)."
+        case .notRaycastFile: return "This doesn\u{2019}t look like a Raycast export (.rayconfig)."
         case .incorrectPassphrase: return "Incorrect passphrase, or the file is corrupted."
         case .corrupt: return "The Raycast export could not be read."
         }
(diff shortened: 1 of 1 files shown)


huyixi:
I found a crash while testing Raycast v1 imports.\n\nextractJSONV1 passes a nonzero-index Data slice to Gunzip.decompress, but Gunzip uses zero-based offsets with subdata(in:), causing an out-of-bounds trap.\n\nFix and regression test:\nhttps://github.com/huyixi/tinycast/commit/4107731\n\nYou can apply it with:\n\ngit cherry-pick 4107731

abue-ammar:
@thedavidweng this PR doesn't follow our  PR issue template please follow it strictly. side by side before after video is encouraged if it's a visual change. and tested with proper tests cases

PR #122, opened 2026-07-31
Fix Toggle Bluetooth showing no icon in the launcher

### LINKED ISSUE #121: [Bug]: Toggle Bluetooth has no icon in the launcher
### What happened?

The Toggle Bluetooth system command shows no icon in the launcher, in the ⌘K Actions menu, and in Settings → Shortcuts → System Commands. Every other system command shows its symbol tile; this one is a blank rounded square.

Root cause: `SystemCommandCatalog.symbol(for:)` uses `"bluetooth"` as the SF Symbol name for `.toggleBluetooth`. That name does not exist in the SF Symbols catalog, Apple has never shipped a literal Bluetooth glyph, since the Bluetooth logo is a Bluetooth SIG trademark. `NSImage(systemSymbolName: "bluetooth", ...)` returns `nil`, and `IconCache.symbolIcon` draws nothing when the symbol lookup fails, leaving the tile empty instead of falling back to anything.

Expected: a recognizable icon in the tile, matching every other command.
Actual: blank tile.

0.1.0 · dev

### macOS version

26.5.2

mathieusouflis:
## Related issue

Closes #121

## What changed

`SystemCommandCatalog.symbol(for:)` used `"bluetooth"` as the SF Symbol name for Toggle Bluetooth. That name has never existed in the SF Symbols catalog, since Apple doesn't ship the Bluetooth logo as a symbol, it's a Bluetooth SIG trademark. `NSImage(systemSymbolName:)` returned `nil` for it, and `IconCache.symbolIcon` draws nothing when the lookup fails, so the tile was blank everywhere: the launcher row, the ⌘K Actions menu, and Settings → Shortcuts → System Commands.

Swapped it for `wave.3.right`, a fanning signal glyph. It's the same stand-in convention most Mac menu-bar utilities use for Bluetooth when they can't use the real logo, and it reads clearly at the small tile size the way every other command icon does.

## Memory footprint

Not applicable, this changes one string constant in a Foundation-only catalog, nothing that affects runtime memory.

## Demo

Static icon change, a before/after screenshot pair covers it better than a video would.

| Surface | Before | After |
| --- | --- | --- |
| | | 

## Drawbacks

`wave.3.right` isn't a literal Bluetooth pictograph, just the closest available stand-in. Open to a different symbol if you'd rather.

## Tests & validation

`system-command-test.swift`: 72 passed, 0 failed. Full Debug build succeeds.

### DIFF
--- a/Tinycast/Assets.xcassets/bluetooth.imageset/Contents.json
+++ b/Tinycast/Assets.xcassets/bluetooth.imageset/Contents.json
@@ -1,0 +1,16 @@
+{
+  "images" : [
+    {
+      "filename" : "bluetooth.svg",
+      "idiom" : "universal"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  },
+  "properties" : {
+    "preserves-vector-representation" : true,
+    "template-rendering-intent" : "template"
+  }
+}
--- a/Tinycast/Assets.xcassets/bluetooth.imageset/bluetooth.svg
+++ b/Tinycast/Assets.xcassets/bluetooth.imageset/bluetooth.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="m7 7 10 10-5 5V2l5 5L7 17"/></svg>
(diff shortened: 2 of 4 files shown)


mathieusouflis:
Should we let this or should i create a new icon who look like the bluetooth one ? @abue-ammar 


PS: I'll set the screens later but icon used is this one : 
Screenshot 2026-07-31 at 10 13 45 AM

abue-ammar:
> Should we let this or should i create a new icon who look like the bluetooth one ? @abue-ammar
> 
> PS: I'll set the screens later but icon used is this one : Screenshot 2026-07-31 at 10 13 45 AM

if try using svg assets i did for about tinycast in the settings you will custom svg. maybe you can get a Bluetooth icon that matches  SF symbols you can  put it? 
or this?
Screenshot_20260731_152450_Brave

abue-ammar (approved):

PR #137, opened 2026-08-01
feat: In-app updates with Sparkle, off by default

### LINKED ISSUE #136: [Feature]: Tinycast should be able to update itself
### What should it do?

Tinycast should be able to keep itself current without the user opening a terminal.

The capability, independent of how it's built:

- **Tell me a new version exists.** A "Check for Updates…" affordance that belongs in the app — the menu-bar menu and the launcher — plus an optional background check.
- **Install it.** Not just "open the release page": download the new build, verify it's really ours, replace the app and relaunch. A button that punts to a browser is a link, not an update.
- **Ship off.** Checking is a network call, so it follows the same rule as currency conversion: off out of the box, behind a Settings toggle whose consent sheet names the provider, the cadence, and exactly what leaves the machine. Off has to mean *no request at all*, not "a quiet one".
- **Tinycast's own dialogs.** Whatever does the work, the user must never see a foreign window. Confirmations through `ModalWindowController`, styling from the existing tokens.
- **Respect the channels.** Beta and stable are separate apps with separate bundle ids; neither may ever be offered the other's build. The dev channel must not be updatable at all.
- **Don't fight Homebrew.** Most people install via the cask. Whatever ships must leave `brew upgrade` correct rather than clobbering a self-updated app or silently downgrading it.

Today there is exactly one update path — `brew upgrade` — and it only reaches people who installed via the cask *and* who remember to run it. Anyone who took the DMG from Releases has no signal at all that a new version shipped; they find out by wandering back to GitHub. For a launcher you leave running for months, "open a terminal to discover you're three versions behind" is a real gap, and it's the kind of thing nobody files an issue about. They just drift, and then they're running an old build with bugs that were fixed.

It also compounds badly with how Tinycast is signed. Because there's no Apple Developer account, a stale install isn't just missing features — it's missing whatever the last release fixed about permissions, quarantine, or the signing identity itself. Updating is the mechanism by which those fixes actually land.

On the weight test: the honest comparison isn't "an updater vs nothing". It's "an updater vs writing the atomic bundle swap, the relaunch handoff, the signature validation and the quarantine handling ourselves, correctly, first time". That's the part where mistakes brick installs, so the question is mostly *who writes that code*, not *whether it exists*.

### Approaches worth weighing

Not asking for a specific one — this is the part to agree on before anyone writes code.

1. **A framework that already solves it** — [Sparkle](https://sparkle-project.org) being the obvious candidate. Biggest cost is ~3 MB of binary; biggest win is not owning the install path. Sparkle 2 lets you replace its entire UI via `SPUUserDriver`, so the "Tinycast's own dialogs" requirement survives.
2. **Notify only, install by hand** — poll the GitHub releases API, show "0.8.0 is available", open the release page or copy the `brew upgrade` line. Roughly 120 lines, no dependency, no new attack surface. Also: not really an update feature.
3. **Shell out to Homebrew** — a button that runs `brew upgrade --cask tinycast`. Tempting, but it only serves cask users, depends on undocumented Caskroom layout to even detect that case, and the cask's own `uninstall quit:` stanza will quit Tinycast mid-upgrade — i.e. kill the process that launched brew.
4. **Roll our own** — download, verify, swap, relaunch. Full control, and every one of those four steps is a way to brick someone's install.

### The objection to get out of the way first

"We're self-signed and not notarized, so nothing like this can work." I went and checked that against Sparkle's actual source rather than assuming, and it's wrong:

- `SUCodeSigningVerifier` lifts the **old** bundle's code-signing *designated requirement* — `identifier "…" and certificate leaf = H"…"` — and checks the downloaded bundle against it. That's a leaf-hash match against the app already installed, not a trust decision about a certificate chain. Gatekeeper is never consulted; `spctl` rejecting the bundle is irrelevant.
- `SUUpdateValidator` accepts an update on **either** a valid EdDSA signature **or** a matching code signature, so the stable `Tinycast Self-Signed` identity is sufficient on its own, and an EdDSA key gives a second independent strand.
- Sparkle strips `com.apple.quarantine` from the bundle it installs, so an in-app update needs none of the `xattr` dance a direct download does.
- macOS App Management (TCC) doesn't engage: that protection keys off Team ID, and ours is `not set`. No "wants to manage apps on your Mac" prompt.

So the constraint that looks fatal isn't. Whatever approach wins, being self-signed isn't the reason to rule one out.

### Things any approach has to answer

- Where the update manifest lives and how it's published per channel.
- What the "off" state actually guarantees, and what opting back out deletes.
- How the cask stops fighting a self-updated app (Homebrew's `auto_updates true` is the intended answer, and since Homebrew 6.0.0 it compares the installed bundle's `Info.plist` rather than skipping outright).
- Whether the signing identity becomes load-bearing in a new way, and what the recovery story is if a key is lost.

suiramdev:
## Related issue

Closes #136

## What changed

In-app updates via [Sparkle](https://sparkle-project.org) 2.9.4, pulled in as an SPM binary target and driven by a **custom `SPUUserDriver`**, so none of Sparkle's own AppKit windows ever appear — confirmations go through `AppCore.askConfirmation`, reports through `showNotice`, like everything else in the app.

- **`Core/Updates/UpdateStore.swift`** — `@MainActor`, owned by `AppCore` as `updates`, started from `start()`. Publishes one `State` that the Settings row renders.
- **`Core/Updates/UpdateUserDriver.swift`** — all 16 required `SPUUserDriver` methods. Progress is *published*, never shown: the app is an accessory, so a background check that finds something must not steal focus.
- **Settings → Miscellaneous** gains a Software Update card (auto-check toggle + consent sheet, version + Check Now), plus a menu-bar item and a launcher command. All three hidden when the channel has no feed.
- **`release.yml`** runs `generate_appcast` over the DMG it already builds and commits the channel's appcast under `website/public/`.
- **Docs** — new `docs/updates.md`; `signing.md`, `development.md`, `architecture.md`, `AGENTS.md` and `README.md` updated.

Three things in the diff are deliberate and won't be obvious:

1. **No `SUFeedURL` in `Info.plist`.** The feed comes from `SPUUpdaterDelegate.feedURLStringForUpdater:`, which wins over the plist and returns `nil` for `com.tinycast.app.dev`. There is no URL to fall back to, which makes the dev channel *provably* un-updatable rather than merely unconfigured.
2. **Two feeds, not `sparkle:channel`.** Sparkle's default channel is always allowed, so a beta host subscribed to a beta channel would still be offered the stable DMG — whose `Tinycast.app` / `com.tinycast.app` matches neither the beta host's filename nor its bundle id, and the install fails with "No suitable install is found". Separate feeds make the mismatch impossible.
3. **`SUEnableAutomaticChecks = NO` is load-bearing twice.** Its *presence* permanently suppresses Sparkle's own permission prompt, so only Tinycast's consent sheet ever asks; and it makes `startUpdater()` early-return before arming a timer. `SUHost` prefers the user default over the plist, so the toggle flips it live — one flag, not two, which is also why it stays out of `AppSettings` where a `SettingsBackup` import could have granted network access.

Why this works at all for a self-signed, non-notarized app — checked against Sparkle's source, not assumed — is written up in `docs/updates.md`: validation is a leaf-hash match against the **old** bundle's designated requirement, Gatekeeper is never consulted, Sparkle strips quarantine itself, and App Management (TCC) doesn't engage because there's no Team ID to protect.

## Memory footprint

Release build, same machine, identical key sequence on both branches, onboarding already completed on both.

| Step                    | Memory | `main` |
| ----------------------- | ------ | ------ |
| Idle after launch       | 20 MB  | 20 MB  |
| Palette open            | 34 MB  | 33 MB  |
| Feature in use (peak)   | 43 MB  | 39 MB  |
| Palette closed, settled | 38 MB  | 38 MB  |

Sparkle costs **nothing resident** while idle — `startUpdater()` early-returns and the framework's pages are file-backed. The 4 MB at peak is the check itself (URLSession + XML parse) and it releases. Settled sits above idle-at-launch on *both* branches; that's the app index and its cached icons, unchanged here.

Leak-tested: `leaks <pid>` after a full check cycle → **0 leaks for 0 total leaked bytes** (169075 nodes / 34940 KB malloced, 42.2 MB footprint).

## Demo

**Outstanding — this is why the PR is a draft.** Three before/after clips still to record, same window size:

1. Settings → Miscellaneous: the Software Update card, toggling it on to show the consent sheet.
2. Palette: typing "check for updates" and pressing ↵.
3. The resulting dialog — verified on screen already (Tinycast's own dark `.error` modal, one OK button, no `NSAlert`), just not captured as video.

## Drawbacks

- **+2.9 MB on disk** (4.2 → 7.1 MB). Real, and the largest single cost here; zero resident cost, per the table. Stripping the XPC services would recover 0.4 MB but needs a re-sign script — not worth the moving part. `README.md`'s "around 3 MB" and "zero dependencies" are corrected in this branch rather than left false.
- **Two keys are now load-bearing, not one.** Sparkle's Developer-ID key-rotation fallback hardcodes `anchor apple generic` and can never fire for us, so the `.p12` and the EdDSA key each independently keep the channel alive — and losing *both* strands every existing install. `signing.md` documents the four recovery cases; back up the login keychain.
- **The release workflow now pushes to `main`.** It replays only the one appcast file onto `origin/main` rather than pushing `HEAD` (the workflow can be dispatched from any ref), but it is still CI writing to the default branch. The alternative was a second publishing mechanism just for the feed.
- **No `Tools/` harness.** Sparkle is a binary framework and both new types are AppKit-bound, so there's nothing a Foundation-only harness could compile. Called out in `docs/updates.md` so the omission reads as deliberate rather than forgotten.
- **The install leg is unexercised.** Everything up to the HTTP response is verified end-to-end; download → verify → swap → relaunch can't run until an appcast is actually live on Pages. The first **beta** release is the real test, and it's the safe one to fail on.

## Tests & validation

Builds clean, Debug and Release, no new warnings. In the Release bundle `Sparkle.framework` is embedded and **re-signed with `Tinycast Self-Signed`** — the identity match the whole scheme rests on — and `codesign --verify --deep --strict` passes.

All ten harnesses from `docs/development.md § Tests` pass: fuzz, ranking, calc, clipboard, scopes, emoji, custom-command, snippets, system-command, window-command. No cases added — see the no-harness note above. SwiftLint isn't installed on this machine, so lint is unrun.

By hand, against the real feed URL, with prefs backed up and restored afterwards:

- **Off means off.** Launched with consent off: `SUHasLaunchedBefore=1` (the updater *did* start), `SULastCheckTime` absent, **zero open sockets**, no Sparkle cache directory.
- **Consent flips live.** Wrote the user default, relaunched: `SULastCheckTime` stamped within a second, and the log shows the request going to exactly `https://abue-ammar.github.io/tinycast/appcast.xml` — the delegate-supplied URL. It 404s (feed not deployed yet) and the background failure produced **no dialog**, which is the correct accessory-app behaviour.
- **User-initiated path renders our UI.** Drove the palette (⌘Space → "check for upd" → ↵) and got Tinycast's own dark error modal, "Update Failed", one OK button.
- **The publishing half works.** Ran the exact CI command against a real DMG: `generate_appcast` emitted a correctly signed item (`sparkle:version=200`, `shortVersionString=0.7.6`, `edSignature`, correct enclosure URL) and preserved the channel metadata.

## Before this can leave draft

Three of these are maintainer-only — I don't have write access to the tap or the secrets:

1. **Record the demo clips** (above).
2. **`gh secret set SPARKLE_ED_PRIVATE_KEY`** — the key currently exists only in my login keychain; `docs/signing.md §3` has the exact commands. Without it the appcast step warns and skips: releases still publish, nothing updates.
3. **Add `auto_updates true`** to `Casks/tinycast.rb` and `Casks/tinycast@beta.rb` in the tap. CI only rewrites `version`/`sha256`, so this is manual — and without it `brew upgrade` will clobber a self-updated app.
4. **Cut a beta first.** The install leg has never run against a live feed; beta is where to find out.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,10 @@ on:
 permissions:
   contents: write
 
+# Pinned in one place so the appcast signer can never drift from the Sparkle SPM pin in project.yml.
+env:
+  SPARKLE_VERSION: "2.9.4"
+
 jobs:
   build:
     runs-on: macos-26
@@ -44,10 +48,12 @@ jobs:
           CHANNEL="${{ github.event.inputs.channel }}"
           BASE_VERSION="${{ github.event.inputs.version }}"
           # Each channel is a distinct, side-by-side app: its own display name + bundle id,
-          # and its own Homebrew cask in the abue-ammar/homebrew-tinycast tap.
+          # its own Homebrew cask in the abue-ammar/homebrew-tinycast tap, and its own Sparkle feed
+          # (two separate appcasts, never sparkle:channel — a beta host must never be offered the
+          # stable DMG). This is the single place channel identity is decided.
           case "$CHANNEL" in
-            beta)   SUFFIX="-beta.${GITHUB_RUN_NUMBER}";  DISPLAY="Tinycast Beta";  BUNDLE="com.tinycast.app.beta";  CASK="tinycast@beta";  PRERELEASE=true ;;
-            stable) SUFFIX="";                            DISPLAY="Tinycast";       BUNDLE="com.tinycast.app";       CASK="tinycast";       PRERELEASE=false ;;
+            beta)   SUFFIX="-beta.${GITHUB_RUN_NUMBER}";  DISPLAY="Tinycast Beta";  BUNDLE="com.tinycast.app.beta";  CASK="tinycast@beta";  APPCAST="appcast-beta.xml";  PRERELEASE=true ;;
+            stable) SUFFIX="";                            DISPLAY="Tinycast";       BUNDLE="com.tinycast.app";       CASK="tinycast";       APPCAST="appcast.xml";       PRERELEASE=false ;;
           esac
           FULL="${BASE_VERSION}${SUFFIX}"
           {
@@ -55,6 +61,7 @@ jobs:
             echo "display_name=${DISPLAY}"
             echo "bundle_id=${BUNDLE}"
             echo "cask_name=${CASK}"
+            echo "appcast_file=${APPCAST}"
             echo "dmg_file=Tinycast-${FULL}.dmg"
             echo "prerelease=${PRERELEASE}"
           } >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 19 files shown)


woxiqingxian:
The request in #136 is still relevant: users who install Tinycast from the DMG have no in-app way to check for updates or update the app. They must manually visit GitHub Releases to find new versions.  
I noticed that #137 implemented in-app updates with Sparkle, but it was reverted in #139 after an accidental merge. Is there a plan to revisit and implement this feature in the future?  
In-app updates are important for DMG users. brew upgrade does not cover them, and long-term users can easily miss bug fixes and new features. At minimum, a “Check for Updates” action would help; ideally, the app could also download, verify, and install updates.

abue-ammar:
@woxiqingxian we will have our built in updater, no 3rd party

woxiqingxian:
> @woxiqingxian we will have our built in updater, no 3rd party

Thank you for open-sourcing this project. I look forward to this feature being released. Also, are there any plans to support multiple languages in the future?

abue-ammar:
> > @woxiqingxian we will have our built in updater, no 3rd party
> 
> Thank you for open-sourcing this project. I look forward to this feature being released. Also, are there any plans to support multiple languages in the future?

no strictly it wont be multi lang. you can fork it stay under same license and have your desired language

PR #145, opened 2026-08-02
Uninstall Application: remove an app and its leftovers

### LINKED ISSUE #50: [Feature Request] Quick uninstall

abue-ammar:
## Related issue

Closes #50

## What changed

Adds **Uninstall Application** — removes an app *and* the files it leaves behind (caches, preferences,
containers, saved state, launch agents, plug-ins, CLI launchers). Reached from the launcher's Actions
menu (⌘K → Uninstall Application) on any `.application` entry; it opens a new `.uninstall` palette
sub-screen scoped to that app, with a checkbox per file and a running size total.

**Everything goes to the Trash.** `FileManager.trashItem` is the only removal call in the feature —
`removeItem` never appears. That's load-bearing rather than incidental: it's what makes display-name
attribution tolerable at all, because a false positive costs a drag back rather than the user's data.

### Layout

The usual pure/effect split, same shape as `WindowManagement`:

| File | Role |
| --- | --- |
| `Core/Uninstall/UninstallTarget.swift` | `UninstallTarget`, `UninstallEvidence`, `UninstallIdentity` — every guard rail lives in `make` |
| `Core/Uninstall/UninstallSearchRoot.swift` | Root table + `binDirectories` |
| `Core/Uninstall/UninstallRules.swift` | Matching, `isAcceptableCandidate` |
| `Core/Uninstall/UninstallProtection.swift` | `PathFacts` → `UninstallProtection` |
| `Core/Uninstall/UninstallPlan.swift` | `UninstallCandidate`, `UninstallPlan`, `UninstallSelection` |
| `Core/Uninstall/UninstallScanner.swift` | **Impure.** `contentsOfDirectory`, `lstat`, sizes, FDA probe |
| `Core/Uninstall/UninstallRunner.swift` | **Impure.** `trashItem`, nothing else |
| `Core/Uninstall/UninstallSession.swift` | `@MainActor` lifecycle |
| `Features/Uninstall/UninstallView.swift` | List, row, actions menu |

The first five compile standalone into `Tools/uninstall-test.swift`. The scanner hands the rules child
**names** and hands the classifier a `PathFacts`, so "no filesystem in the pure layer" is structural
rather than a promise — the harness needs no temp directories at all.

### Non-obvious bits worth a reviewer's eye

- **The separator boundary in `matchesBundleID` is the whole safety story.** A plain `hasPrefix` makes
  `com.apple.SafariTechnologyPreview` a match for `com.apple.Safari` and trashes a different product's
  profile. `-` counts as well as `.`, because vendors name variants that way (`dev.zed.Zed-Preview`).
- **An installed sibling owns its own artifacts.** Without it, uninstalling `com.tinycast.app` would
  also trash `…​.beta` and `…​.dev` — the channel-isolation invariant in reverse.
- **Tinycast refuses to plan its own uninstall**, by bundle ID *and* bundle URL, compared against the
  **running** identity so the Dev channel refuses itself too.
- **`tccRelativePrefixes` is measured, not assumed.** A probe that creates and trashes a throwaway
  directory in each location shows `Containers` / `Group Containers` / `Cookies` refuse the move while
  `Application Scripts` allows it. Listing is *not* the test — both container roots enumerate fine and
  still refuse. That's why Books shows its five `Application Scripts` rows checkable and the five
  `Containers` rows beside them locked.
- **Ownership doesn't gate deletion; the parent directory does.** POSIX unlink needs write on the
  enclosing directory, not ownership of the item. Checking ownership first (as an earlier revision did)
  greyed out `/Applications/Xcode.app` — 4.19 GB that drags to the Trash fine, because `/Applications`
  is `drwxrwxr-x root:admin`. Ownership now decides exactly one case: a sticky parent.
- **The home directory is deliberately not a search root.** Claiming `~/<name>` rests on a name match,
  and VS Code's `CFBundleName` is literally `Code` — `~/Code` is a source tree on a lot of machines.
  Narrowing to dot-folders only moved the problem (an app named "Local" would claim `~/.local`) and
  needed a hand-kept blocklist with no source of truth. Measured across 62 installed apps the whole
  root was worth one 115 kB folder, so it's out. Rationale is in `docs/uninstall.md` so nobody re-adds
  it thinking it was an oversight.
- **CLI launchers are attributed by link target, never by name** — `zed` in `/usr/local/bin` is Zed's
  because it resolves into `Zed.app`, not because of what it's called.
- **`UninstallSelection` is the only thing that can hold a checked set**, and every mutation funnels
  through one intersection with `plan.removableIDs`, so "a locked candidate is never checked" is a
  single line to review rather than a rule spread across the view and the session.

### Also in this diff

- `.swiftlint.yml`: raised `type_body_length` (720→850), `function_body_length` (130→160) and added
  `large_tuple` (3). `AppCore` and `RootPaletteView` grew past the old ceilings.
- `docs/uninstall.md` (new), plus updates to `AGENTS.md` (new invariant), `docs/palette.md`,
  `docs/ui.md` (the shared leading-slot rule), `docs/development.md` and the CI `test` job.

## Memory footprint

> **Not yet measured — I don't have these numbers.** The scan is off-main via `Task.detached` and the
> only retained state is one `UninstallPlan` (a few hundred value-type candidates), released by
> `UninstallSession.cancel()` on every exit from the screen. Sizing uses a streaming `FileManager`
> enumerator capped at 250k entries rather than materialising a file list. Needs filling in before merge.

Leak-tested:

## Demo

> Not recorded — screen recording isn't available from my shell, so this needs a before/after capture
> before merge. Visual surface: the new sub-screen (list, checkboxes, lock rows), the red
> "Uninstall Application" footer pill, and the confirmation dialog.

## Drawbacks

- **`/usr/local/bin` rows render locked, where Raycast shows them checked.** On a stock Mac that
  directory is `drwxr-xr-x root:wheel`; I verified by attempting it that both the create and the trash
  are refused. Raycast's checkbox there cannot succeed without an admin password. Closing the gap needs
  privilege escalation, which this feature deliberately never does — and `trashItem` as root would land
  files in *root's* Trash, weakening the "always recoverable" guarantee that everything else rests on.
  `/opt/homebrew/bin` launchers are removable, because Homebrew leaves that directory group-writable.
- **Apps on the signed system volume (`/System/Applications/Books.app`) stay locked.** Raycast offers
  them as checked; that removal can't succeed. Their user-level leftovers are still cleanable.
- **`~` is not scanned**, so a leftover like `~/OrbStack` is missed. Deliberate — see above.
- **`UninstallProtection` is advisory, not a security boundary.** TCC is evaluated at the syscall, so it
  can be wrong in both directions; the runner still reports per-item failure.
- **Show Info in Finder drives Finder over Apple events**, so first use raises the system Automation
  prompt. Failure is reported with a pointer to Settings; the other actions need no permissions.
- Directory sizes are on-disk allocated bytes (Finder's number), so totals read slightly higher than
  Raycast's logical sizes.

## Tests & validation

**New:** `Tools/uninstall-test.swift` — 117 assertions over the pure layer, registered in
`docs/development.md` § Tests and the CI `test` job. No filesystem, no temp directories.

```sh
swiftc -swift-version 6 Tinycast/Core/Uninstall/UninstallTarget.swift \
    Tinycast/Core/Uninstall/UninstallSearchRoot.swift Tinycast/Core/Uninstall/UninstallRules.swift \
    Tinycast/Core/Uninstall/UninstallProtection.swift Tinycast/Core/Uninstall/UninstallPlan.swift \
    Tools/uninstall-test.swift -o /tmp/uninstall-test && /tmp/uninstall-test
```

Coverage is weighted toward near-misses rather than happy paths: `SafariTechnologyPreview` vs `Safari`,
`iBooksXtra` vs `iBooksX`, sibling release channels in both directions, vendor namespaces,
`group.`/Team-ID stripping and its near-misses, "Books" vs "Books Reader" both ways, name collisions
against other installed apps, self-uninstall refusal for both channels, the protection precedence
ladder, and the locked-can-never-be-checked invariant. It ends with a **cross-identity sweep** — 8
realistic apps × every root × every artifact shape, asserting no app's artifacts are ever attributed to
another. That's the one test that catches a matcher regression as a whole rather than a single rule.

I mutation-tested the harness rather than trusting a green run: removing the trailing-separator guard
fails 5 assertions, breaking the selection invariant fails 3.

**Also run:** every other `Tools/` harness green (fuzz, ranking, calc, clipboard, scopes, raycast,
emoji, custom-command, snippets, hotkey, callout, system-action, volume, window-command), `xcodebuild`
clean, `swiftlint` clean. `Tools/callout-test.swift` gains a guard that the uninstall checkbox fits the
shared `rowIcon` slot (it's the only harness compiling `Theme.swift`).

**By hand, off-repo:**

- **End-to-end** against a throwaway app in a fake `CFFIXED_USER_HOME`: real scanner + real runner, 25
  checks — files gone from source, all recoverable from `~/.Trash`, near-miss fixtures untouched.
- **Dry run over all 62 installed apps**, printing each candidate and its protection without trashing
  anything: **no path is claimed by two apps**. I read every `displayName` and `binSymlink` row by hand;
  all correctly attributed.
- **Permission probes** for the TCC table and the `/usr/local/bin` claim, both described above.

**Not done:** memory profiling, and no visual sign-off — both outstanding above.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,6 +71,12 @@ jobs:
             Tinycast/Core/WindowManagement/WindowLayout.swift \
             Tinycast/Core/WindowManagement/WindowActionMemory.swift Tools/window-command-test.swift \
             -o /tmp/window-command-test && /tmp/window-command-test
+          swiftc -swift-version 6 Tinycast/Core/Uninstall/UninstallTarget.swift \
+            Tinycast/Core/Uninstall/UninstallSearchRoot.swift \
+            Tinycast/Core/Uninstall/UninstallRules.swift \
+            Tinycast/Core/Uninstall/UninstallProtection.swift \
+            Tinycast/Core/Uninstall/UninstallPlan.swift Tools/uninstall-test.swift \
+            -o /tmp/uninstall-test && /tmp/uninstall-test
 
   lint:
     runs-on: macos-26
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -22,16 +22,20 @@ cyclomatic_complexity:
   error: 60
 
 function_body_length:
-  warning: 130
-  error: 200
+  warning: 160
+  error: 220
 
 function_parameter_count:
   warning: 9
   error: 12
 
+large_tuple:
+  warning: 3
+  error: 4
+
 type_body_length:
-  warning: 720
-  error: 900
+  warning: 850
+  error: 1000
 
 file_length:
   warning: 1200
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -110,7 +110,10 @@ Never break these without an explicit task to do so.
   directory **names**, never URLs, and hands the classifier a `PathFacts`. Every `FileManager`,
   `lstat` and Full Disk Access read lives in `UninstallScanner`, which **detects** FDA (a silent,
   promptless probe) and never requests it — this feature asks for no permission and never escalates
-  privilege. A locked candidate can never enter the checked set; that invariant lives in
+  privilege. `tccRelativePrefixes` is **measured, not assumed**: probe a location by creating and
+  trashing a throwaway directory there before adding it, because *listing* is not the test —
+  `~/Library/Containers` enumerates fine and still refuses the move, while
+  `~/Library/Application Scripts` allows it. A locked candidate can never enter the checked set; that invariant lives in
   `UninstallSelection`'s one intersection, not in the view. Tinycast also refuses to plan its own
   uninstall, compared against the **running** identity so the Dev channel refuses itself too.
   See [uninstall.md](docs/uninstall.md).
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -42,8 +42,8 @@ Full detail: [`docs/architecture.md`](docs/architecture.md).
 - **Subsystems:** [palette](docs/palette.md) · [launcher & fuzzy match](docs/launcher.md) ·
   [calculator](docs/calculator.md) · [clipboard](docs/clipboard.md) · [emoji](docs/emoji.md) ·
   [snippets](docs/snippets.md) · [window management](docs/window-management.md) ·
-  [hotkeys](docs/hotkeys.md) · [Raycast import](docs/raycast-import.md) ·
-  [UI & design system](docs/ui.md).
+  [hotkeys](docs/hotkeys.md) · [uninstall](docs/uninstall.md) ·
+  [Raycast import](docs/raycast-import.md) · [UI & design system](docs/ui.md).
 
 ## Critical Invariants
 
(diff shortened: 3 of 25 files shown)


abue-ammar:

PR #154, opened 2026-08-03
Add Quicklinks

### LINKED ISSUE #131: [Feature]: add quicklinks from Raycast
### What should it do?

The same thing it does in raycast

A quick way to open links or folders.

abue-ammar:
## Related issue

Closes #131

## What changed

Quicklinks: a saved URL, search, file, folder or deeplink becomes a first-class command — searchable
in the launcher, bindable to a global shortcut, openable in a chosen app, with dynamic placeholders
resolved at open time. Ships **off** behind a Settings switch. Full contract in
[`docs/quicklinks.md`](docs/quicklinks.md).

**Pure core** — `Core/Quicklinks/`, Foundation (+ SQLite3) only, so `Tools/quicklink-test.swift`
compiles the shipped sources rather than a copy:

- `Quicklink.swift` — model, `QuicklinkError`, and `Quicklink.precedes`, the one display order
  (pinned first by pin time, then by name) that both the store and the `AppIndex` slice sort through.
- `QuicklinkDestination.swift` — destination detection from shape alone, home directory injected.
- `QuicklinkStore.swift` — SQLite with `pinned_at` stamps.
- `QuicklinkArchive.swift` — versioned JSON import/export with name-or-destination dedup.

`QuicklinkLauncher` (every `NSWorkspace` call) and `QuicklinkArgumentSession` (prompt state) sit
beside them and stay out of the harness — the same split `Core/Uninstall/` uses.

**Surfaces** — a `.quicklink` `AppEntry.Kind` with its own launcher section between System Settings
and Snippets; four `CommandID`s (Create / Search / Import / Export); `PaletteMode.quicklinks` with a
full ⌘K menu; `PaletteMode.quicklinkArguments`; a Settings pane and editor sheet; per-item hotkeys;
settings-backup coverage.

### Non-obvious things in the diff

- **One template engine, not two.** Quicklinks expand through `SnippetTemplateEngine` rather than a
  second parser, so they inherit every token and modifier. Three additions, all behaviour-preserving
  because the new parameter defaults to off: a `expand(text:…)` entry point that the existing record
  overload now delegates to, a `ValueEncoding` applied *after* the modifier pipeline (so `| uppercase`
  can't rewrite `%xx` hex) and skipped when the template already spoke for itself, and
  `usesSelection`. This is what finally gives `| raw` a meaning — it opts a value out of the automatic
  percent-encoding a URL destination asks for. `{selectedText}` is accepted as an alias for
  `{selection}`; nothing ever *writes* it, so saved data keeps one spelling.
- **`QuicklinkStore` deliberately diverges from `ClipboardStore` in two places**, both because
  quicklinks are authored data rather than a regenerable cache: the database lives in Application
  Support, not Caches, and **a database that won't open is reported, never discarded**. The
  delete-and-recreate in `ClipboardStore.init` is only sound because history can be rebuilt; doing it
  here would destroy the user's library. There's a harness case that corrupts a file and asserts it
  survives byte-for-byte.
- **`AppEntry` gains one stored property**, `symbolName`. Quicklinks are the only kind whose glyph is
  the user's choice rather than its kind's; every other kind's `symbolIconName` is unchanged.
- **`AppIndex.setQuicklinks(_:commandsVisible:)` is one call for two slices** — the section and the
  four Quicklinks commands — so a toggle can't leave them out of step.
- **`HotKeyManager`'s UUID index is now shared.** Custom commands and quicklinks are both per-item
  rather than per-catalog-entry, so rather than a third copy of the same block there's now one
  `boundIDs` / `index` / `prune` / `persist` set that both namespaces use. `start()` takes a second
  ID set. This is also why `QuicklinkStore.load()` runs even while the feature is off: the
  stale-binding prune reads that list, and an unloaded store would look like "every quicklink was
  deleted" and throw the user's shortcuts away.
- **`AppPickerPopover` was promoted** out of `ClipboardSettingsView` (where it was `private`) into its
  own file, plus `AppPresentation.resolve` for the bundle-ID → name/icon fallback chain. It's the only
  app picker in the repo and the quicklink editor needs it; copying it would have been the wrong move.
  It gained an optional `clearTitle` row, so its callback is now `(String?) -> Void`.
- **`BackupActions` grew two shared panel helpers** (`chooseSaveLocation(named:)`, `chooseJSONFile()`)
  and `exportSettings` / `importSettings` now go through them instead of building panels inline.
- **`SettingsBackup` is `version` 4** — a `quicklinks` array, the flags, and `HotkeyBackup.quicklinks`.
  Unlike `snippetsEnabled`, `quicklinksEnabled` grants no permission class and enables no listening, so
  it's carried; excluding it would have been cargo-culting.
- **`LauncherView.rows` needed an explicit `[(String, [AppEntry])]` annotation.** With a ninth section
  the type-checker times out on the literal. Annotated with a comment saying why.
- **`Tinycast.xcodeproj` is regenerated.** The committed `.pbxproj` lists files individually, so new
  files do need `xcodegen generate` — worth knowing, since it's easy to assume the `path: Tinycast`
  glob resolves at build time.

### Second commit: the search-field placeholder step

Unrelated to Quicklinks in cause, found while testing it. The palette's search field no longer takes a
`prompt`; the placeholder is drawn as a leading-aligned background `Text`.

AppKit renders a `prompt` through either the field's cell or its field editor, and the editor is one
point taller than the field (measured: 24pt editor, 23pt field) — so identical placeholder glyphs sat
1pt higher once the field took the panel's shared field editor. That editor is created lazily then
cached on the window, which is why it only ever fired once per launch, and it only *read* as a bug
when the outgoing and incoming placeholders shared a leading word ("Search quicklinks…" →
"Search for apps and commands…").

Measured with a throwaway rasterizing probe, ink bounding box in backing pixels:

| configuration | ink y-range | shift on focus |
| --- | --- | --- |
| native prompt, unfocused | 29…65 | — |
| native prompt, focused | 27…63 | **−2px = −1pt** |
| background prompt, unfocused | 28…64 | — |
| background prompt, focused | 28…64 | **0** |

Header layout itself was ruled out first — the field lands at `minY 20.5, height 23.0` in all six
modes, so neither the placeholder string nor the mode icon was involved. A background rather than an
overlay so the caret still draws on top; `allowsHitTesting(false)` so clicking the placeholder still
lands the caret; an explicit `accessibilityLabel` because the prompt used to supply it.
`PaletteMode.placeholder` is still the single source of the strings.

## Memory footprint

Leak-tested:

## Drawbacks

Deliberate tradeoffs:

- **"Open in a new window" can't be fully honoured, and the UI says so.** `NSWorkspace` passes
  `arguments` only when it actually launches the app, so `--new-window` reaches Chromium and Firefox
  on a cold start and is dropped by Safari and by an already-running handler. Doing better needs
  per-browser AppleScript. The setting's subtitle states the limitation rather than implying it always
  works. It also absorbs "always create new tab" — off *is* prefer-existing-tabs, so two switches
  would have been one switch twice.
- **"Show website previews" was dropped.** A favicon fetch is a network read, and the invariant is that
  every networked feature ships off behind a consent dialog naming the provider, on an ephemeral
  cacheless session. That's its own feature, not a checkbox on this one. Rows use a local glyph.
- **"Pin to top" means the top of the Quicklinks section**, not above Applications. A second root-search
  position would need a second `AppEntry.Kind`, which the kind invariant forbids for one feature. Pins
  do get their own section inside Search Quicklinks.
- **The palette's ⌘K offers "Open With Default App", not an arbitrary app picker.** `PopoverMenu` is a
  flat list with no nesting or picker; choosing an arbitrary app belongs to the editor, which has one.
  What the palette can usefully offer is the one alternative that always exists.
- **`{snippet:Name}` isn't resolved inside a quicklink.** A quicklink is a destination, not a text
  template; the token stays literal and save-time validation rejects it. Avoids coupling quicklinks to
  whether snippets are enabled.
- **A templated link isn't validated at save.** You can't check a template's output — only the real
  value can be. Non-templated links are rejected at save; templated ones report at open time.

Worth a second opinion:

- **The store loads at launch even with the feature off** (one small SELECT), because the hotkey
  stale-binding prune reads that list. The alternative — skipping the prune for quicklinks — felt
  worse than one short read.
- **Typed text still steps that same 1pt on a focus change**; only the placeholder is pinned. The
  palette's invariant is that the field never resigns first responder, so there shouldn't be a moment
  where it's visible, but it's the same root cause left standing.
- **Two commits, not one.** The placeholder fix is genuinely separate from Quicklinks and reads better
  on its own; say the word and I'll squash.

## Tests & validation

**New:** `Tools/quicklink-test.swift` — 82 checks over the four pure sources. Destination-detection
table (https, bare host with path/query/port, multi-label host, `spotify://`, `shortcuts://`,
`mailto:`, `smb://`, `afp://`, literal space, already-encoded value, `~/`, bare `~`, `/tmp`,
`file://`, and the rejections: empty, whitespace, prose, leftover `{argument}`, `1.5`, `C:/Users`);
tilde expansion against an injected home; the encoding choice per link shape; `Quicklink.precedes`;
store CRUD, validation and pinning; **persistence across a close and reopen**; reading a database
written by the `sqlite3` CLI (proves the bind order matches the read order independently of our own
writer); **a corrupt database left byte-for-byte intact**; archive round-trip, merge counts, and a
hand-written JSON fixture. Registered in `docs/development.md` and `.github/workflows/ci.yml`.

**Extended:** `Tools/snippets-test.swift` — the `expand(text:)` entry point (including `{snippet:X}`
staying literal through it), `.percentEncoding` after the pipeline, `| raw` opting out,
`| percent-encode` not double-encoding, `.none` leaving snippet behaviour byte-identical, the
`{selectedText}` alias in every form (case, modifiers, parameter rejection, and through the record
overload so it isn't quicklink-only), and `usesSelection`.

**All 16 harnesses pass** (`docs/development.md` § Tests) and the app **builds clean** with no new
warnings — `xcodebuild -project Tinycast.xcodeproj -scheme Tinycast -configuration Debug build`.

**Docs updated:** new `docs/quicklinks.md`; `AGENTS.md` (invariant, layout, subsystem links);
`docs/snippets.md` (shared engine, the alias, `raw`'s new meaning); `docs/launcher.md` (publication
order, new section); `docs/palette.md` (the two modes, the argument flow, the placeholder rule);
`docs/hotkeys.md` (the per-item index); `docs/ui.md` (don't reintroduce `prompt:`);
`docs/development.md`.

**By hand:** _to fill in — the manual checklist is in `docs/quicklinks.md` and covers the argument
flow from a hotkey with the palette closed, the failure dialogs, import/export, and the feature
switch teardown._

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,3 +79,8 @@ jobs:
             Tinycast/Core/Uninstall/UninstallProtection.swift \
             Tinycast/Core/Uninstall/UninstallPlan.swift Tools/uninstall-test.swift \
             -o /tmp/uninstall-test && /tmp/uninstall-test
+          swiftc -swift-version 6 Tinycast/Core/Quicklinks/Quicklink.swift \
+            Tinycast/Core/Quicklinks/QuicklinkDestination.swift \
+            Tinycast/Core/Quicklinks/QuicklinkStore.swift \
+            Tinycast/Core/Quicklinks/QuicklinkArchive.swift Tools/quicklink-test.swift \
+            -o /tmp/quicklink-test && /tmp/quicklink-test
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -41,7 +41,8 @@ Full detail: [`docs/architecture.md`](docs/architecture.md).
   accessory apps).
 - **Subsystems:** [palette](docs/palette.md) · [launcher & fuzzy match](docs/launcher.md) ·
   [calculator](docs/calculator.md) · [clipboard](docs/clipboard.md) · [emoji](docs/emoji.md) ·
-  [snippets](docs/snippets.md) · [window management](docs/window-management.md) ·
+  [snippets](docs/snippets.md) · [quicklinks](docs/quicklinks.md) ·
+  [window management](docs/window-management.md) ·
   [hotkeys](docs/hotkeys.md) · [uninstall](docs/uninstall.md) ·
   [Raycast import](docs/raycast-import.md) · [UI & design system](docs/ui.md).
 
@@ -117,6 +118,19 @@ Never break these without an explicit task to do so.
   `UninstallSelection`'s one intersection, not in the view. Tinycast also refuses to plan its own
   uninstall, compared against the **running** identity so the Dev channel refuses itself too.
   See [uninstall.md](docs/uninstall.md).
+- **Quicklinks are authored data, and their store never deletes.** `Core/Quicklinks/` splits like
+  `Core/Uninstall/`: `Quicklink.swift`, `QuicklinkDestination.swift`, `QuicklinkStore.swift` and
+  `QuicklinkArchive.swift` stay Foundation-only (plus SQLite3) and pure for
+  `Tools/quicklink-test.swift` — the home directory is injected, never read — while `QuicklinkLauncher`
+  owns every `NSWorkspace` call and `QuicklinkArgumentSession` the prompt state. The database lives in
+  **Application Support**, not Caches, and a database that won't open is **reported, never discarded**:
+  `ClipboardStore`'s delete-and-recreate is only sound because history is regenerable, and a link
+  library is not. `Quicklink.precedes` is the one display order, sorted through by both the store and
+  the `AppIndex` slice. There is **one template engine**: quicklinks expand through
+  `SnippetTemplateEngine` rather than a second parser, which is what makes `| raw` mean something —
+  it opts a value out of the automatic percent-encoding a URL destination asks for. `{selectedText}`
+  is accepted as an alias for `{selection}`, but nothing ever *writes* it. See
+  [quicklinks.md](docs/quicklinks.md).
 - **`Core/SearchRelevance.swift` is Foundation-only and pure**, so `Tools/fuzz-test.swift` compiles
   the shipped scorer rather than a copy of it. It owns both `FuzzyMatch` (the tiered
   exact/prefix/word-start/substring/subsequence scorer) and the field bands. **Searchable fields stay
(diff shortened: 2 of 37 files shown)


abue-ammar:

Issue #212, opened 2026-08-08
[Bug]: UI Emoji Picker still weir


nhattran998:
### Existing issues

- [x] I searched open and closed issues and didn't find this already reported

### What happened?

<img width="1572" height="1004" alt="Image" src="https://github.com/user-attachments/assets/fbf8628f-f698-4dbd-8c45-e1190e9db3bb" />

One image - you can understand what happened.

### Steps to reproduce

Open TinyCast -> Ctrtl + Cmd + Space -> Display Emoji picker

### Tinycast version + channel

0.9.0

### macOS version

26.5.2 (25F84)

abue-ammar:
its a known bug. fix is on the way.

Issue #142, opened 2026-08-02
[Bug]: First search add scroll to result list


mathieusouflis:
### Existing issues

- [x] I searched open and closed issues and didn't find this already reported

### What happened?

When i search a single letter on tinycast, results are a bit scrolled.

(This happen only when i write the first letter)

EG : 
1. "a": bug
2. Add an a : No bug
2. Remove an a: No bug

### Steps to reproduce

1. Install tinycast Beta 0.8.0
2. Open tinycast
3. Search for "s"

### Screenshots or recording

https://github.com/user-attachments/assets/b908d65b-5363-4fb5-a532-ddc4e4915b19

### Tinycast version + channel

0.8.0 - Beta

### macOS version

26.5.2

abue-ammar:
i need someone with macos 26 fix this bug. i am on macos 27 beta.

isjiajia01:
I prepared and verified a fix for this issue, but the repository currently limits opening pull requests to collaborators, so GitHub does not expose the Create PR form for this fork and rejects both GraphQL and REST creation attempts.

Commit: https://github.com/isjiajia01/tinycast/commit/673ce95e65e99472fffaa49d55eb088ff946cdaa
Compare: https://github.com/abue-ammar/tinycast/compare/main...isjiajia01:fix/142-first-search-scroll

The fix re-emits the top-scroll intent on the next main-queue turn after the compact palette expands, when the newly mounted launcher list can consume it.

Validation completed:
- Debug build succeeded, targeting macOS 26
- All standalone harnesses in `docs/development.md` passed
- Independent final review: APPROVE, no blocker/major/minor findings

The available host runs macOS 27, so the macOS 26-specific visual result remains the one unverified item. A collaborator can open the compare link directly as a PR, or grant collaborator access so I can submit it with the prepared PR description.

abue-ammar:
related #212

isjiajia01:
Thanks for linking #212. I rebased the work onto the current `main` architecture and generalized the fix to cover both newly mounted palette screens and compact-to-expanded launcher transitions.

PR: #214
Commit: https://github.com/isjiajia01/tinycast/commit/865336a

The updated implementation re-emits the top-scroll intent after the mounted list has registered its listener, while cancelling that retry whenever a newer keyboard-follow or reset intent arrives. This avoids stale retries overriding selection navigation or rapid mode switches.

Validation completed:
- All 19 standalone harnesses passed
[...]

PR #214, opened 2026-08-09
Fix initial palette scroll after mounted screens

isjiajia01:
## Related issues

Addresses #212 and generalizes the mount-timing fix discussed in #142.

## What changed

The previous revision retried the top-scroll intent after `Task.yield()`. That fixed the missed mount event but could produce the visible second snap shown in the review recording, and a scheduler yield was not tied to safe-area inset settlement.

This revision makes each scroll view consume `.top` only after its header inset is usable. The shared handler owns all top scrolling; the seven screens retain only selection-follow behavior. Pending mount resets are cancelled when a newer `.follow`, wheel/trackpad interaction, or direct custom-scrollbar offset wins. Coalesced inset and offset changes distinguish the system origin adjustment from user movement, so no standing `.top` policy can repin a list later.

## Validation

- All 19 standalone harnesses pass
- Debug build succeeds targeting macOS 26 (`CODE_SIGNING_ALLOWED=NO`)
- SwiftLint passes with only pre-existing warnings
- Format check, purity check, and `git diff --check` pass
- macOS 27 UI smoke covered first-letter search, launcher to emoji with immediate keyboard navigation, and compact-to-expanded search
- Earlier 60 fps captures showed no second visible snap after removing the yield retry
- Independent final review: APPROVE, no blockers

## Limitation

The available host runs macOS 27, so the macOS 26-specific visual result still needs confirmation on an affected machine.

### DIFF
--- a/Tinycast/DesignSystem/Scrolling/ScrollIntent.swift
+++ b/Tinycast/DesignSystem/Scrolling/ScrollIntent.swift
@@ -21,6 +21,69 @@ extension View {
             Color.clear.frame(height: 0).id(ScrollOrigin.id)
         }
     }
+
+    /// Applies top intents once the header inset is usable; screens still own follow behavior.
+    func applyTopScrollIntent(_ scroll: ScrollIntent, proxy: ScrollViewProxy) -> some View {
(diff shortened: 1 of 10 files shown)


abue-ammar:
went with #217 as it is identical

PR #215, opened 2026-08-09
Fix/142 list scroll

abue-ammar:
#FollowUP #217


abue-ammar:
followup #217  since this branch is too old

PR #222, opened 2026-08-10
Search files through the Spotlight index

### LINKED ISSUE #81: [Feature]: File search
### What should it do?

Just use spotlight index should be fine.

I think it's one of the core functions of  a launcher.

mickyyy68:
## Related issue

Closes #81

A note on process: I had some unexpected free time late at night and was excited to explore this because I use Tinycast myself and genuinely enjoy the project. I went further than I should have before aligning on the issue, which is why I closed the first version of this PR. The maintainer’s concern was performance, so I used that time to validate the approach properly: the feature relies only on Spotlight’s existing index, caps every query, serializes execution, coalesces pending input, and includes measured latency and memory results. I’m reopening it now for discussion, not assuming the feature has already been accepted.

This uses the system Spotlight index proposed in the issue. Tinycast does not build, persist or watch a
second file index, so the launch and idle costs behind the concern in the issue discussion stay zero.

## What changed

- Adds **Search Files** as a built-in launcher command and a dedicated palette screen. It deliberately
  has no global shortcut, Settings control or persisted state.
- Searches filenames with a synchronous `MDQuery` on a detached user-initiated task after a 120 ms
  debounce. One worker serializes Spotlight calls and coalesces changes to the newest pending query, so
  slower typing cannot accumulate concurrent searches and superseded work cannot publish stale rows.
- Searches visible top-level home folders, direct visible home items, iCloud Drive and current
  `~/Library/CloudStorage` providers. `~/Library` is not admitted as a general scope.
- Excludes hidden paths, application bundles, `node_modules`, `DerivedData`, `build`, `dist`, `target`
  and `Pods`.
- Caps Spotlight at 1,000 candidates before execution, fuzzy-ranks those filenames with Tinycast’s
  existing scorer, and publishes at most 200 rows. `MDQuery` is used instead of `NSMetadataQuery`
  specifically because it exposes that source cap.
- Return opens the selected item, Command-Return reveals it in Finder, and the Actions menu can copy
  its full path.

The feature follows the existing AppCore/coordinator ownership and palette row-order contracts. Its
query, result and scope policies remain pure Foundation model files covered by a standalone harness.

## Performance and memory

Measured with the release-optimized `Tests/file-search-performance.swift` against the developer home
and its live Spotlight index:

| Measurement | Result |
| --- | --- |
| First service query | 598–701 ms |
| Repeated-query median | 612–703 ms |
| Standalone process maximum RSS | 27 MB |
| Palette debounce before service work | 120 ms |

An uncapped `mdfind` sanity check took 32.8 seconds for the broad one-letter query, which is why the
source cap is part of the feature invariant. `FileSearchService.search` also emits a permanent interval
on the existing `com.tinycast.perf` signpost subsystem for in-app Instruments runs.

Leak-tested: no private index, cache, watcher or retained query exists. The session owns one serialized
worker, keeps at most one pending query, and cancels publication when the query is cleared, superseded,
hidden or leaves the screen.

A follow-up stress run executed 60 broad one-letter searches sequentially in one process. Repeated
medians remained between 537–585 ms, peak RSS was 28.9 MB, and the process recorded no swaps or
meaningful latency growth. A typing simulation also confirmed that rapid input coalesces to the final
query and that slower input never produces concurrent Spotlight calls.

## Drawbacks

- Results inherit Spotlight freshness, file-permission omissions and provider availability.
- Search is filename-only. It does not inspect file contents or metadata descriptions.
- The 1,000-candidate bound means a matching item outside Spotlight’s bounded result set cannot be
  recovered by Tinycast’s ranking pass.
- `MDQuery`’s synchronous call cannot stop mid-flight. Its bounded result is discarded when superseded,
  then the same worker runs only the newest pending query.
- The measured service latency is roughly 0.6–0.7 seconds on this machine, plus the debounce. The UI
  keeps previous rows visible and exposes the searching state rather than blocking the main actor.

## Tests & validation

- `./Scripts/run-tests.sh` — all 21 harnesses pass, including deterministic coverage for serialized
  execution, debounce coalescing and cancellation.
- `./Scripts/lint.sh` — clean; existing warnings only.
- `Features/*/Model/` remains free of AppKit, SwiftUI and Cocoa imports.
- `git diff --check` passes.
- Unsigned Debug `xcodebuild` succeeds with no new warnings.
- Release-optimized live Spotlight benchmark recorded and documented.

Manual follow-up worth exercising on another home index: open Search Files from the launcher, compare
broad and specific queries, verify iCloud/provider results, then test Return, Command-Return and Copy
Path while watching the `FileSearchService.search` interval in Instruments.


Tinycast/Features/FileSearch/UI/FileSearchList.swift
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                }
+                .padding(.horizontal, Theme.Spacing.md)
+                .padding(.top, Theme.Spacing.xs)
+                .padding(.bottom, Theme.Spacing.md)
+                .hideNativeScrollers()
+                .scrollOriginAnchor()
+            }
+            .edgeDissolve()
+            .thinScrollbar()
```
abue-ammar, line 33: .pinOriginOnInsetSettle(scroll, proxy: proxy) missing. check other list view

abue-ammar:
@mickyyy68  idle time -> during file search -> after file search  memory usage is need. you can put those value keep activity monitor on and real time checking and put those value in summary. a detailed video for this is much needed

abue-ammar:
@mickyyy68 i was looking more at your pr. you have not added any settings at all. we MUST need a settings to completely toggle file search OFF completely. and by default it will stay completely OFF. and in settings we must need ignore pattern exactly like raycast. and how are you handling permissions? if i just type in search field will it annoy user with a prompt?. i havent tested it fully i will give more feedbacks when i will actually do tests

mickyyy68:
thanks, that makes sense. I’ll add a File Search setting, keep it off by default, and make sure turning it off removes the Search Files command completely.

On permissions: typing in Tinycast’s normal search does not start a file search or trigger a prompt, and enabling File Search is quiet too. I tested the actual first query with a clean Dev permission state, though, and I need to correct my earlier assumption: macOS does show Files & Folders prompts when that query reaches protected scopes such as Desktop, Documents, Downloads, and network volumes.

Tinycast still will not ask for Full Disk Access. Denying a location simply leaves it out of the results. I’m going to handle the first-use flow explicitly so those system prompts are explained before they appear instead of surprising the user.

One thing I want to confirm before building the ignore settings: when you say “exactly like Raycast,” do you mean the global patterns in Settings, like *.log and build/**, or should Tinycast also read .gitignore, .ignore, and .rayignore files inside folders? My instinct is to support the global patterns only, since Tinycast uses Spotlight and does not crawl the filesystem, but I’m happy to follow whichever behavior you have in mind.

I’ll also make the detailed video as soon as possible. I’m on vacation right now, but I’ll do it as soon as I can.

mickyyy68:
I tested this again with a clean Tinycast Dev permission state. Enabling File Search itself is quiet, but the first query makes macOS ask for access to protected locations such as Desktop, Documents, Downloads, and network volumes.

I see two ways we could handle this:

1. Keep whole-home search and explain the permission behavior before Search Files opens for the first time. macOS would ask when the user starts searching, and any location they deny would simply be left out of the results.
2. Avoid those automatic prompts by searching only folders the user explicitly selects.

I’m leaning toward the first option because it keeps File Search useful and is closer to Raycast’s home-folder permission model, without asking for Full Disk Access. Enabling the setting would remain quiet, the explanation would appear on the first open, and it would be remembered only on that Mac.

I’d rather check with you before building it. Would you prefer that approach, or explicit folder selection?

mickyyy68:
Quick heads-up: I accidentally committed two local design notes to the branch. I cleaned up the branch history instead of adding a separate deletion commit, so GitHub now shows a force-push and the commit hashes have changed. No implementation code changed during the cleanup. Sorry for the noise.

abue-ammar:
> I tested this again with a clean Tinycast Dev permission state. Enabling File Search itself is quiet, but the first query makes macOS ask for access to protected locations such as Desktop, Documents, Downloads, and network volumes.
> 
> I see two ways we could handle this:
> 
> 1. Keep whole-home search and explain the permission behavior before Search Files opens for the first time. macOS would ask when the user starts searching, and any location they deny would simply be left out of the results.
> 2. Avoid those automatic prompts by searching only folders the user explicitly selects.
> 
> I’m leaning toward the first option because it keeps File Search useful and is closer to Raycast’s home-folder permission model, without asking for Full Disk Access. Enabling the setting would remain quiet, the explanation would appear on the first open, and it would be remembered only on that Mac.
> 
> I’d rather check with you before building it. Would you prefer that approach, or explicit folder selection?

excellent work!. it is almost perfect! just need a few stuffs

we need these in settings
image
also this in the settings
image
command and we dont need alias btw

it got some merge conflict i am rebase it and force pushing it

abue-ammar:
i am working on ur branch a bit @mickyyy68

abue-ammar:
> > I tested this again with a clean Tinycast Dev permission state. Enabling File Search itself is quiet, but the first query makes macOS ask for access to protected locations such as Desktop, Documents, Downloads, and network volumes.
> > I see two ways we could handle this:
> > 
> > 1. Keep whole-home search and explain the permission behavior before Search Files opens for the first time. macOS would ask when the user starts searching, and any location they deny would simply be left out of the results.
> > 2. Avoid those automatic prompts by searching only folders the user explicitly selects.
> > 
> > I’m leaning toward the first option because it keeps File Search useful and is closer to Raycast’s home-folder permission model, without asking for Full Disk Access. Enabling the setting would remain quiet, the explanation would appear on the first open, and it would be remembered only on that Mac.
> > I’d rather check with you before building it. Would you prefer that approach, or explicit folder selection?
> 
> excellent work!. it is almost perfect! just need a few stuffs
> 
> we need these in settings image also this in the settings image command and we dont need alias btw
> 
> it got some merge conflict i am rebase it and force pushing it
i have implemented these and force push it @mickyyy68  great work

mickyyy68:
I appreciate it, thanks for your work too m8 🙏🏻

abue-ammar:
> I appreciate it, thanks for your work too m8 🙏🏻

welcome! also if you have discord join discord, i could give you a separate bug that you can take a look

mickyyy68:
already in it 🤙🏻 (magicmike68/MICHAEL, with the american psycho meme as pfp)

PR #228, opened 2026-08-11
Release File Search icons after dismissal

mickyyy68:
## Summary

Follow-up to #222 that fixes retained icon memory after leaving File Search.

Scrolling through a broad result set materializes fitted native file icons. Those icons were stored in
the shared persistent `IconCache`, so closing File Search left most of that memory resident.

This change:

- gives fitted file-row icons their own cost-bounded 8 MB cache
- purges that cache when File Search or Uninstall disappears, or when the palette closes
- invalidates in-flight loads so a late decode cannot repopulate a purged cache
- still returns an in-flight decode to its active row, preventing it from getting stuck on a placeholder
- leaves the existing persistent launcher-icon cache unchanged

Uninstall is included because it is the other consumer of the fitted-icon API. Purging when its list
disappears keeps the shared cache lifecycle consistent when the user backs out without closing the
palette.

## Measurements

Measured with a broad `.` search, scrolling through the complete result list, then closing File Search:

| | Idle | Peak while visible | After closing |
| --- | ---: | ---: | ---: |
| Physical footprint before | 21 MB | 78 MB | 77 MB |
| Physical footprint after | 20 MB | 78 MB | 52 MB |
| RSS before | 86 MB | 168.7 MB | 168.5 MB |
| RSS after | 86.6 MB | 169 MB | 142.8 MB |

The peak is expected while the rows and icons are visible. After closing, CG image residency falls to
2.2 MB and CoreAnimation residency to 2.7 MB instead of keeping the icon-heavy peak alive.

`leaks` reported only about 14 KB in Apple framework/XPC cycles, so this was retained cache lifetime
rather than unreachable objects.

## Testing

- Added a deterministic harness covering cache invalidation during an in-flight decode
- Added manual regression coverage for leaving and immediately restoring Uninstall without stranding an
  in-flight icon on its placeholder
- All 23 standalone harnesses pass
- Debug build succeeds with no new source warnings
- `./Scripts/lint.sh` is clean with existing warnings only
- Model-layer purity check passes
- `git diff --check` passes
- Manually verified with a broad search and full scroll

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -87,6 +87,7 @@ run hotkey-test            Tinycast/Features/HotKeys/Model/DoubleTapModifier.swi
                            Tinycast/Features/HotKeys/Service/KeyShortcut.swift
 run callout-test           Tinycast/DesignSystem/Theme.swift \
                            Tinycast/Features/HotKeys/UI/CalloutPlacement.swift
+run icon-cache-test        Tinycast/Platform/Images/IconCache.swift
 run system-action-test     Tinycast/Features/SystemActions/Model/SystemAction.swift
 run volume-test            Tinycast/Features/SystemActions/Model/VolumeLevel.swift
 run window-command-test    Tinycast/Features/WindowManagement/WindowCommand.swift \
--- a/Tests/file-search-session-test.swift
+++ b/Tests/file-search-session-test.swift
@@ -4,12 +4,26 @@ actor FileSearchProbe {
     private var active = 0
     private var calls: [String] = []
     private var maximumActive = 0
+    private let pausedQuery: String?
+    private var pauseEnabled: Bool
+    private var pausedContinuation: CheckedContinuation<Void, Never>?
+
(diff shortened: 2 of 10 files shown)


mickyyy68:
Quick note on the red CI check: I dug into it, and it isn’t coming from the icon-cache changes.

The session test waits 25 ms and assumes the first query has started before it submits the second. On the CI runner that task can start late, so the second query arrives while the first is still being debounced and replaces it. That is valid session behavior, but it breaks the test’s timing assumption.

The same assertion also failed in [the earlier #222 run](https://github.com/abue-ammar/tinycast/actions/runs/31432363615), before this branch. The full suite passes locally.

The proper fix is to let the probe explicitly tell the test when the first query has started instead of guessing from a sleep. I’d prefer to keep that as a small test-only follow-up so this PR stays focused.

abue-ammar:
@mickyyy68 ci is failing probably because that test can't be done in ci machine. if you could fix it

PR #245, opened 2026-08-14
Remove tautological and duplicate test assertions

mickyyy68:
# Remove tautological and duplicate test assertions

## What

Removes 33 assertions across 14 harnesses that cannot fail under any implementation. Each was audited against the production source it exercises and verified twice (original audit + independent review pass).

### Exact duplicates — same query, same expected value, asserted twice

- `calc-test`: `10km to mi`, `sin(30deg)`, `1 euro to dollars`, `100 dollars to yen`, `10 pounds to euros`, `1 cup to tbsp`
- `fuzz-test`: `rank("codex")` — and it never exercised the bundle-id path it claimed
- `hover-arming-test`: "keystroke disarms" block is behaviorally identical to "scroll disarms"
- `scroll-reveal-test`: `rowTop: 333` is `band - rowHeight` — same call as three lines up
- `window-command-test`: `frame(.centerHalf)` asserted in both `testTiling` and `testSizing`

### Tautologies — pass with literally any implementation

- `fuzz-test`: `score(...) == nil || score(...) != nil`; boost-cap checks that compare two constants (`4_500 < 10_000`); vacuous `||` whose second disjunct is structurally always true
- `snippets-test`: `decide(...) == active` where `active` was the identical call — `f(x) == f(x)` on a pure function
- `icon-cache-test`: `publish` returns its item unconditionally
- `settings-history-test`: init-parameter readback

### Impossible by construction

- `system-action-test` / `window-command-test`: magic catalog counts and ID-uniqueness checks — `all = ID.allCases.map`, and the production dictionaries are built with `Dictionary(uniqueKeysWithValues:)`, which traps before the test could
- `raycast-test`: asserted the harness's own fixture builder's output
- `palette-placement-test`: the same equation as the line above, rearranged

### Definitional readbacks

- `uninstall-test`: `defaultSelection.checked == removableIDs` — that *is* the definition of `defaultSelection`
- `palette-selection-test`: `[0, 0, 0]` empty-screen block duplicates the `[]` block
- `uninstall-test`: `.app` never stripped — entailed by the `matchableForms` assertion three lines up

## Verification

- `./Scripts/run-tests.sh` — all 31 harnesses pass
- `palette-selection-test` tripwire count in `docs/testing.md` updated (111,684 → 111,682)
- No behavioral coverage lost: every removed assertion was either entailed by a sibling or incapable of failing

## Not included

- The `uninstall-test` home/Library root contract checks — kept; they are a documented safety tripwire (`docs/features/uninstall.md`), not dead assertions
- `notes-test.swift` is untouched — Notes is not on `main` yet; those removals will land with the Notes feature

### DIFF
--- a/Tests/calc-test.swift
+++ b/Tests/calc-test.swift
@@ -81,8 +81,6 @@ struct CalcTests {
         expectDisplay("2sqrt(9)", "6")
         expectDisplay("2(3+1)+1", "9")  // implicit "*" binds like explicit "*", not looser
         expectDisplay("10π ^e", "224.5915772")  // and looser than "^"
-        // Units are never mistaken for a constant/function, so this stays untouched
-        expectDisplay("10km to mi", "6.213711922 mi")
         // Juxtaposition against a bracket carries the unit through, matching explicit "*"
         expectDisplay("2(3)kg", "6 kg")
         expectDisplay("2*(3)kg", "6 kg")
@@ -397,7 +395,6 @@ struct CalcTests {
         expectDisplay("1 rad to deg", "57.29577951 deg")
         expectDisplay("1 turn to deg", "360 deg")
         expectDisplay("200 grad to deg", "180 deg")
-        expectDisplay("sin(30deg)", "0.5")  // trig postfix still works inside parens
 
         // Implied quantity of 1 for number-less conversions
         expectDisplay("day to s", "86,400 s")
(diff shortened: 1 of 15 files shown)


mickyyy68:
right now I am focused on the notes feature, but after its completition I feel like this could be a good polish on the test side

PR #253, opened 2026-08-17
Add per-entry launcher aliases

### LINKED ISSUE #251: [Feature]: Per-entry aliases, set from ⌘K or Settings
### What should it do?

Let every launcher entry (application, command, quicklink, snippet, system action, …) carry one user-chosen **alias**, the way Raycast does:

- Searching the alias ranks that entry **first** — a deliberate user alias outranks every vendor-supplied field (display name, Spotlight alternate names, bundle id).
- Set it from a result's **⌘K Actions menu → Set Alias**: an in-palette panel in the Actions-menu corner with a text field (↵ saves, Esc closes, clearing the field removes the alias), or inline per row in the Settings panes that list launcher items.
- The row shows the alias as a small chip after the name.
- Aliases ride along in a settings backup, and uninstalling an app cleans up its alias like its other per-entry preferences.

Aliases are the launcher primitive for "I call this app something else": two keystrokes to the thing you open fifty times a day, without waiting for learned ranking to catch up and without renaming anything on disk. It's core Raycast behavior for a Raycast-style launcher, and it reuses machinery Tinycast already has (per-entry preference keys, the field-band scorer, the ⌘K menu) — one small store, one band, one panel; no new permissions and ~10 KB of state.

---

I have an implementation ready and will open a PR referencing this issue.

tipybara:
## Related issue

Closes #251

## What changed

Every launcher entry can carry one user-chosen **alias**, Raycast-style, ranked above everything else when searched.

- **Store** — `AliasStore` (`Launcher/Service/`), one alias per `preferenceKey` in UserDefaults, with a revision counter like `VisibilityStore`'s. Cleaned up on app uninstall alongside favorites/ranking; rides settings backups as `launcherAliases` (data, not a capability — nothing here grants a permission class).
- **Matching** — `SearchFields` gains `userAlias`; `SearchRelevance` gains band 6 above `nameLiteral`, matched **literally only** (like the identifier fields — a subsequence of a short alias is noise). `AppIndex` folds the alias in at rank time and keys both memos on the store's revision. The band-stride contract is untouched; the learned boost still reorders only within a band.
- **⌘K → Set Alias** — new row in `AppActionsMenu` for every entry kind. It opens `AliasEditorPanel`, an in-palette glass panel in the Actions menu's corner (same anchor, transition and `glassEffect` recipe as `PopoverMenu`): header, focused text field, footer with the entry's icon + name and Close (Esc) / Save (↵). ↵ or Save commits — an emptied field removes — Esc or click-away closes, and focus returns to the search field. While it is up the palette's ↑/↓/Tab/⌘K/⌃⇧Q handlers yield so the caret owns the keyboard; a click-catcher overlay shields the list and footer. Feedback goes through the existing message pill ("Alias Set" / "Alias Removed").
- **Rows** — a launcher row shows its alias as a small chip after the name (`AppRow` observes the store, so edits re-render live).
- **Settings** — every pane built on `LauncherItemsSection` (Applications, System Settings, System Actions, Commands) gets an inline per-row "Add alias" field; edits write through live and trim on focus loss.

Non-obvious bits: the alias band is deliberately literal-only; the editor panel deliberately does **not** set `PaletteState.menuOpen` (the menu-open input freeze would swallow typing); `AliasStore.setAlias` stores text as typed so a mid-word space isn't fought, with blank-when-trimmed meaning removal and the two commit points (panel save, settings focus loss) trimming.

## Memory footprint

Debug build of this branch, `footprint` phys_footprint:

| Step                                   | Memory |
| -------------------------------------- | ------ |
| Idle after launch                      | 52 MB  |
| Palette open                           | 56 MB  |
| Feature in use (alias panel open, peak)| 62 MB  |
| Palette closed, settled                | 61 MB  |

(Idle figure taken after a long-running session with warm caches; a cold launch idles at ~26 MB.)

Leak-tested: `leaks` reports ~19.7 KB total, all rooted in Apple frameworks (AppIntents/XPC connection blocks); nothing rooted in Tinycast code, and nothing referencing the alias types.

## Demo

Set alias "iterm" on kitty via ⌘K → Set Alias, then typing `iterm` ranks kitty first with an "iterm" chip on the row. Video to follow — flagging up front that it isn't attached yet.

## Drawbacks

- An alias matches literally only; `fga` will not fuzzy-match an alias `figalias`. Deliberate — predictability over reach for a field the user typed themselves.
- One alias per entry (Raycast allows one too). The store shape (`[String: String]`) would need a values-array migration to allow more.
- The Settings alias field stores keystrokes as typed until focus leaves; an edit abandoned mid-word persists un-trimmed until the next commit touches it. Harmless to matching (queries are trimmed) but worth knowing.

## Tests & validation

- `Tests/fuzz-test.swift` (compiles the real scorer): new **user aliases** section — alias tops ranking, band sits at 6×stride, alias beats another entry's exact display name, literal tiers match, subsequence never matches, strongest field still wins on an aliased entry. Band-ordering and stride assertions extended to the new band; the randomized property loop gains a user-alias field variant and corpus entry, and the band-violation bound moves 5 → 6.
- `./Scripts/run-tests.sh` — all 33 harnesses pass. `./Scripts/lint.sh` clean, Debug build with no new warnings, `Features/*/Model` purity grep clean, `xcodegen generate` run and committed.
- By hand (real CGEvent input, AX-verified): set/replace/remove via the ⌘K panel; ↵ saves, Esc and click-away close without saving, focus returns to search; alias chip appears and updates live; searching the alias ranks the entry first and the alias-only query matches nothing else; Settings inline field writes through and trims on blur; backup export includes `launcherAliases`.

### DIFF
--- a/Tinycast/Features/Launcher/Model/SearchRelevance.swift
+++ b/Tinycast/Features/Launcher/Model/SearchRelevance.swift
@@ -10,6 +10,8 @@ enum FuzzyMatch {
         case subsequence
 
         var isLiteral: Bool { self != .subsequence }
+        /// Anchored at the candidate's start: what a short, deliberate field must match to win its band.
+        var isAnchored: Bool { self == .exact || self == .prefix }
     }
 
     struct Match: Sendable {
@@ -120,6 +122,8 @@ enum FuzzyMatch {
 struct SearchFields: Sendable {
     /// The display name, plus anything identifying the entry just as strongly.
     var names: [String]
+    /// The user's own alias for the entry; deliberate, so it outranks every vendor field.
+    var userAlias: String?
     /// Spotlight's `kMDItemAlternateNames`: `iBooks`, `Codex`, `浏览器`.
     var alternateNames: [String] = []
     var bundleID: String?
@@ -135,6 +139,7 @@ enum SearchRelevance {
         case nameSubsequence = 3
         case alternateNameLiteral = 4
         case nameLiteral = 5
+        case userAlias = 6
 
         var offset: Int { rawValue * SearchRelevance.bandStride }
     }
@@ -156,6 +161,13 @@ enum SearchRelevance {
             best = max(best ?? Int.min, band.offset + match.score)
         }
 
+        // Only a hit from the alias's start earns the top band; one inside it ranks like a vendor alias.
+        if let alias = fields.userAlias, let match = FuzzyMatch.match(query, candidate: alias),
+            match.tier.isLiteral
+        {
+            let band: Band = match.tier.isAnchored ? .userAlias : .alternateNameLiteral
+            best = max(best ?? Int.min, band.offset + match.score)
+        }
         for name in fields.names {
             consider(name, literal: .nameLiteral, subsequence: .nameSubsequence)
         }
--- a/Tinycast/Features/Launcher/Settings/ApplicationsSettingsView.swift
+++ b/Tinycast/Features/Launcher/Settings/ApplicationsSettingsView.swift
@@ -12,5 +12,6 @@ struct ApplicationsSettingsView: View {
                 searchPrompt: "Search applications…")
         }
         .formStyle(.grouped)
+        .releasesFocusOnOutsideClick()
     }
 }
--- a/Tinycast/Features/Launcher/Settings/LauncherItemsSection.swift
+++ b/Tinycast/Features/Launcher/Settings/LauncherItemsSection.swift
@@ -11,9 +11,11 @@ struct LauncherItemsSection: View {
     @State private var query = ""
 
     private var entries: [AppEntry] {
-        // Run the matcher once per render, then scope the results to this category.
-        let matched = query.isEmpty ? appIndex.apps : appIndex.matches(query)
-        return matched.filter { $0.kind == kind }
+        let scoped = appIndex.apps.filter { $0.kind == kind }
+        guard !query.isEmpty else { return scoped }
+        // Membership only: score order would move the row being edited out from under the caret.
+        let matched = Set(appIndex.matches(query).map(\.id))
+        return scoped.filter { matched.contains($0.id) }
     }
 
     var body: some View {
@@ -68,6 +70,7 @@ private struct LauncherItemRow: View {
         SettingsRow(title: entry.name) {
             AppIconView(app: entry).frame(width: 18, height: 18)
         } trailing: {
+            AliasField(entry: entry)
             if let action = entry.hotKeyAction {
                 ShortcutRecorder(action: action)
             }
@@ -83,5 +86,67 @@ private struct LauncherItemRow: View {
             get: { visibility.isItemVisible(entry) },
             set: { visibility.setItemVisible($0, for: entry) }
         )
+    }
+}
+
+/// The per-row alias well, dressed like `ShortcutRecorder` so the trailing controls read as a set.
+/// A label until clicked: a form's field editor won't center, and a `Text` centers like Record's.
+private struct AliasField: View {
+    let entry: AppEntry
+    @Environment(AliasStore.self) private var aliases
+    @State private var isEditing = false
+    @State private var draft = ""
+    @FocusState private var focused: Bool
+
+    var body: some View {
+        let shape = RoundedRectangle(cornerRadius: Theme.Radius.menu, style: .continuous)
+        Group {
+            if isEditing {
+                TextField("", text: $draft)
+                    .textFieldStyle(.plain)
+                    .font(Theme.Typography.keyCap)
+                    .focused($focused)
+                    .onSubmit(finishEditing)
+                    .onExitCommand(perform: cancelEditing)
+                    // The pane's `releasesFocusOnOutsideClick` resigns; this catches it landing.
+                    .onChange(of: focused) { _, now in
+                        if !now { finishEditing() }
+                    }
+                    .onAppear { focused = true }
+            } else {
+                Text(aliases.alias(for: entry.preferenceKey) ?? "Add Alias")
+                    .font(Theme.Typography.keyCap)
+                    .foregroundStyle(Theme.Colors.textSecondary)
+                    .lineLimit(1)
+                    .frame(maxWidth: .infinity)
+                    .contentShape(shape)
+                    .onTapGesture(perform: beginEditing)
+            }
+        }
+        .padding(.horizontal, Theme.Spacing.sm)
+        .frame(width: Theme.Size.shortcutRecorder, height: 24)
+        .background(shape.fill(Theme.Colors.cardFill))
+        .overlay(
+            shape.strokeBorder(
+                isEditing ? Color.accentColor : Theme.Colors.cardStroke, lineWidth: 1))
+        .clipShape(shape)
+        .accessibilityLabel("Alias for \(entry.name)")
+    }
+
+    private func beginEditing() {
+        draft = aliases.alias(for: entry.preferenceKey) ?? ""
+        isEditing = true
+    }
+
+    /// The one commit path — ↵ or focus landing elsewhere; a blank draft removes the alias.
+    private func finishEditing() {
+        guard isEditing else { return }
+        isEditing = false
+        aliases.setAlias(
+            draft.trimmingCharacters(in: .whitespacesAndNewlines), for: entry.preferenceKey)
+    }
+
+    private func cancelEditing() {
+        isEditing = false
     }
 }
--- a/Tinycast/Features/Uninstall/UI/UninstallCoordinator.swift
+++ b/Tinycast/Features/Uninstall/UI/UninstallCoordinator.swift
@@ -12,6 +12,7 @@ final class UninstallCoordinator {
     private let favorites: FavoritesStore
     private let visibility: VisibilityStore
     private let ranking: LauncherRankingStore
+    private let aliases: AliasStore
     /// Dialog and message-HUD presentation only — never for state this type owns.
     private unowned let core: AppCore
 
@@ -25,6 +26,7 @@ final class UninstallCoordinator {
         favorites: FavoritesStore,
         visibility: VisibilityStore,
         ranking: LauncherRankingStore,
+        aliases: AliasStore,
         core: AppCore
     ) {
         self.session = session
@@ -36,6 +38,7 @@ final class UninstallCoordinator {
         self.favorites = favorites
         self.visibility = visibility
         self.ranking = ranking
+        self.aliases = aliases
         self.core = core
     }
 
@@ -112,6 +115,7 @@ final class UninstallCoordinator {
         favorites.remove(keys: [app.preferenceKey])
         visibility.removeItemKeys([app.preferenceKey])
         ranking.reset(itemKey: app.preferenceKey)
+        aliases.removeKeys([app.preferenceKey])
     }
 
     private func presentUninstallReport(_ report: UninstallReport) async {
--- a/Tests/fuzz-test.swift
+++ b/Tests/fuzz-test.swift
@@ -10,11 +10,13 @@ struct FuzzTest {
         var alternates: [String] = []
         var bundleID: String?
         var executable: String?
+        var userAlias: String?
 
         /// Mirrors AppEntry.searchFields, including the alternate-name sanitizing the scan applies.
         var fields: SearchFields {
             SearchFields(
                 names: [name],
+                userAlias: userAlias,
                 alternateNames: SearchFields.usableAlternateNames(
                     alternates, displayName: name, fileName: name + ".app"),
                 bundleID: bundleID, executableName: executable)
@@ -53,7 +55,11 @@ struct FuzzTest {
         // Ships an untranslated localization placeholder — see SearchFields.usableAlternateNames.
         App(name: "Maps", alternates: ["ALTERNATE_NAME_1", "Maps.app"]),
         // Alternate that only repeats the display name; contributes nothing.
-        App(name: "Image Playground", alternates: ["Image Playground", "Image Playground.app"])
+        App(name: "Image Playground", alternates: ["Image Playground", "Image Playground.app"]),
+        // The corpus entries with user aliases, so the property loop exercises the alias bands.
+        App(name: "Figma", userAlias: "fg"),
+        // The band-6 overreach repro: `term` inside `iterm` must not beat Terminal's own prefix.
+        App(name: "Kitty", userAlias: "iterm")
     ]
 
     static func app(_ name: String) -> App { apps.first { $0.name == name }! }
@@ -99,6 +105,7 @@ struct FuzzTest {
     static func main() {
         displayNameRanking()
         fieldPriority()
+        userAliases()
         alternateNameSanitizing()
         identifierFields()
         edgeCases()
@@ -197,15 +204,19 @@ struct FuzzTest {
             rank("system preferences").first == "System Settings")
 
         // Band ordering, asserted directly on the scores.
+        let userAlias = score("fg", "Figma")!
         let nameLiteral = score("chatgpt", "ChatGPT")!
         let aliasLiteral = score("codex", "ChatGPT")!
         let nameSubsequence = score("codex", "Code Explorer")!
         let aliasSubsequence = score("aplbks", "Books")!
         let identifier = score("openai", "ChatGPT")!
         let executable = score("electron", "Visual Studio Code")!
-        let ordered = [nameLiteral, aliasLiteral, nameSubsequence, aliasSubsequence, identifier, executable]
+        let ordered = [
+            userAlias, nameLiteral, aliasLiteral, nameSubsequence, aliasSubsequence, identifier,
+            executable
+        ]
         check(
-            "bands are strictly ordered: name > alias > name-fuzzy > alias-fuzzy > id > exec",
+            "bands are strictly ordered: user-alias > name > alias > name-fuzzy > alias-fuzzy > id > exec",
             zip(ordered, ordered.dropFirst()).allSatisfy { $0 > $1 }, "got \(ordered)")
         check(
             "every band is a whole stride apart",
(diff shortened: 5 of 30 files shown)


abue-ammar:
hey nice work! i never wanted it discussion is what i prefer.

1. remove claude's co authrorship

image

2.  this input field looks bad and the cursor is on the right it has to be like raycast, should look like shortcut recorder i guess. u know what i mean

image

3. i am keeping cmd+k popover menu clean. we dont need this setting alias from pop over. because users set this once and it can be done from the settings.

image


i will review code soon.

abue-ammar:
another gap i found. while import settings from raycast does it import alias and map it in tinycast?

Tinycast/Features/Uninstall/UI/UninstallCoordinator.swift
```diff
@@ -112,6 +115,7 @@ final class UninstallCoordinator {
         favorites.remove(keys: [app.preferenceKey])
         visibility.removeItemKeys([app.preferenceKey])
         ranking.reset(itemKey: app.preferenceKey)
+        aliases.removeKeys([app.preferenceKey])
```
abue-ammar, line 118: **Three of the four delete paths orphan the alias key.**

This is the only cleanup path that got `removeKeys`. Deleting a quicklink (`QuicklinkCoordinator.swift:261`), a custom command (`CustomCommandCoordinator.swift:115`) or an extension (`ExtensionCoordinator.swift:138`) still clears favorites, visibility and ranking but leaves the alias in `launcherAliases` forever — no UI can reach a key whose entry is gone, and it's exported in every settings backup from then on.

For extensions it's worse than a leak: `entryID` is `"extension:<name>/<command>"`, so reinstalling the same extension silently reattaches the stale alias.

Quicklinks and custom commands inject their stores, so add `aliases` next to `ranking` and drop it in alongside the others:

```swift
    favorites.remove(keys: entryIDs)
    visibility.removeItemKeys(entryIDs)
    aliases.removeKeys(entryIDs)
    for entryID in entryIDs {
        ranking.reset(itemKey: entryID)
    }
```

`ExtensionCoordinator` reaches its stores through `core`:

```swift
    core.favorites.remove(keys: Set(entryIDs))
    core.visibility.removeItemKeys(Set(entryIDs))
    core.aliases.removeKeys(Set(entryIDs))
```

Plus `aliases: aliases` in the two `AppCore` initializers that don't have it yet.

Tinycast/Features/Launcher/Settings/LauncherItemsSection.swift
```diff
@@ -63,11 +63,22 @@ struct LauncherItemsSection: View {
...
     let entry: AppEntry
     @Environment(VisibilityStore.self) private var visibility
+    @Environment(AliasStore.self) private var aliases
+    @FocusState private var editingAlias: Bool
 
     var body: some View {
         SettingsRow(title: entry.name) {
             AppIconView(app: entry).frame(width: 18, height: 18)
         } trailing: {
+            TextField("", text: aliasBinding, prompt: Text("Add alias"))
```
abue-ammar: **Live re-ranking tears down this field editor mid-word.**

`aliasBinding` writes per keystroke → `AliasStore.revision` bumps → `entries` re-reads `appIndex.matches(query)`, which is now alias-aware. With a non-empty filter the edited row can re-rank to the top mid-word, flipping the `if entry.id != entries.first?.id { Divider() }` branch for two rows. That structural change tears down the row's field editor and drops first responder — exactly the hazard called out at `RootPaletteView.swift:492`.

Ranking is incidental; plain relevance does it too. The fix is to stop score-ordering this list at all — use `matches` for membership and keep the index's name order:

```swift
private var entries: [AppEntry] {
    let scoped = appIndex.apps.filter { $0.kind == kind }
    guard !query.isEmpty else { return scoped }
    // Membership only: score order moves the row being edited out from under the caret.
    let matched = Set(appIndex.matches(query).map(\.id))
    return scoped.filter { matched.contains($0.id) }
}
```

`appIndex.apps` is already name-sorted, so this also settles something that predates this PR: today an empty filter gives alphabetical order and typing one flips the list into frecency order. Best-match-first belongs in the launcher, not in a list of checkboxes — a management list should keep rows where you left them.

abue-ammar (commented):
Reviewed `f9f27f0` — `./Scripts/run-tests.sh` passes (33/33).

The relevance work is sound: band 6 as literal-only is the right call for a 2–3 character alias (a subsequence of one would be pure noise), the boost cap of 4,500 sits well under `bandStride` so frecency can never lift a name match over an alias, and both `AppIndex` memo keys correctly carry `aliasRevision`. Backup round-trip and `ApplySummary` wiring check out, and `AliasStore` is injected into both hosting surfaces so there's no `@Environment` trap.

Three things to fix, inline.

Non-blocking: `docs/architecture.md:73-75` enumerates the stores `AppCore` owns and wasn't updated with `AliasStore` — though that list already omits `EmojiIndex`, so it was non-exhaustive before this PR.

Tinycast/Features/Launcher/UI/AppActionsMenu.swift
```diff
@@ -24,6 +24,9 @@ enum AppActionsMenu {
                     favorites.toggle(app)
                 })
         }
+        items.append(
+            PopoverMenuItem(
+                title: "Set Alias", systemImage: "character.cursor.ibeam", action: onSetAlias))
```
abue-ammar: **Let's drop Set Alias from the Actions menu and keep alias editing in Settings only.**

An alias is set once and then never touched again — it isn't a per-invocation action like Open, Show in Finder or Uninstall, which are the things ⌘K is for. Settings already has the inline field next to the visibility checkbox and the shortcut recorder, which is where the rest of an entry's one-time configuration lives. Two editing surfaces for one value is a maintenance cost with no payoff.

It also removes a good chunk of this diff, including a real bug (see my other comment on `RootPaletteView.swift` — now moot):

- `Tinycast/Features/Launcher/UI/AliasEditorPanel.swift` — the whole file
- `AppActionsMenu.swift` — this row and the `onSetAlias` parameter
- `LauncherScreen.swift` — the `openAliasEditor` property, its init parameter and the `onSetAlias:` argument
- `RootPaletteView.swift` — `aliasEditTarget`, both overlays, `openAliasEditor`/`closeAliasEditor`, the two `aliasEditTarget = nil` resets, and all seven keypress guards (↑ ↓ ↵ Esc Tab ⌘K ⌃⇧Q)
- `LauncherCoordinator.swift` — `alias(for:)` and `saveAlias(_:for:)` have exactly one caller each (the panel), so they go too, along with the `aliases` injection and the `aliases:` argument in `AppCore`

What stays: `AliasStore`, the `SearchRelevance` band, the `AppIndex` keys, the backup wiring, and the alias badge in `LauncherList` — the whole point of the feature is that the alias works in the launcher, not that it's editable there.

One knock-on: with Settings as the only editing surface, the field-editor teardown I flagged in `LauncherItemsSection.swift` goes from annoying to the only way to hit this feature, so it's worth fixing in the same pass.

tipybara:
All addressed — squashed and force-pushed as 8a2f8d2.

1. **Co-authorship removed** from the commit (on #252 as well).
2. **Alias field restyled** to match `ShortcutRecorder`: same well (`shortcutRecorder` width × 24, `cardFill` + `cardStroke`, `Radius.menu`), `keyCap` font, centered text and caret, accent border while editing, "Add Alias" placeholder. The trailing controls now read as a set.
3. **⌘K row dropped, Settings is the only editing surface.** Removed exactly per your list: `AliasEditorPanel.swift`, the menu row + `onSetAlias`, the `LauncherScreen` plumbing, all the `RootPaletteView` state/overlays/keypress guards, and `alias(for:)`/`saveAlias` + the `aliases` injection in `LauncherCoordinator`/`AppCore`. What stays is the store, the band, the memo keys, the backup wiring and the row chip.
4. **The three orphaning delete paths now clear aliases** — `QuicklinkCoordinator` and `CustomCommandCoordinator` get `aliases` injected next to `ranking`, `ExtensionCoordinator` goes through `core.aliases`, and both `AppCore` initializers carry the new argument. Good catch on the extension reinstall reattaching a stale alias.
5. **`LauncherItemsSection` filters by membership only**, using your exact shape — `appIndex.apps` scoped by kind, `matches(query)` reduced to an id set. The field editor stays put mid-word now, and the list keeps name order whether or not a filter is typed.
6. `docs/architecture.md`'s store list now includes `AliasStore` (non-blocking, but it was one word).

Validation on the new head: all 33 harnesses pass, lint clean, Debug build with no new warnings, and by hand — typing an alias into a filtered Settings list no longer drops focus, and deleting a quicklink/custom command clears its alias from `launcherAliases`.

tipybara:
> while import settings from raycast does it import alias and map it in tinycast?

Not yet, in either format.

- A **v1** export doesn't carry aliases at all — nothing in that payload shape to map.
- A **v2** export does: Raycast keeps the per-command alias in the same `settings.commands[]` array that `mapHotkeys`/`mapFavorites` already walk, so mapping application aliases into `launcherAliases` mirrors `mapFavorites` almost line for line (filter `extensionId == "e:r:applications"`, resolve the bundle id via `appPath(fromCommandID:)`, keep the alias string).

I left it out of this PR only because I don't have a fresh v2 export on hand to confirm the exact field name, and I'd rather not bake a guessed key into the mapper. If you can check a real export for what the alias field is called, I'll add the mapping — either here under a new `RaycastImportOptions` bit (an "Aliases" checkbox next to Favorites) or as a follow-up PR, whichever you prefer.

abue-ammar:
> > while import settings from raycast does it import alias and map it in tinycast?
> 
> Not yet, in either format.
> 
> * A **v1** export doesn't carry aliases at all — nothing in that payload shape to map.
> * A **v2** export does: Raycast keeps the per-command alias in the same `settings.commands[]` array that `mapHotkeys`/`mapFavorites` already walk, so mapping application aliases into `launcherAliases` mirrors `mapFavorites` almost line for line (filter `extensionId == "e:r:applications"`, resolve the bundle id via `appPath(fromCommandID:)`, keep the alias string).
> 
> I left it out of this PR only because I don't have a fresh v2 export on hand to confirm the exact field name, and I'd rather not bake a guessed key into the mapper. If you can check a real export for what the alias field is called, I'll add the mapping — either here under a new `RaycastImportOptions` bit (an "Aliases" checkbox next to Favorites) or as a follow-up PR, whichever you prefer.

i have raycast v2 would you like me to take care of it or you want to try?

tipybara:
Raycast alias import is in now (`d0fa18f`) — I got hold of a real Raycast X export (0.71.5.0) and confirmed the shape: each `settings.commands[]` entry carries a plain `alias` string, alongside `macosHotkey`/`favoriteOrder`.

- `RaycastImportV2.mapAliases` maps **application** aliases to `launcherAliases`, resolving the bundle id from the command id exactly like the hotkey and favorites mappers (`appPath(fromCommandID:)` → `Bundle`). Same scope as favorites: other command kinds don't have a Tinycast entry to attach to.
- New **Aliases** checkbox in the import picker (`RaycastImportOptions.aliases`), carried through `Result.selecting`. v1 exports carry no aliases, so the option is absent from `RaycastFormat.v1.supportedOptions` — asserted in `raycast-test` alongside the launch-at-login case.
- Verified end-to-end against the real export: two aliases (`Codex` on ChatGPT, `iterm` on kitty) come out as `["com.openai.codex": "Codex", "net.kovidgoyal.kitty": "iterm"]` and land in the launcher — typing the alias ranks the app first, chip on the row.
- `docs/features/raycast-import.md` updated; all 33 harnesses + lint pass.

Tinycast/Features/Launcher/Settings/ApplicationsSettingsView.swift
```diff
@@ -12,5 +12,6 @@ struct ApplicationsSettingsView: View {
                 searchPrompt: "Search applications…")
         }
         .formStyle(.grouped)
+        .releasesFocusOnOutsideClick()
```
abue-ammar, line 15: ### The palette's search field can go dead while this pane is open

`releasesFocusOnOutsideClick()` installs a **local** `NSEvent` monitor (`FocusRelease.swift:17`). A local monitor sees mouse-downs in **every window of the app**, not just this pane's. `release(on:)` does read `event.window`, but it then resigns first responder in whatever window received the click — so a click in the palette drops the *palette's* search focus.

Every palette mode shares one `TextField` (`RootPaletteView.swift:526`), so this hits the **app launcher, clipboard history and the emoji picker** alike.

It does not recover on its own. Focus only returns on a `focusToken` bump, which fires when the palette is shown or becomes key (`PaletteWindowController.swift:134`). The palette is already key, so nothing bumps it — you have to close and reopen the palette.

**How to reproduce**

1. Open Settings, go to the **Applications** pane, leave the window open.
2. Press the global hotkey to summon the palette on top of it.
3. Click anywhere in the palette that is not a text field (a result row, the blank area).
4. Type. Nothing lands in the search field.

The two windows only overlap when the palette is summoned deliberately, since it auto-hides on losing key status (`PaletteWindowController.swift:128`). `main` already has this via the Extensions pane, but this PR widens it to four more panes that are far more likely to be open.

**Fix: scope the monitor to the window the modifier lives in**

One change in `DesignSystem/Interaction/FocusRelease.swift` covers all five panes, so no per-pane edit is needed.

```swift
private struct FocusReleaseOnOutsideClick: ViewModifier {
    @State private var monitor: Any?
    // A box, not `@State` on the window: resolving it must not invalidate the view mid-layout.
    @State private var host = HostWindowBox()

    func body(content: Content) -> some View {
        content
            .background(HostWindowReader { host.window = $0 })
            .onAppear { /* unchanged */ }
            .onDisappear { /* unchanged */ }
    }

    private func release(on event: NSEvent) {
        // A local monitor sees every window in the app; only this pane's own is ours to touch.
        guard let window = event.window, window === host.window,
            // Only while something is actually being edited: the field editor is the responder.
            let editor = window.firstResponder as? NSTextView, editor.isFieldEditor
        else { return }
        let hit = window.contentView?.hitTest(event.locationInWindow)
        // A click on the field being edited, or on another text control, is that control's business.
        guard let hit, !hit.isDescendant(of: editor), !(hit is NSTextView) else { return }
        window.makeFirstResponder(nil)
    }
}

private final class HostWindowBox {
    weak var window: NSWindow?
}

/// The window a SwiftUI view landed in; `NSViewRepresentable` is the only route to it.
private struct HostWindowReader: NSViewRepresentable {
    var onResolve: (NSWindow?) -> Void

    func makeNSView(context: Context) -> NSView { HostWindowReaderView(onResolve: onResolve) }
    func updateNSView(_ nsView: NSView, context: Context) {}
}

private final class HostWindowReaderView: NSView {
    private let onResolve: (NSWindow?) -> Void

    init(onResolve: @escaping (NSWindow?) -> Void) {
        self.onResolve = onResolve
        super.init(frame: .zero)
    }

    required init?(coder: NSCoder) { fatalError() }

    override func viewDidMoveToWindow() {
        super.viewDidMoveToWindow()
        onResolve(window)
    }
}
```

Two details that matter:

- The window is held in a **box** rather than plain `@State`. Writing `@State` from `viewDidMoveToWindow` runs during layout and trips SwiftUI's "modifying state during view update".
- `host.window` is `nil` until the view lands in a window, and the `===` guard then fails closed — no release, which is the safe default.

**How to verify the fix**

Repeat the repro. The palette search field keeps focus and typing works. Then click blank space inside the Settings pane while editing an alias field — focus should still drop there, which is the behaviour the modifier exists for.

Tinycast/Features/Launcher/Model/SearchRelevance.swift
```diff
@@ -156,6 +159,10 @@ enum SearchRelevance {
             best = max(best ?? Int.min, band.offset + match.score)
         }
 
+        // Literal only, like the identifiers: an alias is short, and a subsequence of one is noise.
+        if let alias = fields.userAlias {
+            consider(alias, literal: .userAlias, subsequence: nil)
```
abue-ammar: ### A partial alias hit claims the top band, so it outranks an exact app name

`consider` accepts any tier where `isLiteral` is true, and `isLiteral` is `!= .subsequence` (`SearchRelevance.swift:12`). So `wordStart` and `substring` also land in band 6 — above `nameLiteral` (band 5), where an exact name match sits.

**How to reproduce**

1. Settings → Applications, set the alias `iterm` on **Kitty**.
2. Open the launcher and search `term`.
3. Kitty ranks above **Terminal**.

`term` is only a *substring* of `iterm`, yet it takes band 6; Terminal's own name matches as a prefix and sits in band 5. Any alias containing a common fragment does this — the shorter the query, the more entries it displaces.

**This is partly deliberate, so the fix is about the band, not the matching**

The test at `Tests/fuzz-test.swift:248` ("a user alias matches literally: exact, prefix and substring") says substring matching is intended, and that seems right — an alias `mail2` should still be findable by `ail`. The problem is only that a weak hit gets the *strongest* band. Keep it matchable, rank it honestly:

```swift
// in FuzzyMatch.Tier
/// Anchored at the candidate's start: what a short, deliberate field must match to win its band.
var isAnchored: Bool { self == .exact || self == .prefix }
```

```swift
// Only a hit from the alias's start earns the top band; one inside it ranks like a vendor alias.
if let alias = fields.userAlias, let match = FuzzyMatch.match(query, candidate: alias),
    match.tier.isLiteral {
    let band: Band = match.tier.isAnchored ? .userAlias : .alternateNameLiteral
    best = max(best ?? Int.min, band.offset + match.score)
}
```

This mirrors the explicit tier check already used for `bundleID` a few lines below, so it stays in keeping with the rest of the function.

If reusing `.alternateNameLiteral` for an alias reads wrong semantically, the alternative is a dedicated `case userAliasUnanchored` inserted below `nameLiteral` and the later cases renumbered. Same ranking, larger diff — your call.

**Test changes that go with it**

- `Tests/fuzz-test.swift:248` still passes: substring hits keep scoring non-`nil`.
- `Tests/fuzz-test.swift:258` ("the strongest field still wins on an aliased entry") asserts `>= 5 * bandStride`, which band 6 also satisfies — it cannot fail if the alias band wrongly claims the score. Add the upper bound so this stays fixed:

```swift
let score = SearchRelevance.score(query: "figma", fields: app("Figma").fields)!
check(
    "the strongest field still wins on an aliased entry",
    score >= 5 * SearchRelevance.bandStride && score < 6 * SearchRelevance.bandStride)
```

- Worth adding one case for the repro itself: an entry named `Kitty` aliased `iterm` must score below an entry named `Terminal` for the query `term`.

abue-ammar:
@tipybara the PR is solid very close to merging it. i found 2 issues. fix these if these are not false positive.

tipybara:
Both were real — fixed in `e79223b`.

**1. FocusRelease monitor scoped to its own window.** Took your fix as written: `HostWindowReader` (`NSViewRepresentable` → `viewDidMoveToWindow`) resolves the pane's window into a weak `HostWindowBox`, and `release(on:)` now guards `window === host.window` before touching first responder — nil fails closed. One change in `FocusRelease.swift`, covers all five panes.

Verified live with your repro: Settings › Applications open, palette summoned on top, typed `test1` into the search field, clicked blank palette area, typed `abc` — the field reads `test1abc`, focus never dropped. The in-window release path is unchanged by construction (`===` is always true for the pane's own clicks).

**2. Only anchored alias hits earn band 6.** Added `Tier.isAnchored` (exact/prefix) and replaced the alias `consider` with the explicit-match shape you suggested — an inside hit now lands in `.alternateNameLiteral` (kept your smaller-diff option), a subsequence still matches nothing.

Test changes: the corpus gains `Kitty` aliased `iterm`; new cases assert `term` ranks Terminal above Kitty while Kitty stays findable, `ite` still puts Kitty first, and an inside hit's score sits inside band 4. "Strongest field still wins" got your `< 6 × bandStride` upper bound so it can actually fail now, and the randomized inversion check is tier-aware (anchored must beat the name, inside must rank below it). `docs/features/launcher.md`'s band table and alias section updated to say anchored.

All 33 harnesses pass, lint clean, no new warnings.

abue-ammar:
hey nice work merging it. thanks a lot!

PR #254, opened 2026-08-18
fix(palette): hide search field placeholder during IME composition / 修复输入法候选输入阶段占位符文本未清除的问题

mrasong:
fix(palette): hide search field placeholder during IME composition  
fix(palette): 修复输入法候选输入阶段占位符文本未清除的问题  
──────

#### Summary / 概述

Switched the palette search TextField to use SwiftUI's native prompt parameter instead of a custom .background placeholder view.

将搜索栏 TextField 的占位符（Placeholder）改为使用 SwiftUI 原生 prompt 参数实现，移除之前基于 .background 自定义叠加的占位符逻辑。

#### Problem / 问题

When typing with an IME (e.g., Chinese, Japanese) where marked text / pinyin composition is active before confirming characters into the text field, vm.query remains empty. The custom .background overlay checked vm.query.isEmpty, causing
the placeholder text to overlap with the ongoing IME composition text.

在使用中文/日文等输入法（IME）输入拼音或处于候选组合状态（Composition / Marked Text）但尚未将文字正式确认上屏时，绑定的 vm.query 仍为空字符串 ""。原有通过 .background 并在 vm.query.isEmpty 为 true
时展示的自定义占位符不会被隐藏，导致占位符文本与正在输入的候选拼音/字母发生视觉重叠。

screenshot-260818-130754

#### Solution / 修改方案

• Used SwiftUI's native prompt: Text(searchPrompt).foregroundStyle(Theme.Colors.textTertiary) directly on TextField.
• AppKit automatically manages placeholder visibility during active text input and IME composition states (hasMarkedText), eliminating the overlap.
• Removed custom overlay background logic on the search field.

• 直接在 TextField 上使用原生 prompt: Text(searchPrompt).foregroundStyle(Theme.Colors.textTertiary) 参数。
• 借助 AppKit / NSTextView 原生机制，在输入法处于 marked text（组合输入）阶段时由系统自动处理并隐藏占位符。
• 移除手动在 .background 绘制的占位符相关逻辑，精简代码。

screenshot-260818-130814

#### Testing / 测试验证

[✓] Verified placeholder displays when the search field is empty.
[✓] Verified placeholder disappears immediately upon IME composition (e.g. typing Pinyin) before characters are committed.
[✓] Verified placeholder restores properly when query and composition are cleared.

[✓] 搜索栏内容为空时正常展示占位符。
[✓] 使用输入法输入拼音/候选词时，占位符立即隐去，不再与候选字符重叠。
[✓] 清空输入或取消输入法组合时，占位符恢复正常显示。

### DIFF
--- a/Tinycast/Palette/PalettePanel.swift
+++ b/Tinycast/Palette/PalettePanel.swift
@@ -8,11 +8,29 @@ final class PalettePanel: NSPanel {
     var onBareBackspace: (() -> Bool)?
     /// Command chords the field editor swallows, plus the ones no main menu handles.
     var onCommandShortcut: ((NSEvent) -> Bool)?
+    /// Tracks IME marked-text composition so the hand-drawn placeholder can hide during input.
+    private var selectionObserver: NotificationToken?
     /// Arms hover from `sendEvent`, the one place both event streams pass through.
     weak var paletteState: PaletteState? {
         didSet {
             paletteState?.onMenuOpenChanged = { [weak self] open in self?.setSearchCaretHidden(open) }
+            attachSelectionObserver()
         }
+    }
+
+    private func attachSelectionObserver() {
+        guard let editor = firstResponder as? NSTextView else { return }
+        selectionObserver = NotificationToken(
+            NotificationCenter.default.addObserver(
+                forName: NSTextView.didChangeSelectionNotification,
+                object: editor,
+                queue: .main
+            ) { [weak self] _ in
+                guard let self, let editor = self.firstResponder as? NSTextView else { return }
+                self.paletteState?.isComposing = editor.hasMarkedText()
+            },
+            center: .default
+        )
     }
 
     /// Keys driving an open menu; they reach `onKeyPress` even while editing is frozen.
@@ -151,6 +169,14 @@ final class PalettePanel: NSPanel {
         // The controller owns the frame; without this the top edge drifts on the swap.
         hosting.sizingOptions = []
         contentView = hosting
+    }
+
+    override func makeFirstResponder(_ responder: NSResponder?) -> Bool {
+        let result = super.makeFirstResponder(responder)
+        if result, responder is NSTextView {
+            attachSelectionObserver()
+        }
+        return result
     }
 
     override var canBecomeKey: Bool { true }
--- a/Tinycast/Palette/PalettePanel.swift
+++ b/Tinycast/Palette/PalettePanel.swift
@@ -20,6 +20,7 @@ final class PalettePanel: NSPanel {
 
     private func attachSelectionObserver() {
         guard let editor = firstResponder as? NSTextView else { return }
+        paletteState?.isComposing = editor.hasMarkedText()
         selectionObserver = NotificationToken(
             NotificationCenter.default.addObserver(
                 forName: NSTextView.didChangeSelectionNotification,
@@ -169,6 +170,11 @@ final class PalettePanel: NSPanel {
         // The controller owns the frame; without this the top edge drifts on the swap.
         hosting.sizingOptions = []
         contentView = hosting
+    }
+
+    func syncMarkedTextState() {
+        guard let editor = firstResponder as? NSTextView else { return }
+        paletteState?.isComposing = editor.hasMarkedText()
     }
 
     override func makeFirstResponder(_ responder: NSResponder?) -> Bool {
--- a/Tinycast/Palette/PaletteState.swift
+++ b/Tinycast/Palette/PaletteState.swift
@@ -7,6 +7,8 @@ final class PaletteState {
     var mode: PaletteMode = .launcher
     var query: String = ""
     var selection: Int = 0
+    /// True while an IME (e.g. Pinyin) has marked text in flight.
+    var isComposing: Bool = false
     /// The clipboard screen's type filter, reset with the rest of the screen state on each summon.
     var clipboardFilter: ClipboardFilter = .all
     /// Changes every time the palette is shown so the search field can re-focus.
@@ -43,6 +45,7 @@ final class PaletteState {
         self.mode = mode
         query = ""
         selection = 0
+        isComposing = false
         commandArguments = [:]
         clipboardFilter = .all
         forceExpanded = false
--- a/Tinycast/Palette/PaletteWindowController.swift
+++ b/Tinycast/Palette/PaletteWindowController.swift
@@ -58,10 +58,12 @@ final class PaletteWindowController: NSObject, NSWindowDelegate {
             // Non-activating, so summoning never raises our own aux windows behind it.
             panel.makeKeyAndOrderFront(nil)
             panel.orderFrontRegardless()
+            panel.syncMarkedTextState()
             // A never-activated login item can drop the first key request, so re-assert.
             DispatchQueue.main.async { [weak panel] in
                 guard let panel, panel.isVisible, !panel.isKeyWindow else { return }
                 panel.makeKeyAndOrderFront(nil)
+                panel.syncMarkedTextState()
             }
         }
     }
@@ -134,6 +136,7 @@ final class PaletteWindowController: NSObject, NSWindowDelegate {
     func windowDidBecomeKey(_ notification: Notification) {
         DispatchQueue.main.async { [weak self] in
             self?.core.palette.focusToken = UUID()
+            self?.panel?.syncMarkedTextState()
         }
     }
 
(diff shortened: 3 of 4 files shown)


abue-ammar:
Thanks for this — the bug is real and the screenshots make it obvious. During IME composition `vm.query` is still `""`, so the hand-drawn placeholder stays up and collides with the pinyin. That's a genuine defect and I want it fixed.

I can't take this patch as-is, though. It re-opens a bug we deliberately fixed and documented. Details below, plus two ways forward.

感谢反馈，这个 bug 是真实存在的。但这个改法会让我们之前修好并写进文档的另一个 bug 回归。下面是详细说明和两个替代方案。

---

## Why the placeholder is drawn by Tinycast, not by `prompt:`

Documented in `docs/features/palette.md`, section *"The placeholder is Tinycast's, not the field's"*, and in `docs/ui.md`, section *"The palette search field"*, which says in as many words: **"Don't reintroduce `prompt:` on that field."**

The reason:

- AppKit gives an `NSTextField` a field editor **one point taller** than the field itself — measured, a 24pt editor in a 23pt field. `PalettePanel.fieldEditorSlack` exists for the same reason.
- A `prompt` is rendered by whichever of the **cell** and the **field editor** currently owns the text.
- So the same placeholder glyphs sit one point higher once the field takes the panel's shared field editor. The placeholder **steps vertically** when focus moves.
- That editor is created lazily and then cached on the window for its lifetime, which is why the step was only ever visible on the **first summon after launch** — and only where the eye could track it, i.e. when the outgoing and incoming placeholders share a leading word (`Search…` → `Search apps…`).

Drawing it in SwiftUI pins it to the layout instead: measured ink is identical in both focus states, against a two-backing-pixel step for the real prompt. This is the same class of bug as the menu-open input freeze documented just below it — both come from the cell / field-editor swap.

So this PR trades a visible IME bug for a subtle focus-transition bug. I'd rather not trade at all.

## Other issues in the current diff

**1. Placeholder colour may silently change.** macOS commonly ignores `foregroundStyle` on a `TextField` prompt and renders the placeholder in the system placeholder colour instead. The palette's placeholder must be `Theme.Colors.textTertiary` (white 0.40), the same token every other tertiary label uses. This needs a side-by-side check on a real build before we can call it equivalent.

**2. Stale comment.** The line

```swift
// The prompt used to carry this; without it the field would be unlabelled.
.accessibilityLabel(Text(searchPrompt))
```

is now false — the prompt *does* carry it again. With `prompt:` back, either the explicit `accessibilityLabel` goes or the comment does. Two sources for one label is exactly the drift we try to avoid.

**3. Missing newline at end of file.** `RootPaletteView.swift` now ends with `\ No newline at end of file`. SwiftLint's `trailing_newline` flags it. Warning-only so CI won't block, but it's avoidable noise.

**4. `lineLimit(1)` dropped.** Harmless in practice (a single-line `NSTextField` truncates anyway), noting it for completeness.

**5. Docs not updated.** `AGENTS.md` asks that any doc a change makes wrong is fixed in the same commit. As it stands three places would contradict the code: the `docs/ui.md` section above, the whole `docs/features/palette.md` section, and that section's own heading.

---

## Preferred fix: keep the SwiftUI placeholder, hide it while composing

We get both fixes instead of trading one for the other, and the plumbing already exists.

`PalettePanel` already reaches into SwiftUI's live field editor to hide the caret while a menu is open (`setSearchCaretHidden`, `docs/features/palette.md#menu-open-input-freeze`). The same editor answers `hasMarkedText()`. So:

1. Add an observed `var isComposing = false` to `PaletteState` (observed, not `@ObservationIgnored` — the view needs to react to it).
2. In `PalettePanel`, once the search field is first responder, observe `NSTextView.didChangeSelectionNotification` on that editor and mirror `editor.hasMarkedText()` into `state.isComposing`. Selection changes fire on every marked-text update **and** on commit, so it covers the candidate window too. Use `NotificationToken` from `Platform/` so it unregisters cleanly.
3. Gate the existing background on it:

```swift
.background(alignment: .leading) {
    if vm.query.isEmpty && !vm.isComposing {
        Text(searchPrompt)
            …
    }
}
```

Two things to watch when implementing:

- **Don't poll `hasMarkedText()` from `sendEvent`.** Committing a candidate by clicking the IME candidate window never reaches `PalettePanel.sendEvent` — that's a separate window — so the flag would stick. The notification is the reliable signal.
- **Don't resign or re-take first responder** anywhere in this path. That triggers the cell / field-editor swap and shifts the text by a point, which is the bug we're avoiding in the first place.

I haven't run this on-device, so please verify the notification actually fires for marked-text updates in your IME before we call it done.

**测试项:** 拼音输入时占位符立即隐藏；用鼠标点击候选词上屏后占位符保持隐藏；清空输入后占位符恢复；启动后第一次唤起面板、切换模式时占位符**不出现上下 1 点的跳动**。

## If we decide the trade is worth it after all

Then the PR needs to say so explicitly and carry the rest with it:

- Update both doc sections, including the `palette.md` heading, and record that the 1-point step on first summon is a known, accepted cost.
- Resolve the duplicate accessibility label (issue 2).
- Confirm the placeholder colour still matches `textTertiary` on a real build (issue 1).
- Restore the trailing newline (issue 3).

Either route is fine by me, but the preferred fix keeps both behaviours correct and I think it's the smaller change in the end. Happy to review a follow-up push, and thanks again for finding this — it's the kind of bug that only shows up if you actually type in an IME every day.

mrasong:
fix(palette): hide search field placeholder during IME composition

### Summary

  During IME composition (e.g. Chinese Pinyin or Japanese Kana), `vm.query` remains empty until the text is committed, causing the hand-drawn placeholder to remain visible and collide with the marked text. Additionally, when
re-summoning or focusing the palette while an IME composition is in flight, the placeholder could reappear under the marked text.

  This PR hides the placeholder while marked text is in flight without reintroducing `TextField(prompt:)`, preserving layout stability and avoiding the 1pt focus-transition jump documented in `docs/features/palette.md`.

### Changes

  - **`PaletteState`**: Added observed `isComposing: Bool` state, reset in `prepare(mode:)`.
  - **`PalettePanel`**:
    - Subscribes to `NSTextView.didChangeSelectionNotification` on the active field editor via `NotificationToken` to mirror `editor.hasMarkedText()` into `paletteState.isComposing`.
    - Exposes `syncMarkedTextState()` and syncs composition state immediately upon attaching the observer or when becoming first responder.
  - **`PaletteWindowController`**: Calls `panel.syncMarkedTextState()` on `show()` and `windowDidBecomeKey` to ensure marked text state remains in sync across summon and focus cycles.
  - **`RootPaletteView`**: Gated the background placeholder view on `vm.query.isEmpty && !vm.isComposing`.

### Testing

  - [x] Placeholder hides immediately upon typing IME marked text (Pinyin).
  - [x] Re-summoning the palette or clicking "Open Tinycast" from the menu bar while composition is in flight keeps the placeholder hidden.
  - [x] Committing candidates by keyboard or mouse click keeps the placeholder hidden.
  - [x] Clearing the search query restores the placeholder.
  - [x] Cold-summoning the palette / focus transitions show zero vertical 1pt jump.
  - [x] Existing test harnesses pass.

mrasong:
> **测试项:** 拼音输入时占位符立即隐藏；用鼠标点击候选词上屏后占位符保持隐藏；清空输入后占位符恢复；启动后第一次唤起面板、切换模式时占位符**不出现上下 1 点的跳动**。

This works well.

abue-ammar:
> > **测试项:** 拼音输入时占位符立即隐藏；用鼠标点击候选词上屏后占位符保持隐藏；清空输入后占位符恢复；启动后第一次唤起面板、切换模式时占位符**不出现上下 1 点的跳动**。
> 
> This works well.

cool then i will take a look i have fully refactored the search. could you checkout to this pr https://github.com/abue-ammar/tinycast/pull/255 and do `xcodegen generate` and test it? if  1 px placeholder moves up layout shift issue on cold start. and this existing issue solves.

mrasong:
> > > **测试项:** 拼音输入时占位符立即隐藏；用鼠标点击候选词上屏后占位符保持隐藏；清空输入后占位符恢复；启动后第一次唤起面板、切换模式时占位符**不出现上下 1 点的跳动**。
> > 
> > 
> > This works well.
> 
> cool then i will take a look i have fully refactored the search. could you checkout to this pr #255 and do `xcodegen generate` and test it? if 1 px placeholder moves up layout shift issue on cold start. and this existing issue solves.

I'm on it

mrasong:
> > **测试项:** 拼音输入时占位符立即隐藏；用鼠标点击候选词上屏后占位符保持隐藏；清空输入后占位符恢复；启动后第一次唤起面板、切换模式时占位符**不出现上下 1 点的跳动**。
> 
> This works well.


https://github.com/user-attachments/assets/a1dcbd83-86b0-408c-88e7-3c81c7445f6b

abue-ammar:
i will fix this pr once i get home. this pr became old

abue-ammar:
@mrasong this branch is a bit got corrupted after so many changes merged on top. so i make a new branch and adopted all these changes in https://github.com/abue-ammar/tinycast/pull/282. so i am closin this one. also i preserved ur authorship in that pr

PR #255, opened 2026-08-18
Own the palette search field as an NSTextView

abue-ammar:
Replaces the palette's SwiftUI `TextField` with an `NSTextView` the palette owns, which removes the cause of the 1-point placeholder step instead of working around it. Fixes the IME bug reported in #254 as a side effect.

> [!NOTE]
> **Built, gated and hand-verified on macOS 26.** Debug compiles with no new warnings, `./Scripts/run-tests.sh` passes all 33 harnesses, `./Scripts/lint.sh` is clean, and the palette sweep below was walked by hand. The branch was originally written without a Swift toolchain; everything that needed doing has been done.

---

## Why

An `NSTextField` has **two** things that draw the same string:

- unfocused, the **cell** draws it;
- focused, the window's shared **field editor** draws it.

AppKit sizes that editor one point taller than the field (measured: a 24pt editor in a 23pt field) and the two paths round the baseline differently, so the same glyphs sat one point higher while focused. The editor is created lazily and cached on the window, so the step only showed on the **first summon after launch**, and only where the eye could track it — when the outgoing and incoming placeholders share a leading word. A SwiftUI `TextField` is the same machinery, and SwiftUI force-casts its field editor to a private subclass, so a corrected one cannot be vended.

The previous fix avoided the swap: draw the placeholder in SwiftUI as a background `Text` gated on `vm.query.isEmpty`. That pinned the placeholder to the layout, but `vm.query` stays empty during IME composition, so the placeholder stayed visible and overlapped in-flight pinyin — the bug #254 reports.

An `NSTextView` **is** its own editor. One renderer, no swap, nothing to step, and marked text lands in its own storage so the placeholder clears the instant composition starts. Both bugs go away for the same reason.

## What changed

**New — `Tinycast/Palette/PaletteSearchField.swift`**

- `PaletteSearchTextView: NSTextView` — draws its own placeholder in `draw(_:)` **before** `super`, so the caret still draws over it (the same property the old background `Text` had). The origin is `textContainerOrigin` plus the container's `lineFragmentPadding`: the exact origin TextKit gives the first real glyph, so placeholder and query align by construction rather than by a tuned constant.
- `PaletteSearchField: NSViewRepresentable` — binds `text` and `isFocused`, carries the `prompt`, and forwards keys through `onKeyCommand`.
- Single line via `isFieldEditor = true`; horizontal scrolling via an `NSScrollView` with both scrollers and elasticity off.
- Vertical centring is `textContainerInset.height`, recomputed in `layout()` against the view's own bounds. One number, one owner.
- `matchWidthToClipView()` keeps the view at least as wide as its clip view so a click past the last glyph still lands the caret.
- Literal input: quote/dash/text-replacement substitution, spelling correction and smart insert/delete are all off, and `usesFindPanel = false`.

**The caret**

`lineFragmentPadding` is `fieldEditorPadding` (2), not 0. macOS draws the caret as an `NSTextInsertionIndicator` layer 2pt wide **centred** on the insertion point, so a text box flush with its clip view loses the caret's left half at column 0 — it rendered 1pt on an empty query and snapped to 2pt once the first glyph pushed it clear. AppKit's own `NSTextField` field editor pads by exactly 2 for this reason (a bare `NSTextView` uses 5), so this is the padding a field editor is meant to have rather than a constant tuned against the caret. Measured full width at both edges afterwards, including the `typed + 3` field the header shrinks to when an argument accessory sits beside it. `drawPlaceholder` already adds `lineFragmentPadding` to its origin, so placeholder and query still align by construction.

**IME**

- Marked text is in the view's storage, so `string` is non-empty during composition and the placeholder is hidden.
- `textDidChange` returns early while `hasMarkedText()`, so `PaletteState.query` still changes only on commit and a half-typed romanisation never reaches a search.

**Key routing**

- `doCommand(by:)` maps `moveUp:` / `moveDown:` / `moveLeft:` / `moveRight:` (plus their `…AndModifySelection:` and begin/end-of-document spellings), `insertNewline:` / `insertLineBreak:` / `insertNewlineIgnoringFieldEditor:`, `insertTab:` / `insertBacktab:` and `cancelOperation:` onto `RootPaletteView.handleSearchKey`.
- The closure returns `true` to keep the key and `false` to fall through to `super.doCommand(by:)` — that is what keeps ← and → stepping the emoji grid while staying with the caret everywhere else.
- `keyDown` records the modifier flags, because a selector does not carry the chord that produced it.
- The matching `onKeyPress` handlers are **still on the root view**. They serve the inline argument fields, which are SwiftUI's and hold focus themselves. Both paths now call the same five methods (`moveUp`, `moveDown`, `moveHorizontally`, `submit`, `cancel`) instead of duplicating the logic.

**Focus**

- `searchFocused` moved from `@FocusState` to `@State`. It is now intent, driven two-way: `updateNSView` takes first responder when it is `true`, and `becomeFirstResponder` / `resignFirstResponder` report back (deferred through a `Task` so the write does not land inside a SwiftUI update).

**Deletions**

- `PalettePanel.fieldEditorSlack` and the 2pt cursor-rect outset — they existed only because AppKit's field editor overhung the field. The owned view *is* the published frame.
- `Theme.Typography.searchField` (the SwiftUI `Font`), now unused. `searchFieldSize` and `searchFieldNSFont` stay.
- The SwiftUI placeholder `background`, its `allowsHitTesting(false)`, and the explicit `accessibilityLabel` (the text view sets its own).

**Docs**

- `docs/features/palette.md`: the placeholder section is rewritten as "The search field is Tinycast's" (anchor changed); the `lineFragmentPadding` invariant is stated with its reason; cursor-policy, emacs-chord and menu-open-freeze sections lost their field-editor references.
- `docs/ui.md`: "The palette search field" and the typography note.
- `docs/testing.md`: five new palette checks, including the CJK IME one and the caret thickness one.

## Deliberate behaviour changes

- **⌥← / ⌥→ now do word movement in the field** instead of stepping the emoji grid. Previously `onKeyPress(.leftArrow)` fired regardless of modifiers. The new behaviour is more correct for a text field, but it *is* a change — revert by mapping `moveWordLeft:` / `moveWordRight:` in `paletteKey(for:)` if you disagree.
- **The cursor rect is now exact.** If the I-beam flickers at the field's edge, the published frame is wrong — do **not** re-add the slack to paper over it.
- **Glyphs start 2pt in**, as they do in every `NSTextField`. Worth an A/B against `main` if the header's leading metric is meant to be pixel-identical; the 2pt comes off the `headerGutter` before the field if so, never off the padding.

## Explicitly out of scope

The **menu-open input freeze** is untouched. Its stated reason (never resign first responder, because resigning swapped the `NSTextField` between cell and field editor and shifted the text a point) is now obsolete, and `docs/features/palette.md` says so and marks it a candidate for simplification. Changing first-responder behaviour there is a separate, testable change and does not belong in this branch.

## Verification

**Gate**

- [x] `./Scripts/run-tests.sh` — all 33 harnesses pass
- [x] Debug build compiles with no new warnings
- [x] `./Scripts/lint.sh` clean
- [x] `Features/*/Model/` imports no AppKit or SwiftUI

**Field width and scrolling**

- [x] A click well past the last glyph lands the caret
- [x] A click in an empty field lands the caret at position 0
- [x] A query longer than the field scrolls horizontally and keeps the caret visible
- [x] Deleting back down never leaves the field narrower than its clip view

**The bug this branch exists to kill**

- [x] Placeholder and typed text share a baseline
- [x] **First summon after launch**, stepping between screens with different placeholders (`Search…` → `Search apps…`) does not move it a point
- [x] Placeholder colour is `textTertiary` (white 0.40)
- [x] Compact → expanded does not shift the search bar vertically

**Caret**

- [x] Same thickness on an empty query as mid-word, and after deleting back to empty
- [x] I-beam over the field, arrow everywhere else, no flicker along the edge

**IME (#254)**

- [x] The placeholder clears the moment composition starts, and composing text never overlaps it
- [x] The list filters only once the candidate is committed
- [x] Committing by clicking the IME candidate window still publishes the query
- [x] Cancelling composition restores the placeholder

**Focus handoff**

- [x] Reopening the palette focuses the field with an empty query
- [x] Tab walks search field → each argument → back
- [x] Clicking an argument field directly is not stolen back by the search field
- [x] ↵ on a blank required argument focuses it instead of launching
- [x] Moving the selection off a command hands focus back to the search field

**Keys**

- [x] ↑/↓ move the highlight and scroll it into view
- [x] ↓ from the compact bar expands and selects the first row; ↑ moves the caret
- [x] ←/→ step the emoji grid; elsewhere they move the caret
- [x] ⌃N/⌃P/⌃F/⌃B behave as ↓/↑/→/←
- [x] Plain ↵ activates, ⌘↵ runs the secondary action, ⌥↵ pastes keeping the window open
- [x] Escape closes an open menu, then an extension screen, then the palette
- [x] Tab toggles launcher ↔ clipboard when the selection has no arguments
- [x] Bare Backspace on an empty query backs out of a sub-screen
- [x] ⌘. pins, ⌃X deletes, ⌘K opens Actions
- [x] While a menu is open: typing does not change the query, the caret is hidden, and ↑/↓/↵/Esc drive the menu

Closes #254 — this supersedes it. The bug report is what drove this fix, and it is the reporter's in origin.

### DIFF
--- a/Tinycast.xcodeproj/project.pbxproj
+++ b/Tinycast.xcodeproj/project.pbxproj
@@ -153,6 +153,7 @@
 		7FDE88F4E065FF80DCB92445 /* NotesStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = FC2B77DA46386E7E459406E4 /* NotesStore.swift */; };
 		7FF6A0A4702E1C1365F2D76A /* LauncherScreen.swift in Sources */ = {isa = PBXBuildFile; fileRef = B97A8C3E88FEBCBBEADF415B /* LauncherScreen.swift */; };
 		8090FBE343746D9D43EE2B24 /* PalettePanel.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA9F3FA9925873D25EAF7D03 /* PalettePanel.swift */; };
+		8090FBE343746D9D43EE2B25 /* PaletteSearchField.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA9F3FA9925873D25EAF7D04 /* PaletteSearchField.swift */; };
 		80E77985425032856DEECDFD /* CalloutShape.swift in Sources */ = {isa = PBXBuildFile; fileRef = 92AF20F1007F7F3DC1013C2A /* CalloutShape.swift */; };
 		82D95EF7539498A9A484B3D1 /* SystemActionsSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AEEB022D080E6DAF703D8132 /* SystemActionsSettingsView.swift */; };
 		8453CF4C0D8D015A66C9B400 /* HoverArming.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9D4C3D9FA2F775DC650985D8 /* HoverArming.swift */; };
@@ -514,6 +515,7 @@
 		B97A8C3E88FEBCBBEADF415B /* LauncherScreen.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = LauncherScreen.swift; sourceTree = "<group>"; };
 		B9F568F2F6EF857274F1A9D8 /* ExtensionIconCache.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ExtensionIconCache.swift; sourceTree = "<group>"; };
 		BA9F3FA9925873D25EAF7D03 /* PalettePanel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PalettePanel.swift; sourceTree = "<group>"; };
+		BA9F3FA9925873D25EAF7D04 /* PaletteSearchField.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PaletteSearchField.swift; sourceTree = "<group>"; };
 		BDB241B179A08518424F7C1B /* DialogPanel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DialogPanel.swift; sourceTree = "<group>"; };
 		BE3331A5E8F892CC257AD31B /* SnippetMarkdownSerializer.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SnippetMarkdownSerializer.swift; sourceTree = "<group>"; };
 		BE6972F177922B1EB8831735 /* OnboardingCard.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OnboardingCard.swift; sourceTree = "<group>"; };
@@ -1329,6 +1331,7 @@
 				418527DD2F2370A1B82A13B7 /* PaletteDropGuideView.swift */,
 				1534DB79CC23709ABC5E9D9B /* PaletteMode.swift */,
 				BA9F3FA9925873D25EAF7D03 /* PalettePanel.swift */,
+				BA9F3FA9925873D25EAF7D04 /* PaletteSearchField.swift */,
 				802BD2F4CC057228F2528217 /* PalettePlacement.swift */,
 				445B0374FD5D74DBE23243CD /* PaletteScreen.swift */,
 				188A6D3C5B6BC67716F15F83 /* PaletteState.swift */,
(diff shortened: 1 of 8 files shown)


abue-ammar:
image
@mrasong  could you test these if these are fine and perfectly working

mrasong:
I'll try later

mrasong:
Seems there is still something wrong... (ver 0.9.7)

image


https://github.com/user-attachments/assets/238d7fd1-fa9c-4344-86fe-15e3ba963188

PR #267, opened 2026-08-18
Add input source switcher

mishamyrt:
## What changed

 Added an `Auto-switch Input Source` picker to Settings → General. The picker is shown only when more than one keyboard input source is available.

When the palette opens, Tinycast captures the current input source and applies the configured source after the search field receives focus. When the palette closes or Tinycast terminates, the previously active input source is restored.

## Memory footprint

Not measured

## Demo

https://github.com/user-attachments/assets/daf97cb1-ebec-44eb-bb5d-2f5a9fc077c4

## Drawbacks

Input sources are refreshed when Tinycast is reactivated to reduce CPU cycles, not immediately while it remains active.

Input-source enumeration and global restoration use HIToolbox TIS APIs because AppKit does not provide equivalent global lifecycle control.

## Tests & validation

- Ran all 33 test harnesses with `./Scripts/run-tests.sh`.
- Ran `./Scripts/lint.sh`.
- Verified the pure Model layer contains no AppKit, SwiftUI, or Cocoa imports.
- Built the Debug configuration successfully with code signing disabled.
- No new warnings were introduced.
- No new harness cases were added because the behavior depends on the real AppKit input context and installed macOS input sources.

PR #282, opened 2026-08-19
Revert the owned search text view and fix the IME placeholder on the SwiftUI field

abue-ammar:
Reverts #255 and lands #254's fix instead, with the two changes that were merged on top of #255 rewired onto the restored search field.

> [!NOTE]
> Debug builds with no new warnings, `./Scripts/run-tests.sh` passes all 34 harnesses, `./Scripts/lint.sh` is clean, and the purity grep returns nothing. The IME behaviour and the input-source switch still need a hand check — see **Needs manual verification** below.

## Why

#255 replaced the palette's SwiftUI `TextField` with an owned `NSTextView` and folded #254's IME fix in as a side effect. Going back to the `TextField` means the IME bug needs its own fix again — which is what #254 was.

## Commit 1 — revert #255

`57ae056` reverted: `PaletteSearchField.swift` is gone, `RootPaletteView` is back to a `TextField` with the placeholder drawn as a leading-aligned background `Text`, `Theme.Typography.searchField` is back, and `PalettePanel.fieldEditorSlack` with its 2pt cursor-rect outset is back.

Three commits landed on top of #255. Nothing is reverted out from under them:

**#274 (Light appearance)** keeps its tokens. The field's tint is `Theme.Colors.textPrimary`, not the `.white` the pre-#255 code carried; `PalettePanel.setSearchCaretHidden` already read the token and is untouched. `Theme.Colors` itself is not in this diff, so the frozen dark branches and `appearance-test` are unaffected.

**#267 (input-source switcher)** needed rewiring. It applied the switch through `PaletteSearchField.onFocused`, which only existed inside the representable. The context it actually needs is the window's own field editor — which `PalettePanel` already reaches for the menu-open caret policy — so the panel now publishes `fieldEditorContext` and reports it through `onFieldEditorFocused` whenever a field takes focus. The switch stays scoped to that context: a non-activating panel must never select an input source globally.

Two triggers, because focus lands at different times. On a first summon `beginSession` runs before focus, so the `makeFirstResponder` report is what applies it. On a re-summon first responder is still the field editor from last time and never moves, so no focus event fires — `windowDidBecomeKey` applies a pending session directly. `applySession` is idempotent, so whichever runs first wins and the other is a no-op.

**#264** only reflowed a `handleSearchKey` signature that #255 introduced; it goes away with the function.

## Commit 2 — the IME fix (#254)

An IME composes into the field editor's own storage, so the bound `query` stays empty for the whole romanisation and the hand-drawn placeholder sat under the in-flight pinyin.

`PalettePanel` publishes the field editor's `hasMarkedText()` as `PaletteState.isComposing`, and the placeholder is gated on that as well as an empty query. The observation follows first responder, since SwiftUI hands the window's one field editor to whichever field holds focus.

It watches `NSTextView.didChangeSelectionNotification` rather than `NSText.didChangeNotification`. Measured on a bare `NSTextView` — setting marked text, extending it, cancelling it, then committing:

| step | `didChange` | `didChangeSelection` | `hasMarkedText` |
| --- | --- | --- | --- |
| set marked `n` | 0 | 2 | true |
| extend to `ni` | 0 | 2 | true |
| cancel composition | 0 | 2 | false |
| commit `你` | 1 | 1 | false |

`didChange` fires on the commit alone, a whole composition too late. The selection notification is the only one that carries a marked-text change.

This differs from #254's branch in three ways. It observes from a `makeFirstResponder` override rather than also from `paletteState`'s `didSet`, where first responder is never a text view yet and the call was dead. It holds the observer in a `NotificationToken` cleared when first responder stops being a text view. And it collapses #254's three `syncMarkedTextState()` calls in `PaletteWindowController` to one on `windowDidBecomeKey`: the other two sit either side of it on the same summon, and the focus override covers first-mount.

That one is load-bearing, not defensive. Pop to Root defaults to **Immediately**, so `prepare(mode:)` clears the flag on every hide — but with a timeout set, a reopen inside the window keeps the screen state and never moves first responder, so a composition abandoned by clicking away would leave the flag set and the placeholder missing on an empty query. `trackComposition()` re-reads the editor rather than assuming, so marked text that genuinely survived the hide still hides the placeholder.

## Needs manual verification

I have no CJK IME or second input source to drive here.

- [ ] With a CJK IME: the placeholder clears as soon as composition starts, the composing text never overlaps it, cancelling composition brings it back, and the list filters only once the candidate is committed
- [ ] The same on a **second** summon, where first responder never moved
- [ ] Settings → General → the configured input source is applied on summon and restored on hide
- [ ] Placeholder and typed text still share a baseline across screens on the first summon after launch
- [ ] Menu-open freeze: typing does not change the query, the caret is hidden, ↑/↓/↵/Esc drive the menu
- [ ] I-beam over the field, arrow everywhere else

Closes #254. The fix is mrasong's; the commit carries their `Co-authored-by`.


mrasong:
All test passed with v0.9.8. sorry for the late.

abue-ammar:
> All test passed with v0.9.8. sorry for the late.

it's totally fine bro! good to hear that

PR #285, opened 2026-08-19
Reorder favorites, and launch them by ⌘-digit

### LINKED ISSUE #272: [Feature]: Move up and move down favorites
### What should it do?

in tinycast :
Image

in raycast :
Image

we should be able to move up and down favorites. 

PS: add also shortcuts (cmd + 1, cmd +2, ...) to quickly select the favorites without scrolling

When you add an element if you want to hate it in the first you will have to delete every favorite and then remember the correct order you want...

abue-ammar:
## Related issue

Closes #272

## What changed

Two passes over favorites: they can now be **ordered**, and they can be **launched by number** from
either palette size.

### Reordering

`FavoritesStore.keys` has always been the order, but nothing could change it — a new favorite is
appended, so moving one to the top meant removing and re-adding everything after it.

- **Move Favorite Up / Down** in an entry's ⌘K menu, built only in a direction that exists: the
  first favorite has no Up row, the last has no Down.
- **⇧⌘F** on Add / Remove from Favorites, **⌥⌘↑ / ⌥⌘↓** on the moves.
- Reordering shows only at an empty query, where `AppIndex.orderedResults` pins favorites as a
  prefix — the only place the order is observable.

### ⌘-digit slots

The compact bar bound ⌘1–⌘5 to its icons; the expanded launcher bound nothing. One rule now covers
both: **⌘N launches the Nth favorite.**

- `FavoriteSlots` is the whole table — **⌘1…⌘9 then ⌘0 for the tenth**, ten digit keys being the
  physical ceiling. The eleventh favorite stays listed and reorderable, with no chord and no number.
- **Holding ⌘ numbers the favorite rows**, replacing the kind label on the right. Only favorite rows
  are numbered — the chord launches favorites by position, so numbering anything else would
  advertise a dead shortcut.
- The compact **"…" stops being a numbered slot** and becomes a button after the five icons, still
  only shown when there are more than five, so no favorite loses its digit to the overflow. ⌘6–⌘0
  launch favorites the strip has no room to draw; the button and ↓ both expand.

## Non-obvious bits of the diff

- **The chords need no `sendEvent` interception.** AppKit's `StandardKeyBinding.dict` binds ⌘, ⌥, ⌃
  and ⇧⌘ with an arrow, but nothing is bound to ⌥⌘ with one, so it reaches `onKeyPress` on its own.
- **`FavoritesStore.exchange` swaps positions rather than removing and re-inserting.** `keys`
  retains entries `VisibilityStore` hides or that aren't currently indexed — `ordered(_:)` drops
  them with `compactMap` and never prunes them, which is how a favorite survives an unmounted
  volume — so exchanging the two *visible* keys leaves every such key on its own slot.
- **`AppActionsMenu.content` no longer takes `FavoritesStore`.** It takes a `FavoriteActions` built
  by `LauncherScreen`, so every row runs the same call its chord does. This was a real bug mid-
  review: the rows called `favorites.toggle(app)` directly, looked identical on screen, and landed
  the highlight somewhere else, because the store knows nothing about the selection.
- **Where the highlight lands differs by action, on purpose.** A move follows its entry, since that
  position is the point of the action. A toggle stays with the section: the top of Favorites on
  add, the neighbour above the one that left on remove.
- **Both palette sizes share one prefix.** `paletteIsCollapsed` already requires an empty query, so
  compact implies empty implies `favoriteCount` *is* the pinned prefix — which makes the compact
  bar's separate `orderedResults(query: "")` recompute provably redundant. It's gone, along with
  `CompactFavoriteSlot` and its synthetic id.
- **`AppRow` observes `commandHeld` itself.** Reading it any higher attaches it to
  `RootPaletteView`'s body and rebuilds the whole palette on every ⌘ press; read in the row, a press
  re-runs only the rows the `LazyVStack` has realized. The flag clears in `PalettePanel.resignKey`
  rather than `prepare`, which a re-show that preserves state skips entirely, and is deliberately
  not seeded on show so a ⌘-carrying global hotkey doesn't flash numbers on every summon.
- **The arrow handlers moved to the `keys:` form** to read modifiers. They pass
  `phases: [.down, .repeat]` explicitly — that is the bare-key form's default, and dropping it
  silently kills key repeat on ↑/↓.

## Memory footprint

Leak-tested:

## Drawbacks

- **No chord past the tenth favorite.** ⌘0-as-ten is a tab-bar convention that reads a little oddly
  in a list, and there is no eleventh digit key; reaching further would mean a second modifier.
- The move rows close the menu like every other row, so walking an entry several places from the
  menu means reopening it. The chord is the repeatable path, and works with the menu closed.
- The compact strip still draws five icons, so ⌘6–⌘0 launch favorites that aren't visible there.
- Reordering is keyboard/menu only — no drag.
- `FavoritesStore` itself has no harness: it reaches `AppEntry`, which lives in `AppIndex.swift`
  behind `import AppKit`, so it can't compile into a pure one without moving `AppEntry` out.

## Tests & validation

- `./Scripts/run-tests.sh` — 35 harnesses pass, including a new `favorites-test` over
  `FavoriteSlots` that pins the ⌘0→index 9 off-by-one and round-trips every slot.
- Debug build with no new warnings; `./Scripts/lint.sh` clean; the `Features/*/Model/` import grep
  returns nothing (which is what keeps the new `FavoriteSlots` Foundation-only).
- By hand: the Up/Down rows appearing only where they apply; both chords with the menu open and
  closed; key repeat on plain ↑/↓ and on ⌥⌘↑/↓; the toggle's landing spot from both the chord and
  the menu; ⌘1–⌘0 in expanded and compact; the ⌘-hold numbering following a reorder; the "…"
  appearing only past five favorites; order surviving a relaunch.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -72,6 +72,7 @@ run file-search-session-test Tinycast/Platform/Signposts.swift \
                              Tinycast/Features/FileSearch/Service/*.swift
 run ranking-test           $L/SearchRelevance.swift $L/LauncherRankingStore.swift
 run scopes-test            $L/SearchScopes.swift
+run favorites-test         $L/FavoriteSlots.swift
 run calc-test              Tinycast/Features/Calculator/Model/*.swift
 run clipboard-test         Tinycast/Features/Clipboard/Model/ClipboardStore.swift \
                            Tinycast/Features/Clipboard/Model/ClipboardFilter.swift
--- a/Tests/favorites-test.swift
+++ b/Tests/favorites-test.swift
@@ -1,0 +1,43 @@
+import Foundation
+
+@main
+struct FavoritesTest {
+    static func main() {
+        var failures = 0
+
+        func check(_ description: String, _ condition: @autoclosure () -> Bool) {
+            if condition() {
+                print("PASS  \(description)")
+            } else {
+                print("FAIL  \(description)")
+                failures += 1
+            }
+        }
+
+        // Ten digit keys is the ceiling: there is no ⌘10, so the eleventh favorite gets no chord.
+        check("ten slots", FavoriteSlots.digits.count == 10)
+        check("no digit repeats", Set(FavoriteSlots.digits).count == FavoriteSlots.digits.count)
+
+        check("⌘1 is the first favorite", FavoriteSlots.index(for: "1") == 0)
+        check("⌘9 is the ninth", FavoriteSlots.index(for: "9") == 8)
+        check("⌘0 is the tenth, not the first", FavoriteSlots.index(for: "0") == 9)
+        check("a letter is not a slot", FavoriteSlots.index(for: "a") == nil)
+
+        check("the first row shows 1", FavoriteSlots.digit(at: 0) == "1")
+        check("the tenth row shows 0", FavoriteSlots.digit(at: 9) == "0")
+        check("the eleventh row shows nothing", FavoriteSlots.digit(at: 10) == nil)
+        check("a negative index shows nothing", FavoriteSlots.digit(at: -1) == nil)
+
+        // The row overlay and the key handler must agree, or a row advertises another's chord.
+        for index in FavoriteSlots.digits.indices {
+            guard let digit = FavoriteSlots.digit(at: index) else {
+                check("slot \(index) has a digit", false)
+                continue
+            }
+            check("slot \(index) round-trips", FavoriteSlots.index(for: digit) == index)
+        }
+
+        print(failures == 0 ? "\nALL PASSED" : "\n\(failures) failed")
+        if failures > 0 { exit(1) }
+    }
+}
--- a/Tinycast.xcodeproj/project.pbxproj
+++ b/Tinycast.xcodeproj/project.pbxproj
@@ -151,6 +151,7 @@
 		74ACDD85B9F03C36E9258695 /* SettingsSplitViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = B7AA1305D43303E055DEE1E9 /* SettingsSplitViewController.swift */; };
 		74BE9E8319C8302EF525E6B2 /* QuicklinkStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 685FB7FD7E4E5BE6606440D0 /* QuicklinkStore.swift */; };
 		75AED41394E66D6E9CBDA7EA /* NotesRepository.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7F84B7743B803F55C41D5C4A /* NotesRepository.swift */; };
+		75EF3CFF00A2215D57A4FBFF /* FavoriteSlots.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3FE591549F24AE554C13CD0C /* FavoriteSlots.swift */; };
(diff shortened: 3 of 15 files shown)


mathiasdotdev:
@abue-ammar I have tried the ⌘-digit and I don't know it's not working for me... you still can click or select the favorite by scrolling to them but  if I press ⌘-3 nothing happens (it should trigger the window command "maximize")

mathiasdotdev:
should also be the case for clipboard history: you should trigger pinned elements with shortcuts ⌘-1, ⌘-2 or ⌘-3

mathiasdotdev:
> should also be the case for clipboard history: you should trigger pinned elements with shortcuts ⌘-1, ⌘-2 or ⌘-3

do you want me to create an issue for that ?

abue-ammar:
> @abue-ammar I have tried the cmd digit and I don't know it's not working for me... you still can click or select the favorite by scrolling to them but if I press cmd-3 nothing happens (it should trigger the window command "maximize")

which keyboard layout do you have. is it English? how about you should a video? or try quit reopening the app? and i am not built another raycast. so keep the expectation low for a free and open source app. if you love raycast stick to it. this app is only for haters. i WON'T bring everything from raycast. clipboard history stays as it's. make a separate issue pointing with proper video, steps to produce. and mention keyboard layout. i will take a look

mathiasdotdev:
I have a french azerty keyboard.

I don't want to bring everything from raycast just the small shortcuts that simplify the daily uses...

if I scroll down in the clipboard history and close THEN open again I'm at the same position... so I have to scroll back to the top ... if I do ⌘-1, it's immediate.

abue-ammar:
> I have a french azerty keyboard.
> 
> I don't want to bring everything from raycast just the small shortcuts that simplify the daily uses...
> 
> if I scroll down in the clipboard history and close THEN open again I'm at the same position... so I have to scroll back to the top ... if I do ⌘-1, it's immediate.

yeah i had a feeling it will be a separate keyboard layout. could you try with English? if possible make a new issues with proper infos

mathiasdotdev:
oooh ok it's working : I have to type everything (window command instead of just window or window co...)

mathiasdotdev:
> > I have a french azerty keyboard.
> > 
> > I don't want to bring everything from raycast just the small shortcuts that simplify the daily uses...
> > 
> > if I scroll down in the clipboard history and close THEN open again I'm at the same position... so I have to scroll back to the top ... if I do ⌘-1, it's immediate.
> 
> yeah i had a feeling it will be a separate keyboard layout. could you try with English? if possible make a new issues with proper infos

i'll try 👍

abue-ammar:
> oooh ok it's working : I have to type everything (window command instead of just window or window co...)

i clearly said that you gotta write the whole thing and because of this we have a ranking algorithm so frequently typed words gets higher ranking. and search by category breaks that pattern. so i had to ignore ranking if category = app name is an exact match. otherwise ranking algorithm will behave odd

mathiasdotdev:
Ok working with the English keyboard for the ⌘-1, ⌘-2, ... with favorites

mathiasdotdev:
> > oooh ok it's working : I have to type everything (window command instead of just window or window co...)
> 
> i clearly said that you gotta write the whole thing and because of this we have a ranking algorithm so frequently typed words gets higher ranking. and search by category breaks that pattern. so i had to ignore ranking if category = app name is an exact match. otherwise ranking algorithm will behave odd

sorry did not catch that 😖

PR #287, opened 2026-08-19
Anchor the launcher to the menu-bar display when the cursor setting is off

### LINKED ISSUE #270: [Bug]: Launcher - follow the cursor across screens
### Existing issues

### What happened?

When I want the launcher to stay on the main screen, I uncheck the option that follows the cursor across different screens... but it doesn't work.

### Steps to reproduce

1. make sure to uncheck "follow the cursor across displays"
2. move your cursor to another screen
3. notices that the launcher does not stay on the main screen but moves to other screens

### Screenshots or recording

Image

0.9.5 stable

### macOS version

26.5.2

abue-ammar:
Fixes #270.

## The bug

Unchecking **Follow the cursor across displays** changed nothing — the launcher kept opening on whichever display the pointer was on.

The off branch of `targetScreen()` returned `NSScreen.main`, and both the code comment and `palette.md` asserted that `main` means "the menu-bar display". It does not. AppKit documents `NSScreen.main` as *the screen containing the window with keyboard focus*, so it follows the frontmost app's window across displays — precisely what turning the setting off is meant to stop. The setting was implemented with an API that follows the user.

## The fix

- **`NSScreen.primary`** (new, in `ScreenTarget.swift`) — the screen whose `frame.origin` is `.zero`, i.e. the one with the menu bar, with `screens.first` as the fallback so it is nil only when there are no displays.
- **`targetScreen()`** collapses to `openOnCursorScreen ? .underCursor : .primary`, calling the existing `underCursor` helper instead of duplicating its `NSMouseInRect` hit test and the reasoning behind it.
- **`MessageHUDController`** carried the identical `.main`, so the confirmation pill drifted off the launcher's display whenever the setting was off. Both surfaces now name the same screen.
- **`CursorScreen.swift` → `ScreenTarget.swift`** — the file answers which display a surface targets, and half that answer is no longer about the cursor.
- `palette.md` had the false premise written down as an explanation; corrected, along with the `Platform/` file list in `architecture.md`.

A remembered drag position still outranks the setting, unchanged: `PalettePlacement.restored` never consulted a display preference. Worth repeating to anyone who hits this — if the launcher was ever dragged onto a second display it will keep opening there, and snapping it back onto the drop guides clears the stored position.

## Verification

- `xcodegen generate` re-run for the rename; `Tinycast.xcodeproj` committed alongside.
- Debug build succeeds, no new warnings.
- `./Scripts/run-tests.sh` — all 35 harnesses pass.
- `./Scripts/lint.sh` clean; `Features/*/Model/` import grep clean.

**Not verified on hardware:** I have one display, so I could not watch the launcher stay put across two. On a single display `main` and `primary` resolve to the same screen — I confirmed that by dumping both — which makes this provably a no-op for single-monitor users. The two-display behaviour rests on the documented contract of `NSScreen.main` versus `screens[0]`. Confirmation from anyone with a second monitor would be welcome.

### DIFF
--- a/Tinycast.xcodeproj/project.pbxproj
+++ b/Tinycast.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		0047FE6A50E746CCF0FC6810 /* SystemActionRunner.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2AEDBB526AB2C0406D8226D5 /* SystemActionRunner.swift */; };
+		00D813D3D99D0856E743168B /* ScreenTarget.swift in Sources */ = {isa = PBXBuildFile; fileRef = A7C7C096F258D1EC8B42F818 /* ScreenTarget.swift */; };
 		0120AF7AB3AD249C484C8A03 /* ExtensionStoreResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7732FB0ED92C8FBFE22599CE /* ExtensionStoreResponse.swift */; };
 		017FB0EDDC9AF28ED81EA6A8 /* EntryIconView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 08279257B0F5995E86FF2C03 /* EntryIconView.swift */; };
 		02452CD187D25FD9CD3EF2CF /* QuicklinkLauncher.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1B865FF16D3C395A8D509B94 /* QuicklinkLauncher.swift */; };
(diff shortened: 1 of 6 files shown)


mathiasdotdev:
@abue-ammar how can I test this to see if the fix is good ? I'll tell you right away... (related issue:  #270)

abue-ammar:
> @abue-ammar how can I test this to see if the fix is good ? I'll tell you right away... (related issue: #270)

try it in 0.9.8 stable it is now CI building in progress

mathiasdotdev:
@abue-ammar all good in v0.9.8👍

Issue #292, opened 2026-08-20
[Feature]: clipboard history - shortcuts pinned elements


mathiasdotdev:
### What should it do?

Adding shortcut to clipboard history for pinned elements would be really great.

### Why does it belong in Tinycast?

When you scroll through your clipboard history, copy a value AND immediately open it again, you will need to scroll to the top to copy the pinned item. If we add a shortcut (similar to the one for the favorite in the launcher : ⌘-1, ⌘-2, ...) the user experience would be super smooth

abue-ammar:
sorry but this stays as it is. aint planning on adding for clipboard history

mathiasdotdev:
@abue-ammar so, is there no way to make it jump to the top when I open the clipboard history? Just so I don't have to scroll every time...

abue-ammar:
> [@abue-ammar](https://github.com/abue-ammar) so, is there no way to make it jump to the top when I open the clipboard history? Just so I don't have to scroll every time...

yeah

mathiasdotdev:
> > [@abue-ammar](https://github.com/abue-ammar) so, is there no way to make it jump to the top when I open the clipboard history? Just so I don't have to scroll every time...
> 
> yeah

how do you do that ? is there an option somewhere ?

abue-ammar:
> > > [@abue-ammar](https://github.com/abue-ammar) so, is there no way to make it jump to the top when I open the clipboard history? Just so I don't have to scroll every time...
> > 
> > 
> > yeah
> 
> how do you do that ? is there an option somewhere ?

nope not at this moment probably never will but probably i will add ocr based search. but clipboard history it stays as it is. you you are unsatisfied you can either stick to so good raycast or maintain a fork of tinycast staying under same license and modify it to your likings even u can add ur desired keyboard layout

mathiasdotdev:
Dude, I didn't mean to be offensive... I think this project is great... if I wanted to stick with Raycast I wouldn't have given this feedback.

I understand the idea of keeping it minimalist on this project; I'm also in the process of not redoing all of Raycast's features... but I didn't think these features were too extensive for this project....

abue-ammar:
@mathiasdotdev 
do a pr. if it is something ground breaking it will not be merged

Issue #291, opened 2026-08-20
[Feature]: Favorites - allow shortcuts with other keyboard layouts


mathiasdotdev:
## Request
When I open the launcher and press ⌘-1, ⌘-2, ⌘-3, ... it does not work because I have a french keyboard layout.

I want to use my french keyboard layout and be able to open any favorite using the shortcut ⌘-1 to ⌘-0

## More informations
french keyboard layout : 
<img width="1406" height="380" alt="Image" src="https://github.com/user-attachments/assets/85734b26-5537-4df9-a8da-6b1d123b95c0" />

english keyboard layout :
<img width="1154" height="848" alt="Image" src="https://github.com/user-attachments/assets/ab964b22-6603-42f0-a6b7-a7224e939055" />

### Why does it belong in Tinycast?

not being able to use the shortcut when using a french keyboard is a pity... we should make this resilient to multiple keyboard or maybe add something to map the 10 favorites shortcuts with the keyboard ?

that way you won't be dependent on a particular keyboard

mathiasdotdev:
linked to #285

abue-ammar:
1st of all tinycast is english us only and we will not have localization. i ignored lots of chinese related keyboard layouts it the past. but still considering to investigate. i would like to know how it behaves in raycast v2 for you. a video would be much appreciated

mathiasdotdev:
https://github.com/user-attachments/assets/d82664de-29ee-43b6-ab03-b4331c3c9297

as you can see with raycast you can do with a french keyboard layout the ⌘-3 and it works

<img width="1742" height="1114" alt="Image" src="https://github.com/user-attachments/assets/c03ef196-f14e-4e58-92a4-c34068c935ba" />

mathiasdotdev:
I don't have raycast v2 installed it's the v1 

<img width="892" height="306" alt="Image" src="https://github.com/user-attachments/assets/89c1dbcd-1a14-4aa5-bd50-a515d023d16f" />

abue-ammar:
i will look into it. if it's a huge breaking feature. i might not implement it. because tinycast is english us only. no localization either.

mathiasdotdev:
That's why I thought about having 10 shortcuts that could be manually added in the settings, so it wouldn't be tied to a specific keyboard.

abue-ammar:
i can't satisfy 100% of the users. i will choose the majority.
| Rank  | Layout family                                                              |          Estimated global users |
[...]

PR #315, opened 2026-08-22
feat: add shared shortcuts for launcher favorites and pinned clipboard items

mathiasdotdev:
## Related issue

Related to #291 #292

## What changed

This PR adds a shared `⌘`-number shortcut system for both launcher favorites and pinned clipboard entries.

### 1. Physical number-row shortcut mapping

`FavoriteSlots` now maps physical macOS number-row key codes to ten slots:

- `⌘1` through `⌘9`
- `⌘0` as the tenth slot

The mapping uses hardware key codes rather than the character produced by the keyboard layout. This makes the shortcuts work consistently on layouts such as QWERTY and AZERTY.

The existing digit-based mapping remains responsible for displaying slot labels, while the new key-code mapping resolves actual keyboard events.

### 2. AppKit handles the shortcuts

The shortcut detection was moved from SwiftUI's `onKeyPress` handling into `PaletteWindowController`.

When a non-repeated command-only number-row event is received:

1. The key code is converted into a favorite slot index.
2. The event is consumed by the palette window.
3. The resolved index is published through `PaletteState`.

This avoids keyboard-layout-dependent behavior and ensures the shortcuts are intercepted before AppKit or the field editor handles them.

### 3. `PaletteState` forwards shortcut events

`PaletteState` now contains:

- `favoriteSlotToken`, a change token used to notify SwiftUI.
- `favoriteSlotIndex`, the resolved slot index.
- `noteFavoriteSlot(_:)`, which stores the index and generates a new token.

`RootPaletteView` observes that token and dispatches the action to the currently active screen.

This preserves the existing architecture: AppKit captures the low-level keyboard event, while the active SwiftUI screen decides what the slot means.

### 4. Launcher favorite behavior

On `LauncherScreen`, `⌘1…⌘0` still launches favorites by position.

The previous SwiftUI key handler was removed. The behavior remains available in both palette sizes, subject to the existing rule that favorites must be enabled in compact mode.

Favorites beyond the tenth item remain visible and reorderable, but do not receive a keyboard slot.

### 5. Pinned clipboard shortcuts

`ClipboardScreen` now supports the same ten shortcuts through `activatePinned(at:)`.

A new `ClipboardStore.pinnedItem(at:in:filter:)` method resolves the requested item from the visible result set:

- The current search query is applied.
- The current clipboard type filter is applied.
- Only the visible pinned section is considered.
- Slot numbering follows the visible pinned order.
- Missing slots are ignored without producing an action.

For example, after filtering for `beta`, `⌘1` selects the first visible pinned result containing `beta`, rather than the first pinned item in the entire history.

### 6. Clipboard row UI

Pinned clipboard rows now receive an optional slot label. When the Command key is held, the first ten visible pinned rows display shortcuts such as:

- `⌘ 1`
- `⌘ 2`
- …
- `⌘ 0`

The labels are hidden otherwise, so the normal clipboard list remains visually compact.

## Memory footprint

| Step | Memory |
| --- | ---: |
| Idle after launch | 71.9 MB |
| Palette open | 77.5 MB |
| Feature in use (peak) | 81.4 MB |
| Palette closed, settled | 80.9 MB |

Leak-tested:

### Before

2026-08-22 at 23 57 35

https://github.com/user-attachments/assets/19acd787-ee49-4d8b-a892-465dec193538

As you can see the previous version was not compatible with multiple keyboard layout (espacially with French Layout (Azerty).

### After : using AppKit instead of SwiftUI's `onKeyPress`

The mapping uses hardware key codes rather than the character produced by the keyboard layout. This makes the shortcuts work consistently on layouts such as QWERTY and AZERTY.

https://github.com/user-attachments/assets/2f993a91-f453-4908-942a-391850bcc829

## Drawbacks

None known.

### Added tests

The commit adds coverage in two standalone harnesses.

#### `Tests/favorites-test.swift`

The favorites harness now verifies that:

- Physical key code `18` resolves to the first slot.
- Physical key code `25` resolves to the ninth slot.
- Physical key code `29` resolves to the tenth slot (`⌘0`).
- An unknown key code does not resolve to a slot.

The existing tests still verify the ten-slot limit, the `1…9,0` display order, the absence of an eleventh slot, negative-index handling, and the round-trip agreement between displayed digits and slot indices.

#### `Tests/clipboard-test.swift`

The new `pinnedSlotResolutionUsesVisiblePins()` case verifies that:

- The first and third slots resolve to the corresponding pinned rows.
- A slot beyond the visible pinned rows returns `nil`.
- A search query is applied before slot numbering.
- The visible order is preserved after filtering.
- A type filter is applied before slot numbering.
- Filtering can make a previously available slot unreachable.

Together, these tests protect the contract that shortcuts operate on the visible pinned block, rather than on the complete unfiltered clipboard history.

### Commands

```sh
./Scripts/run-tests.sh favorites-test
./Scripts/run-tests.sh clipboard-test
./Scripts/lint.sh
```

The tests exercise the shipped model sources through the repository's standalone harnesses. Manual validation should also confirm that `⌘1…⌘0` launches the expected favorite or pastes the expected visible pinned clipboard item, including with a query active and with the palette in compact mode.


Tinycast.xcodeproj/project.pbxproj
```diff
@@ -2154,6 +2050,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = tinycast;
 				CODE_SIGN_ENTITLEMENTS = Tinycast/Tinycast.entitlements;
+				"CODE_SIGN_IDENTITY[sdk=macosx*]" = "-";
```
abue-ammar, line 2209: why do we need it? i dont think we need this line

Tinycast/Features/CustomCommands/Service/ShellCommandRunner.swift
```diff
@@ -31,7 +31,7 @@ enum ShellCommandRunner {
         let process = Process()
         process.executableURL = URL(fileURLWithPath: "/bin/zsh")
         // zsh reads `.zshrc` only for interactive shells, so `-l` alone sees no aliases.
-        process.arguments = [loadingShellEnvironment ? "-ilc" : "-lc", command]
+        process.arguments = [loadingShellEnvironment ? "-il" : "-l", "+m", "-c", command]
```
abue-ammar, line 34: why are we changing shell? this is not related

abue-ammar (commented):
Thanks for this. The core idea is right: matching the physical number row instead of the typed character is the correct fix, and the AZERTY bug is real. The clipboard half looks sound. Three things need to change before this can land.

## 1. Extension ⌘-number shortcuts stop working

`onCommandShortcut` now consumes **every** command-only number-row key and returns `true`:

```swift
if let index = FavoriteSlots.index(forKeyCode: event.keyCode) {
    self.core.palette.noteFavoriteSlot(index)
    return true
}
```

That handler runs inside `PalettePanel.sendEvent`, before `super.sendEvent`, so it wins over every SwiftUI `onKeyPress`. A Raycast extension can declare `{ modifiers: ["cmd"], key: "1" }`, and `ExtensionScreen.keyEquivalent` maps any single character through its `default` branch. Those shortcuts used to reach `ExtensionCommandScreen.dispatchShortcut`. With this change they never arrive.

## 2. The slots are swallowed on every other screen too

Snippets, quicklinks, emoji, calculator history and the quicklink-argument form all consume ⌘1…⌘0 now, and `activateFavoriteSlotShortcut` does nothing for them. Same in compact mode when `showFavoritesInCompactMode` is off: the old `onKeyPress` returned `.ignored` there and let the key pass, the new path eats it.

Both this and (1) come from the same shape: AppKit decides to consume before anything knows whether the slot means something. Two ways out, either is fine:

- Gate the interception on `core.palette.mode` so only launcher and clipboard intercept.
- Or have the screen report back, and only consume when it actually handled the slot.

## 3. Two unrelated changes are riding along

Both have an inline comment below. The reasoning:

**`project.pbxproj`** gains `CODE_SIGN_IDENTITY[sdk=macosx*] = "-"` on the Debug config, so `Tinycast Dev.app` becomes ad-hoc signed instead of using `CODE_SIGN_IDENTITY: "Tinycast Self-Signed"` from `project.yml`. Two problems: XcodeGen owns the project, so the next `xcodegen generate` deletes it; and while it stands, every Debug build gets a fresh identity, so the Dev app's Accessibility grant resets each time. Please drop it.

**`ShellCommandRunner`** goes from `-ilc` to `-il`, `+m`, `-c`. Turning off job control may well be a genuine fix, but it changes how every custom command runs, it is not mentioned in the description, it has no test, and the comment above the line now explains only half of the flags. Please split it into its own PR.

## Smaller notes

- The handler now requires modifiers to be exactly `.command`, so ⌘⇧1 no longer launches a favorite. The old `press.modifiers.contains(.command)` allowed it. Probably what you want, worth one line in the description.
- `favoriteSlotIndex` is never cleared after use. Harmless with the token pattern, just a stale value that lingers.

## What looks good

`pinnedItem(at:in:filter:)` reuses the cached `search`, and `prefix(while: \.isPinned)` matches the existing invariant that pins lead the results. The row-label counter in `ClipboardList.rows` counts pins the same way, so the labels and the actions cannot disagree. Test coverage for the query and filter cases is the right contract to lock down.

*Review by reading only: no macOS toolchain on hand, so the build, `run-tests.sh` and `lint.sh` are unverified here.*

abue-ammar:
@mathiasdotdev  test everything in english keybaord manually . remove unrelated changes QWERTY is  no. priority. if these changes breaks QWERTY probably i will not go with these changes

mathiasdotdev:
I'll ping you when it's done

Tinycast/Features/CustomCommands/Service/ShellCommandRunner.swift
mathiasdotdev: if I don't do that, the script ./Scripts/run-tests.sh is not finishing :

```shell
# display-name ranking
PASS  'chrome' top is Google Chrome
PASS  'chrome' does not include Chess
PASS  'ch' includes Google Chrome
PASS  'ch' includes Chess
PASS  'ch' ranks Chess (prefix) above Chrome
PASS  'saf' top is Safari
PASS  'tm' includes Time Machine
PASS  'code' includes Visual Studio Code
PASS  'terminal' exact top
PASS  'xyz' matches nothing
PASS  shorter Wick wins the default prefix tie
PASS  learned boost promotes WhatsApp within the prefix tier
PASS  invisible format mark does not demote a prefix match
PASS  learned marked WhatsApp can outrank Wick
PASS  exact tier
PASS  prefix tier
PASS  word-start tier
PASS  substring tier
PASS  subsequence tier
PASS  no match is nil
PASS  only subsequence is non-literal

# field priority
PASS  'codex' finds ChatGPT at all
PASS  'codex' ranks the exact alias above a subsequence display-name hit
PASS  'code' ranks the prefix display name above the alias holder
PASS  'ical' finds Calendar by alias
PASS  'ibooks' finds Books by alias
PASS  'address book' finds Contacts by alias
PASS  'browser' finds Safari by alias
PASS  non-Latin alias matches
PASS  'system preferences' finds System Settings by alias
PASS  bands are strictly ordered: user-alias > name > alias > name-fuzzy > alias-fuzzy > id > exec
PASS  every band is a whole stride apart
PASS  Safari's exact display name beats its own alias band
PASS  an entry with no matching field scores nil

# user aliases
PASS  'fg' finds Figma by user alias
PASS  the user alias sits in the band above the display name
PASS  a user alias outranks another entry's exact display name
PASS  a user alias matches literally: exact, prefix and substring
PASS  a user alias never subsequence-matches
PASS  the strongest field still wins on an aliased entry
PASS  an inside alias hit does not beat another entry's own prefix
PASS  ...but the aliased entry is still findable
PASS  an alias prefix hit still ranks first
PASS  an inside alias hit ranks in the vendor-alias band

# alternate-name sanitizing
PASS  'app' still finds App Store
PASS  'app' matches nothing by display name that isn't a real hit
PASS  '.app' alternates are dropped entirely
PASS  'alternate' matches nothing
PASS  the ALL_CAPS placeholder is dropped
PASS  an alternate repeating the display name is dropped
PASS  real aliases survive
PASS  empty and whitespace-only names are dropped
PASS  case-insensitive dedupe keeps the first spelling
PASS  names are trimmed
PASS  a name matching the file name but not the display name is still dropped
PASS  an ALL_CAPS name without an underscore is kept
PASS  a multi-word name with an underscore is kept

# identifier fields
PASS  bundle-id vendor component matches
PASS  the trimmed bundle id matches as a prefix
PASS  a pasted full bundle id matches
PASS  bundle ids do not subsequence-match
PASS  a short query does not drag in every reverse-DNS id
PASS  'com' matches nothing by bundle id
PASS  'co' matches nothing by bundle id
PASS  'com.' matches nothing by bundle id
PASS  a bundle id with no dot still matches
PASS  executable name matches literally
PASS  executable name does not subsequence-match
PASS  an entry with no bundle id or executable still matches on its name
PASS  ...and matches nothing else

# edge cases
PASS  empty query scores 0
PASS  empty query never returns nil for any entry
PASS  a query longer than every candidate matches nothing
PASS  emoji query does not trap
PASS  an RTL query does not trap
PASS  a format-scalar-only query is treated as empty
PASS  an entry with every field empty matches nothing
PASS  a snippet keyword ranks at display-name strength
PASS  ranking is deterministic across repeats
PASS  the learned boost cap cannot cross a FuzzyMatch tier
PASS  the learned boost cap cannot cross a band

# randomized property loop
PASS  scores always sit inside their band
PASS  the max learned boost never lifts a score out of its band
PASS  scoring is deterministic
PASS  rank order is stable
PASS  the loop actually exercised matches
PASS  the same text always scores lower in a weaker field

ALL PASSED
ok    fuzz-test             
File search tests passed
ok    file-search-test      
File search session tests passed
ok    file-search-session-test
PASS  query key trims surrounding whitespace
PASS  query key folds case
PASS  query key folds diacritics
PASS  query key is locale-independent
PASS  unvisited result has no boost
PASS  an unlearned query yields an empty table
PASS  a learned query is recalled through its normalized key
PASS  visit teaches first query prefix
PASS  visit teaches full normalized query
PASS  visit does not teach a different query
PASS  one visit, same day, scores 2100
PASS  ten visits, same day, score 3576
PASS  ten visits, one half-life later, score 2627
PASS  ten visits, a year later, score 2076
PASS  the blend is capped at 4500
PASS  frequency increases the boost
PASS  recency decays over time
PASS  a newer habit can overtake an old frequent result
PASS  one pass returns every item learned for the query
PASS  records persist across store instances
PASS  per-item reset clears every learned query
PASS  per-item reset preserves other items
PASS  global reset clears all learned ranking
PASS  the learned table is saturated for this query
PASS  a saturated boost cannot lift an alias hit over a display-name hit
PASS  a saturated boost cannot lift an identifier hit over an alias hit
PASS  a saturated boost cannot cross a FuzzyMatch tier
PASS  a saturated boost still reorders within one tier
PASS  the observed boost cap stays under a band stride

ALL PASSED
ok    ranking-test          
PASS  direct and one-level-nested .app children are indexed
PASS  non-app children are skipped
PASS  hidden bundles are skipped
PASS  bundles nested two levels deep are not indexed
PASS  a deeply nested folder works as its own scope
PASS  an .app scope is indexed directly
PASS  a missing .app scope yields nothing
PASS  a missing directory scope is skipped without failing the rest
PASS  scopes are scanned in order
PASS  expand resolves a tilde
PASS  abbreviate restores the tilde
PASS  tilde survives a round trip
PASS  expand leaves an absolute path alone
PASS  a trailing slash is trimmed
PASS  root survives trimming
PASS  normalize dedups after abbreviating
PASS  normalize preserves order
PASS  normalize drops blanks
PASS  defaults are already normalized

ALL PASSED
ok    scopes-test           
PASS  ten slots
PASS  no digit repeats
PASS  ⌘1 is the first favorite
PASS  ⌘9 is the ninth
PASS  ⌘0 is the tenth, not the first
PASS  a letter is not a slot
PASS  keycode 1 maps to first slot
PASS  keycode 9 maps to ninth slot
PASS  keycode 0 maps to tenth slot
PASS  unknown keycode is not a slot
PASS  the first row shows 1
PASS  the tenth row shows 0
PASS  the eleventh row shows nothing
PASS  a negative index shows nothing
PASS  slot 0 round-trips
PASS  slot 1 round-trips
PASS  slot 2 round-trips
PASS  slot 3 round-trips
PASS  slot 4 round-trips
PASS  slot 5 round-trips
PASS  slot 6 round-trips
PASS  slot 7 round-trips
PASS  slot 8 round-trips
PASS  slot 9 round-trips

ALL PASSED
ok    favorites-test        

500 passed, 0 failed
ok    calc-test             
69/69 passed
ok    clipboard-test        
emoji-test: all checks passed (2100 records)
ok    emoji-test            
111684 passed, 0 failed
ok    palette-selection-test
# dark branches are the shipped literals
# tokens that absorbed a literal duplicated across views
# every surface token resolves per appearance
# the scrim inverts rather than ramping: it lightens the light surface
# .system hands the choice back to AppKit

43 passed, 0 failed
ok    appearance-test       
24 passed, 0 failed
ok    palette-placement-test
15 passed, 0 failed
ok    scroll-reveal-test    
14 passed, 0 failed
ok    hover-arming-test     
69 passed, 0 failed
ok    hotkey-test           
30 passed, 0 failed
ok    callout-test          
Icon cache tests passed
ok    icon-cache-test       
PASS  each icon prints uniquely: 9/9
PASS  the same symbol is equal
PASS  two symbols differ
PASS  two tints of one symbol differ
PASS  a tinted tile differs from the plain one
PASS  one file at two extents differs
PASS  hashing collapses only equal values
PASS  file draws something
PASS  symbol draws something
PASS  tintedSymbol draws something
PASS  all three rasterize
PASS  a tinted tile differs from the plain one
PASS  two tints do not share a cache entry
PASS  a smaller extent draws smaller ink: 0.6041666666666666 vs 0.9166666666666666
PASS  two extents do not share a cache entry
PASS  the same icon is stable across two asks
PASS  a cold icon is not reported warm
PASS  a drawn icon is reported warm
Entry icon tests passed
ok    entry-icon-test       
Extension icon tests passed
ok    ext-icon-test         
76 passed, 0 failed
ok    system-action-test    
81 passed, 0 failed
ok    volume-test           
325 passed, 0 failed
ok    window-command-test   
121 passed, 0 failed
ok    space-gesture-test    
PASS  add trims the name
PASS  add trims the command
PASS  add keeps the flags
PASS  the entry id round-trips to the UUID
PASS  a name differing only in case is rejected
PASS  update keeps the id
PASS  update applies the new name
PASS  update applies a flag
PASS  update clears a flag left out of the draft
PASS  commands survive a reload with their flags
PASS  import preserves every flag
PASS  a zero exit reports success
PASS  commands start in the user's home directory
PASS  the TINYCAST marker is exported so a shell config can detect us
PASS  a non-zero exit reports its status and stderr
```

it stays here forever...

abue-ammar:
> Thanks for this. The core idea is right: matching the physical number row instead of the typed character is the correct fix, and the AZERTY bug is real. The clipboard half looks sound. Three things need to change before this can land.
> 
> ## 1. Extension ⌘-number shortcuts stop working
> `onCommandShortcut` now consumes **every** command-only number-row key and returns `true`:
> 
> ```swift
> if let index = FavoriteSlots.index(forKeyCode: event.keyCode) {
>     self.core.palette.noteFavoriteSlot(index)
>     return true
> }
> ```
> 
> That handler runs inside `PalettePanel.sendEvent`, before `super.sendEvent`, so it wins over every SwiftUI `onKeyPress`. A Raycast extension can declare `{ modifiers: ["cmd"], key: "1" }`, and `ExtensionScreen.keyEquivalent` maps any single character through its `default` branch. Those shortcuts used to reach `ExtensionCommandScreen.dispatchShortcut`. With this change they never arrive.
> 
> ## 2. The slots are swallowed on every other screen too
> Snippets, quicklinks, emoji, calculator history and the quicklink-argument form all consume ⌘1…⌘0 now, and `activateFavoriteSlotShortcut` does nothing for them. Same in compact mode when `showFavoritesInCompactMode` is off: the old `onKeyPress` returned `.ignored` there and let the key pass, the new path eats it.
> 
> Both this and (1) come from the same shape: AppKit decides to consume before anything knows whether the slot means something. Two ways out, either is fine:
> 
> * Gate the interception on `core.palette.mode` so only launcher and clipboard intercept.
> * Or have the screen report back, and only consume when it actually handled the slot.
> 
> ## 3. Two unrelated changes are riding along
> Both have an inline comment below. The reasoning:
> 
> **`project.pbxproj`** gains `CODE_SIGN_IDENTITY[sdk=macosx*] = "-"` on the Debug config, so `Tinycast Dev.app` becomes ad-hoc signed instead of using `CODE_SIGN_IDENTITY: "Tinycast Self-Signed"` from `project.yml`. Two problems: XcodeGen owns the project, so the next `xcodegen generate` deletes it; and while it stands, every Debug build gets a fresh identity, so the Dev app's Accessibility grant resets each time. Please drop it.
> 
> **`ShellCommandRunner`** goes from `-ilc` to `-il`, `+m`, `-c`. Turning off job control may well be a genuine fix, but it changes how every custom command runs, it is not mentioned in the description, it has no test, and the comment above the line now explains only half of the flags. Please split it into its own PR.
> 
> ## Smaller notes
> * The handler now requires modifiers to be exactly `.command`, so ⌘⇧1 no longer launches a favorite. The old `press.modifiers.contains(.command)` allowed it. Probably what you want, worth one line in the description.
> * `favoriteSlotIndex` is never cleared after use. Harmless with the token pattern, just a stale value that lingers.
> 
> ## What looks good
> `pinnedItem(at:in:filter:)` reuses the cached `search`, and `prefix(while: \.isPinned)` matches the existing invariant that pins lead the results. The row-label counter in `ClipboardList.rows` counts pins the same way, so the labels and the actions cannot disagree. Test coverage for the query and filter cases is the right contract to lock down.
> 
> _Review by reading only: no macOS toolchain on hand, so the build, `run-tests.sh` and `lint.sh` are unverified here._

@mathiasdotdev your statements on these? and you have not removed unrelated stuffs

mathiasdotdev:
> > Thanks for this. The core idea is right: matching the physical number row instead of the typed character is the correct fix, and the AZERTY bug is real. The clipboard half looks sound. Three things need to change before this can land.
> > ## 1. Extension ⌘-number shortcuts stop working
> > `onCommandShortcut` now consumes **every** command-only number-row key and returns `true`:
> > ```swift
> > if let index = FavoriteSlots.index(forKeyCode: event.keyCode) {
> >     self.core.palette.noteFavoriteSlot(index)
> >     return true
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That handler runs inside `PalettePanel.sendEvent`, before `super.sendEvent`, so it wins over every SwiftUI `onKeyPress`. A Raycast extension can declare `{ modifiers: ["cmd"], key: "1" }`, and `ExtensionScreen.keyEquivalent` maps any single character through its `default` branch. Those shortcuts used to reach `ExtensionCommandScreen.dispatchShortcut`. With this change they never arrive.
> > ## 2. The slots are swallowed on every other screen too
> > Snippets, quicklinks, emoji, calculator history and the quicklink-argument form all consume ⌘1…⌘0 now, and `activateFavoriteSlotShortcut` does nothing for them. Same in compact mode when `showFavoritesInCompactMode` is off: the old `onKeyPress` returned `.ignored` there and let the key pass, the new path eats it.
> > Both this and (1) come from the same shape: AppKit decides to consume before anything knows whether the slot means something. Two ways out, either is fine:
> > 
> > * Gate the interception on `core.palette.mode` so only launcher and clipboard intercept.
> > * Or have the screen report back, and only consume when it actually handled the slot.
> > 
> > ## 3. Two unrelated changes are riding along
> > Both have an inline comment below. The reasoning:
> > **`project.pbxproj`** gains `CODE_SIGN_IDENTITY[sdk=macosx*] = "-"` on the Debug config, so `Tinycast Dev.app` becomes ad-hoc signed instead of using `CODE_SIGN_IDENTITY: "Tinycast Self-Signed"` from `project.yml`. Two problems: XcodeGen owns the project, so the next `xcodegen generate` deletes it; and while it stands, every Debug build gets a fresh identity, so the Dev app's Accessibility grant resets each time. Please drop it.
> > **`ShellCommandRunner`** goes from `-ilc` to `-il`, `+m`, `-c`. Turning off job control may well be a genuine fix, but it changes how every custom command runs, it is not mentioned in the description, it has no test, and the comment above the line now explains only half of the flags. Please split it into its own PR.
> > ## Smaller notes
> > 
> > * The handler now requires modifiers to be exactly `.command`, so ⌘⇧1 no longer launches a favorite. The old `press.modifiers.contains(.command)` allowed it. Probably what you want, worth one line in the description.
> > * `favoriteSlotIndex` is never cleared after use. Harmless with the token pattern, just a stale value that lingers.
> > 
> > ## What looks good
> > `pinnedItem(at:in:filter:)` reuses the cached `search`, and `prefix(while: \.isPinned)` matches the existing invariant that pins lead the results. The row-label counter in `ClipboardList.rows` counts pins the same way, so the labels and the actions cannot disagree. Test coverage for the query and filter cases is the right contract to lock down.
> > _Review by reading only: no macOS toolchain on hand, so the build, `run-tests.sh` and `lint.sh` are unverified here._
> 
> @mathiasdotdev your statements on these? and you have not removed unrelated stuffs

I did not push my changes : When I removed
```swift
// zsh reads `.zshrc` only for interactive shells, so `-l` alone sees no aliases.
process.arguments = [loadingShellEnvironment ? "-il" : "-l", "+m", "-c", command]
```
The test was hanging on the first call to:

```swift
ShellCommandRunner.run("tinycast_probe", loadingShellEnvironment: true)
```
When loadingShellEnvironment is true, the command is executed through:
```swift
/bin/zsh -ilc tinycast_probe
```

The -i option starts zsh in interactive mode. However, the Swift Process API does not provide a pseudo-terminal (PTY). As a result, zsh behaves like an interactive shell without a real terminal attached.
In this situation, zsh may try to enable job control, which can cause the shell process to be stopped by the operating system. In ps, the process appears with state T (stopped).
Since the process is stopped rather than exited, waitUntilExit() waits forever, causing the test to hang.
The fix is to disable job control by adding +m:
```Swift
process.arguments = [
loadingShellEnvironment ? "-il" : "-l",
"+m",
"-c",
command
]
```
+m turns off job control, preventing zsh from suspending itself when running interactively without a terminal. The command can then execute normally and exit, allowing waitUntilExit() to return.

mathiasdotdev:
Instead of attaching a PTY to `Process`, I applied a simple fix by using `+m` to disable job control, since it isn't needed when no terminal is attached.

I will add this in another PR.

abue-ammar:
> Instead of attaching a PTY to `Process`, I applied a simple fix by using `+m` to disable job control, since it isn't needed when no terminal is attached.
> 
> I will add this in another PR.

do you want me to close this pr? because i won't merge if reviews are ignored, and think those issues might be related to ur zsh (may be)

mathiasdotdev:
> > Instead of attaching a PTY to `Process`, I applied a simple fix by using `+m` to disable job control, since it isn't needed when no terminal is attached.
> > I will add this in another PR.
> 
> do you want me to close this pr? because i won't merge if reviews are ignored, and think those issues might be related to ur zsh (may be)

I'm not ignoring your reviews though... I'm still testing every features in local like you suggested ...

mathiasdotdev:
@abue-ammar I will push my changes soon with every fix you suggested I want to make sure it's tested and functionnal with both keyboard layouts

abue-ammar:
@mathiasdotdev any updates on this? i might close this as this pr is being very outdated and will be hard to merge it in the future

mathiasdotdev:
ok close this PR i'll make a new one with all the new changes because with recent changes it's difficult to merge : i'll start fresh : I made all those tests and it worked with both AZERTY and QWERTY.

## Manual Keyboard Validation

Run every keyboard test with both layouts:

- QWERTY: `Cmd+1` through `Cmd+9`, then `Cmd+0`
- AZERTY: the same physical number-row keys, including `Shift` where required

The physical key position must determine the slot, not the character produced by the layout.

### 1. Launcher Favorites

- [x] Create 10 launcher favorites.
- [x] QWERTY: physical keys 1 through 9 launch favorites 1 through 9.
- [x] QWERTY: physical key 0 launches favorite 10.
- [x] AZERTY: the same physical keys launch the same favorites.
- [x] The displayed labels are `1` through `9`, then `0`.
- [x] With fewer than 10 favorites, unused slots do nothing.
- [x] With 10 favorites, `Add to Favorites` is disabled.
- [x] Remove a favorite, then confirm a new favorite can be added.
- [x] Reorder favorites and confirm the slots follow the new order.

### 2. Launcher Compact Mode

- [x] Enable `Show favorites in compact mode`.
- [x] Open the launcher in compact mode.
- [x] QWERTY: test all 10 physical number-row keys.
- [x] AZERTY: test the same 10 physical keys.
- [x] Confirm visible favorites launch correctly.
- [x] Confirm slots 6 through 0 launch favorites not visible in the compact bar.
- [x] Disable `Show favorites in compact mode` while keeping favorites configured.
- [x] Repeat all 10 physical keys on QWERTY and AZERTY.
- [x] Confirm favorites still launch even though they are hidden.
- [x] Expand the palette and confirm the slot mapping is unchanged.

### 3. Clipboard Pinned Slots

- [x] Add and pin at least 3 clipboard entries.
- [x] QWERTY: each physical slot pastes the matching pinned entry.
- [x] AZERTY: the same physical keys paste the same entries.
- [x] Confirm slot order matches visible pin order.
- [x] Search for a subset of pinned entries and retest the slots.
- [x] Test text, link, image, and mixed clipboard filters.
- [x] Test an unavailable slot and confirm nothing unrelated is pasted.
- [x] Confirm pasting does not reorder pinned entries.
- [x] Hold `Cmd` and confirm the slot labels appear; release it and confirm they disappear.

### 4. Other Palette Screens

For each screen below, test all 10 physical number-row keys with QWERTY and AZERTY:

- [x] Snippets
- [x] Quicklinks
- [x] Quicklink argument form
- [x] Emoji picker
- [x] Calculator history
- [x] File search
- [x] Uninstall screen

### 5. Modifier and Existing Shortcut Behavior

- [x] QWERTY: test `Cmd+Shift+1`.
- [x] AZERTY: test the equivalent physical combination.
- [x] Confirm `Shift` does not prevent a favorite slot from working.
- [x] Test `Cmd+Option+1` and `Cmd+Control+1`; neither launches a favorite.
- [x] Hold a slot key and confirm key repeat does not cause unwanted repeated launches.
- [x] Test `Cmd+,` and confirm Settings opens.
- [x] Test `Cmd+.` and confirm the selected item is pinned where applicable.
- [x] Test `Cmd+W` and confirm the palette closes.
- [x] Test `Escape` and confirm existing escape behavior is unchanged.
- [x] Switch between QWERTY and AZERTY without restarting Tinycast.
- [x] Repeat a launcher slot and clipboard slot after switching layouts.

### 7. Final Regression and Shell Smoke Tests

- [x] Close and reopen Tinycast, then retest one launcher slot in both layouts.
- [x] Confirm favorite persistence after relaunch.
- [x] Confirm clipboard pin persistence after relaunch.


I found 2 bugs during those tests (idk if it's intended) :
1. when you open a note, and use the clipboard history to paste something in a window, it's always pasted into the note (the Note always has the focus)
2. you can have more than 10 favorites in the launcher (not reachable with keyboard shortcuts (I made an option to disable favorites when the maximum number is reached (shortcuts and click is disabled (you still can remove an element from the favorites if you want)

PR #319, opened 2026-08-23
Add AI chat with BYOK providers and an optional ChatGPT subscription

notsoshu:
## Related issue

Discussed directly — you saw the chat screenshots and asked for the feature as one PR, with the menu click-away fix separately. That fix landed as #318, and its commit has dropped out of this branch on rebase.

## What changed

An AI chat screen in the palette, with bring-your-own-key HTTP providers and an optional ChatGPT subscription route. Four feature commits, each reviewable on its own:

1. **Wire an AI chat screen into the palette and Settings** — the upstream touch points: two `PaletteMode` cases (chat, chat history), an `AI Chat` launcher command, an `AI` Settings tab, three `AppSettingsKey`s excluded from backups, and the state `AppCore` owns. `HeaderMenuButton` lifts the clipboard filter's control so the chat header's model switcher shares it (`ClipboardFilterButton` now uses it, visually unchanged); `PopoverMenuIcon` gains a bundled-asset case for vendor marks. `PaletteMode.title` was unused and is removed rather than extended.
2. **Add AI chat** — everything under `Features/AI/`:
   - **Providers.** OpenAI-compatible Chat Completions (OpenAI, Gemini, OpenRouter, custom) and Anthropic Messages, streamed over a private `.ephemeral`, `urlCache = nil` session per request, decoded off the main actor. Keys live in the login Keychain keyed by connection UUID and never enter defaults, logs, errors or backups. Remote endpoints must be HTTPS, plain HTTP is accepted only on loopback (Ollama, LM Studio), and any other scheme is rejected outright.
   - **ChatGPT subscription** via the user's own Codex CLI run as `codex app-server` in a private `CODEX_HOME` under Application Support, tools disabled, approvals never, read-only no-network sandbox, every server approval request declined. Started only for a stored sign-in or an explicit Connect, and stopped after ten idle minutes — counted from a failed or still-waiting Connect too, so a sign-in abandoned in the browser cannot leave it resident. Stopped on quit. Tinycast never installs Codex, never reads `~/.codex`, and links to the install docs when the binary is missing.
   - **Chat.** The search field is the composer; Return sends, or stops a streaming reply. Replies render through a block-level Markdown parser (headings, lists, task boxes, quotes, fenced code with copy, tables, rules; inline spans via `AttributedString`). Thinking and web-search state show in the bubble without entering the transcript. ⌘V stages an image for models that take one, decoded off the main actor and capped at six images or 10 MB a turn, with a HUD when one will not fit. A reply keeps streaming while the palette is hidden and saves when it finishes.
   - **History.** Conversations persist in `ai-chats.sqlite3` (system SQLite, already linked); a second mode lists them bucketed by day like Clipboard, with preview, open, delete and delete-all.
   - **Settings.** Default model, web search (off by default), ChatGPT sign-in with plan and rate-limit windows, and API connections with search-as-you-type model discovery from the provider's own list, capped at twelve results, with manual entry for gateways that expose none.
3. **Harnesses** — `ai-provider-test` (105 assertions), `ai-chat-test` (69) and `codex-turn-test` (4), including four captured OpenRouter streams in `Tests/ai-fixtures/` decoded whole and in 7-byte slices.
4. **Docs** — `docs/features/ai.md`, opening with the invariants.

Four follow-ups sit on top, each self-contained:

5. **Keep a stored API key from following a retargeted connection** — a key is issued for one endpoint. Changing a connection's provider or base URL now asks for a key for the new one rather than carrying the saved secret to a host it was never meant to reach, and a retarget onto loopback drops it. `AIEndpointPolicy.sameDestination` is the single rule; both Save and model discovery honour it.
6. **Disown a staged image the chat has moved past** — the ⌘V decode is detached, so anything that consumed or cleared the staged images while it ran — a send, a new chat, Remove Attachments, or leaving the conversation for another through history — no longer lets it land on whatever message came next. The generation lives on `AIChatState` beside the images it guards, so a route that drops them cannot forget to move it.
7. **Carry the third-party notice inside the app bundle** — the MIT mark's licence asks the notice to travel with the work rather than sit only in the repo, so `NOTICE.md` ships as a bundle resource.
8. **Say what the context bound actually promises** — the invariant read as though every message were bounded. What a turn resends stops growing with the chat; the newest message is still sent whole, because truncating what someone just typed is worse than the provider's own error.

Four more came out of testing the built app, and one of them adds a visible control:

9. **Keep a wide ordered-list marker from wrapping** — the marker sat in a fixed 20pt column, so any number past one digit wrapped inside it: `10.` rendered as `10` above a lone `.`, and `100.` as `10` above `0.`. The column is a minimum width now, the marker never wraps, and padding to the list's widest number keeps it aligned.
10. **Resolve a first default model without a trip to Settings** — choosing a default when none was stored happened only in the AI settings pane, so a signed-in ChatGPT subscription still sent the reader there to pick what the app already knew about; merely opening the pane fixed it. Chat resolves the same way, including when the subscription signs in after the screen is already up.
11. **Clear the palette footer beneath a transcript's last message** — the last message's copy button and timestamp sat close enough to the bottom edge to read as one row with the palette's own footer controls.
12. **Let a reader hold their place in a streaming reply** — following the tail was dropped only during a `ScrollPhase` change, and a plain wheel reports none, so a reply dragged those readers back down on every flush. The offset itself is the signal now, which every input device gives. **This one adds a control:** a fast reply grows the transcript quicker than anyone can scroll toward it, so arriving at the end cannot be the only way to resume — a **Jump to Latest** pill appears whenever the reader is away from the tail. It is the one new piece of visible UI since your review.

Non-obvious in the diff: `DateBucket` is reused from `ClipboardView.swift` for history's section headers; the brand marks are monochrome template SVGs — twelve from Simple Icons plus Z.ai from `@lobehub/icons`, with provenance, the MIT notice and the trademark position recorded in `NOTICE.md`, which now ships inside the app bundle; `Theme` gains the chat's tokens rather than the feature hard-coding sizes.

No AI network request and no Codex process starts until the feature is configured or used: the default-model picker is empty, no provider session exists, nothing is spawned. One thing does run at launch — the chat history store opens its SQLite file, and so creates it, on the same deferred `Task` the clipboard store already uses, off the launch path.

## Since your review

All fourteen comments are answered here rather than in threads, grouped by what actually changed — with the three declined ones and their reasons at the end:

- **Codex turn lifecycle** (#954, #956, #957, #973) — Stop now interrupts the turn even when it lands before anything has named it: the thread is armed, and whichever of `turn/started` or the `turn/start` response names the turn first spends a single `turn/interrupt`. The client holds the `Process` and terminates through it, and the JSON-RPC read buffer is capped at 8 MB — failing loud with the error surfaced rather than dropping silently, since an unterminated multi-MB line means the helper is not speaking JSON-RPC. `codex-turn-test` is new and pins this against a stub app-server stalled in the race window — it fails on the code as it stood before the fix.
- **Request size** (#984) — `ChatSession.boundedContext` caps what a turn resends, and `AIAttachmentBudget` caps the turn itself: six images or 10 MB, refused in the composer with a HUD rather than silently dropped at send time.
- **Persistence** (#981, #979) — saves rewrite only from the last stored message, so image blobs are written once; the history database runs WAL like the other stores; image decode, rescale and the file read moved off the main actor onto a detached task.
- **Discovery and keys** (#948, #977, #982) — one shared `URLSession` for discovery on the `CurrencyRateStore` shape, `catalogURL` optional with discovery throwing rather than force-unwrapping, and `APIKeyStore.hasKey` checking presence without materializing the secret.
- **Streaming cadence** (#962) — deltas buffer in `AIChatState` and flush on a 40 ms cadence, immediately on finish, failure or cancel.
- **Declined, with the reason** (#968, #972, #975) — the transcript's `onChange(of: messages)` compares no image bytes: `Data ==` short-circuits on shared storage and every unmutated message keeps its COW storage across snapshots, while keying on text lengths would stop following search-row inserts and the clean-finish transition. The per-chunk tasks cannot reorder: the readability handler is serial per handle, and explicitly `@MainActor` task closures are enqueued on the actor in creation order (SE-0431). The per-byte SSE loop stays: `chunks(ofCount:)` is not in Foundation, `AsyncBytes.lines` drops the blank lines SSE uses as frame separators, and the loop reads from `AsyncBytes`' internal buffer, so the measured cost is negligible.
- **Found while re-reviewing, not raised by you** — `ftp://localhost` passed endpoint validation; a failed or abandoned ChatGPT Connect never armed idle shutdown; the brand marks shipped with no third-party notice. Commits 5–8 above are the rest of that pass.
- **Found by using the built app** — commits 9–12: list markers wrapping past nine, a default model that only resolved inside Settings, the transcript's last row crowding the footer, and a streaming reply that would not let a wheel-mouse reader hold their place.

## Memory footprint

Physical footprint (`footprint`, what Activity Monitor shows), unsigned Release builds, palette hidden by activating another app, `main` measured the same way as the control.

| Step                    | Memory | `main` |
| ----------------------- | ------ | ------ |
| Idle after launch       | 26 MB  | 26 MB  |
| Palette open            | 43 MB  | 43–44 MB |
| Feature in use (peak)   | 61 MB — one prompt, ~1.8k-char reply with code block, table and list, streamed via ChatGPT | — |
| Palette closed, settled | 60 MB (+40 s; transcript still mounted until pop-to-root) · 44 MB after a palette-only session | 45 MB |

The optional Codex helper is a separate process: 16 MB (node) + 26 MB (native) while resident, gone after ten idle minutes and on quit.

Release binary 7,240,520 B, from 6,315,640 B on `main` at the commit this branch is rebased onto — +924,880 B, +14.64%. Both figures are unsigned Release builds of this branch and of `8cd1b16`, measured on the same machine and toolchain; the `main` number reproduces byte-for-byte between runs. Worth naming rather than leaving you to find it: `docs/standards.md` and `docs/testing.md` both record "under 4 MB", and `docs/testing.md` adds "under 2% growth for an ordinary change" against a logged 3,655,736 B. `main` is already 51% past the 4 MB line on its own, so that budget reads as stale rather than something this PR crosses — but +14.64% plainly crosses the growth rule, and whether a whole feature counts as "an ordinary change" is yours to call. If it does not, the ChatGPT route is the largest single piece and is cleanly deletable; otherwise the recorded budget wants updating in the same commit.

On bringing it down, since deleting the feature should not be the only option on the table. The feature's code concentrates in a few places rather than spreading evenly: `AISettingsView` and the ChatGPT route are the two largest pieces, and `AIBrand` is outsized for its 62 lines — a static table of string-array literals costs far more than it reads like. The `AIBrand` one looks cheap and self-contained; the settings view would mean splitting a large view body, which is more churn in a file you have already read. Say which of those is worth doing and I will measure it rather than guess.

One number that is not mine to spend, but is worth you having: Release sets no `SWIFT_OPTIMIZATION_LEVEL`, so it defaults to `-O`. Built with `-Osize`, this branch links at 6,554,424 B — 686,096 B smaller, −9.5%, roughly three quarters of everything this PR adds. It would shrink `main` by about as much, so it does not make the feature free; it means the app-wide budget has a far bigger lever in it than any one feature does. That is a whole-app codegen change with a possible runtime cost, so I would rather raise it as its own PR with a perf check than fold it into this one.

Re-measured on the current branch after the rebase — idle 26 MB, palette open 44–47 MB, peak 56 MB, settled 53 MB, helper 40 MB across its two processes. The table above therefore stands, and is conservative on peak and settled. Its `main` control column was measured before the rebase and is the one column not re-taken.

Leak-tested: Instruments run to follow in a comment, with a footprint reading taken after the transcript actually unmounts. The settled row is with it still mounted, so it is not yet the return to baseline the bar asks for. The binary figures above are freshly measured against the current base.

## Drawbacks

- It is the largest feature in the app. The ChatGPT route alone is ~1,200 lines (JSON-RPC client, turn runner, lifecycle, settings UI) and is the only part with an external binary. It is cleanly deletable if you would rather land BYOK first.
- The Markdown block parser is hand-written (341 lines, tested) because `AttributedString` only does inline spans and SwiftUI cannot lay out its block intents.
- `HeaderMenuButton` and `PopoverMenuIcon.asset` touch three existing menu renderers; the `ExtensionActionsPanel` arm exists because the switch is exhaustive.
- Web search and image attachments each add a little plumbing across the request, transcript and history; both are off or inert until a model offers them.
- Brand marks are third-party trademarks used nominatively to label a vendor. `NOTICE.md` records the position and ships inside the app bundle; say if you would rather ship generic glyphs instead.

## Tests & validation

`./Scripts/run-tests.sh` — all 42 harnesses pass, including the three new ones. Debug build with zero new warnings. `./Scripts/format.sh --check` clean. `xcodegen generate` reproduces the committed project file exactly. Lint runs in CI; not run locally. `grep -rln 'import AppKit\|import SwiftUI\|import Cocoa' Tinycast/Features/*/Model/` returns nothing.

By hand: list to follow in a comment.


abue-ammar (commented):
## Review

I built the branch, ran the harnesses and checked the invariants. Everything you claimed holds:

| Check | Result |
| --- | --- |
| Debug build | Passes, no Swift warnings |
| `./Scripts/run-tests.sh` | Passes — 40 harnesses, `ai-provider-test` 99, `ai-chat-test` 31 |
| `./Scripts/lint.sh` | Clean |
| No AppKit/SwiftUI under `Features/*/Model/` | Clean |

**What is good.** Keys stay in the Keychain, out of defaults, out of backups and out of error text. `AIEndpointPolicy` is the single place that decides HTTPS, and loopback is the only exception. Each request opens its own ephemeral session and closes it after, as `CurrencyRateStore` does. The Codex server starts with tools off, approvals never, and a read-only sandbox with no network, and it declines every approval request. `docs/features/ai.md` opens with the invariants.

**Four things to fix before merge.** They are in the inline comments:

1. `AISettingsView` builds a new `URLSession` on each view build, and nothing closes it.
2. Stop does not interrupt a ChatGPT turn if it is pressed during the `turn/start` round trip. The model keeps writing and it spends the user's plan quota.
3. `kill(pid, SIGTERM)` runs one second after the `Process` is released, so it can reach a different process.
4. The full reply goes through the Markdown parser again on every token, on the main actor. The work grows with the square of the reply length.

The rest of the inline notes are smaller.

**On size.** 7,080 lines and 84 files is a lot for one review. Your own Drawbacks section offers the split, and I would take it: land the BYOK HTTP route first, then the ChatGPT/Codex route on its own. Problems 2 and 3 both live in the Codex part, and that part would get a sharper read alone.

Tinycast/Features/AI/Settings/AISettingsView.swift
```diff
@@ -0,0 +1,809 @@
...
+    let onCancel: () -> Void
+
+    @State private var connection: AIConnection
+    @State private var key = ""
+    @State private var modelQuery = ""
+    @State private var discovery: ModelDiscoveryState = .waitingForKey
+    @State private var discoveryRevision = 0
+    @State private var error: String?
+
+    private let modelDiscovery = AIModelDiscoveryService()
```
abue-ammar, line 490: **A new `URLSession` on each view build, and nothing closes it.**

SwiftUI rebuilds this struct each time the parent view draws. Each build runs `AIModelDiscoveryService()`, and that initialiser makes a `URLSession`. No code calls `invalidateAndCancel()` or `finishTasksAndInvalidate()`, so each session stays alive until the app stops.

`CurrencyRateStore` — the reference the AGENTS.md points at — keeps one session for the life of the app:

```swift
// AIModelDiscoveryService.swift
final class AIModelDiscoveryService: Sendable {
    static let shared = AIModelDiscoveryService()

    private let session: URLSession
    private init() { ... }
}
```

Then this line becomes:

```suggestion
    private let modelDiscovery = AIModelDiscoveryService.shared
```

Tinycast/Features/AI/Service/CodexTurnRunner.swift
```diff
@@ -0,0 +1,240 @@
...
+            let turnResponse = try await client.request(
+                method: "turn/start",
+                params: [
+                    "threadId": threadID,
+                    "model": model,
+                    "approvalPolicy": "never",
+                    "sandboxPolicy": ["type": "readOnly", "networkAccess": false],
+                    "input": turnInput(for: request.messages[promptIndex])
+                ])
+            activeTurnID = turnResponse["turn"]?.objectValue?["id"]?.stringValue
```
abue-ammar: **Stop does not interrupt the turn, if the user presses it here.**

`activeTurnID` is set only after `turn/start` answers. That round trip is not fast. If the user presses Stop inside it, `onTermination` runs `interruptActiveTurn()`, which finds `activeTurnID == nil` and only clears the state — it sends no `turn/interrupt`.

Then this line writes the turn ID back into a state that is already clear.

Two results:
- The model keeps writing on the server, and it spends the user's plan quota.
- `isActive` reads `false`, so the idle timer can stop the client under a live turn.

One way out is to remember the request and act on it when the ID arrives:

```swift
private var interruptRequested = false

// in startTurn, after turn/start answers:
activeTurnID = turnResponse["turn"]?.objectValue?["id"]?.stringValue
if interruptRequested { interruptActiveTurn() }

// in interruptActiveTurn:
guard let threadID = activeThreadID, let turnID = activeTurnID else {
    interruptRequested = activeThreadID != nil   // the thread exists, the turn ID has not arrived
    if !interruptRequested { clearActiveTurn() }
    return
}
```

The same gap makes the "A newer request replaced this response" path leave the old turn running, although `AIChatState.send` blocks a second send while one streams, so that path is hard to reach today.

Tinycast/Features/AI/Service/CodexAppServerClient.swift
```diff
@@ -0,0 +1,283 @@
...
+
+    /// Closing stdin is the clean exit — the server leaves on EOF — and SIGTERM is the backstop.
+    func stop() {
+        guard let process else { return }
+        process.terminationHandler = nil
+        let pid = process.processIdentifier
+        cleanup(error: ClientError.processExited("Codex stopped."))
+        Task.detached {
+            try? await Task.sleep(for: .seconds(1))
+            kill(pid, SIGTERM)
```
abue-ammar: **This can send `SIGTERM` to a different process.**

`cleanup(error:)` on the line above releases the `Process`. Foundation then reaps the child, and macOS is free to give that PID to something else. One second later this `kill` reaches whatever holds the PID now — any process of the same user.

Hold the object and let it do the work:

```swift
func stop() {
    guard let process else { return }
    process.terminationHandler = nil
    cleanup(error: ClientError.processExited("Codex stopped."))
    // Closing stdin is the clean exit; terminate() is the backstop, and it cannot hit a reused PID.
    Task { @MainActor in
        try? await Task.sleep(for: .seconds(1))
        if process.isRunning { process.terminate() }
    }
}
```

Tinycast/Features/AI/Service/CodexExecutableLocator.swift
```diff
@@ -0,0 +1,74 @@
...
+            }
+            process.standardInput = FileHandle.nullDevice
+            process.standardError = FileHandle.nullDevice
+            let stdout = Pipe()
+            process.standardOutput = stdout
+            do { try process.run() } catch { return nil }
+            let pid = process.processIdentifier
+            let watchdog = Task {
+                try await Task.sleep(for: .seconds(5))
+                kill(pid, SIGTERM)
```
abue-ammar: Same PID-reuse shape as `CodexAppServerClient.stop()`, in a smaller window: `waitUntilExit()` returns, the child is reaped, and `watchdog.cancel()` runs one statement later. If the sleep finishes inside that gap, this `kill` reaches whatever holds the PID now.

`process.terminate()` closes it, because the object knows whether the child is still its own:

```swift
let watchdog = Task {
    try await Task.sleep(for: .seconds(5))
    if process.isRunning { process.terminate() }
}
```

Tinycast/Features/AI/UI/ChatTranscriptView.swift
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// Only a reply is markdown — what the user typed is shown back exactly as they typed it.
+    @ViewBuilder private var rendered: some View {
+        if message.role == .assistant {
+            VStack(alignment: .leading, spacing: Theme.Spacing.lg) {
+                ForEach(Array(message.segments.enumerated()), id: \.offset) { _, segment in
+                    switch segment {
+                    case .text(let text):
+                        MarkdownView(blocks: MarkdownBlock.parse(text))
```
abue-ammar, line 195: **The whole reply is parsed again on every token.**

`MarkdownBlock.parse(text)` runs inside `body`. Each delta changes `message.text`, SwiftUI re-evaluates, and the parser walks the full reply from the start again — `components(separatedBy:)` allocates the line array each time. Every paragraph then goes through `MarkdownInline.attributed`, which calls `AttributedString(markdown:)` again. All of it on the main actor.

The cost grows with the square of the reply length. Your own comment on `ChatImageThumbnail` says "a streaming transcript re-renders every delta" — that is exactly the pressure this puts on the main actor.

Two fixes, either is enough:

```swift
// 1. Parse when the text changes, not when the view draws.
struct ChatMessage {
    var text: String { didSet { blocks = MarkdownBlock.parse(text) } }
    private(set) var blocks: [MarkdownBlock] = []
}
```

```swift
// 2. Or group the deltas in AIChatState, so a fast stream draws at a fixed rate
//    instead of once per token.
```

A measurement would settle which: a reply of a few thousand characters, with a code block and a table, is the case to watch.

Tinycast/Features/AI/UI/ChatTranscriptView.swift
```diff
@@ -0,0 +1,205 @@
...
+            // A reopened chat starts at its latest message, not its first.
+            .defaultScrollAnchor(.bottom)
+            .onScrollGeometryChange(for: Bool.self) { geometry in
+                geometry.contentOffset.y + geometry.containerSize.height
+                    >= geometry.contentSize.height - Theme.Spacing.xl
+            } action: { _, atBottom in
+                followsTail = atBottom
+            }
+            .onChange(of: messages.count) { follow(proxy, always: true) }
+            .onChange(of: messages) { follow(proxy, always: false) }
```
abue-ammar, line 66: `messages` is `[ChatMessage]`, and `ChatMessage` holds `[AIImage]`, which holds `Data`. So this `onChange` compares every image byte on every delta of a streaming reply.

The line above it already follows on `messages.count`. This one only needs to know that the last reply grew:

```swift
.onChange(of: messages.last?.text.count) { follow(proxy, always: false) }
```

Tinycast/Features/AI/Service/CodexAppServerClient.swift
```diff
@@ -0,0 +1,283 @@
...
+                "PATH": (commandPaths + [inheritedPath]).joined(separator: ":")
+            ]
+        ) { _, value in value }
+        process.standardInput = stdin
+        process.standardOutput = stdout
+        process.standardError = stderr
+        stdout.fileHandleForReading.readabilityHandler = { [weak self] handle in
+            let data = handle.availableData
+            guard !data.isEmpty else { return }
+            Task { @MainActor in self?.consumeOutput(data) }
```
abue-ammar, line 111: Each chunk starts its own unstructured `Task`. Swift does not promise that such tasks run in the order you create them, so two chunks can reach `consumeOutput` swapped. If they do, the JSON-RPC line framing breaks, and a reply is dropped or mis-parsed.

In practice the main actor usually keeps FIFO order for equal priority, so this will be rare — which makes it the kind of bug that only shows up under load.

One reader over an `AsyncStream` keeps the order by construction:

```swift
let (stream, continuation) = AsyncStream<Data>.makeStream()
stdout.fileHandleForReading.readabilityHandler = { handle in
    let data = handle.availableData
    if !data.isEmpty { continuation.yield(data) }
}
readerTask = Task { @MainActor [weak self] in
    for await data in stream { self?.consumeOutput(data) }
}
```

Tinycast/Features/AI/Service/CodexAppServerClient.swift
```diff
@@ -0,0 +1,283 @@
...
+        }
+    }
+
+    private func send(_ data: Data) throws {
+        guard let input else { throw ClientError.processExited("Codex is not running.") }
+        try input.write(contentsOf: data)
+    }
+
+    private func consumeOutput(_ data: Data) {
+        outputBuffer.append(data)
```
abue-ammar, line 200: `stderrBuffer` is capped at 8 KB in `consumeStderr`, but `outputBuffer` has no limit. A server that writes a long line with no `\n` — a broken build, a crash dump on stdout — grows this without end.

The cap belongs here too:

```suggestion
        outputBuffer.append(data)
        // A line this long is not JSON-RPC; drop it rather than grow without a bound.
        if outputBuffer.count > 8 * 1_024 * 1_024 { outputBuffer.removeAll(keepingCapacity: false) }
```

Tinycast/Features/AI/Service/HTTPAIProvider.swift
```diff
@@ -0,0 +1,194 @@
...
+                            "The provider returned an invalid HTTP response.")
+                    }
+                    guard response.statusCode == 200 else {
+                        throw AIProviderError.responseFailed(Self.statusMessage(response))
+                    }
+                    var decoder = AIStreamDecoder(shape: configuration.shape)
+                    var chunk = Data()
+                    chunk.reserveCapacity(2_048)
+                    // Per line, not per 2 KB: a short reply must show before the stream closes.
+                    for try await byte in bytes {
```
abue-ammar, line 32: One `await` per byte. `URLSession.AsyncBytes` buffers inside, so most steps resume without a real suspension, but this is still one async call for each byte of the reply.

The comment above says the split must be per line, not per 2 KB — that stays true if you read the chunk and cut it at each newline:

```swift
for try await chunk in bytes.chunks(ofCount: 4_096) { ... }
```

or keep the loop and let the decoder see whole chunks, since `SSEParser` already buffers a part frame:

```swift
for try await line in bytes.lines {
    try Task.checkCancellation()
    for event in try decoder.feed(Data((line + "\n\n").utf8)) { continuation.yield(event) }
    if decoder.isTerminal { break }
}
```

Not urgent — the reply size makes it small — but it is repeated work the file does not need.

Tinycast/Features/AI/Settings/AISettingsView.swift
```diff
@@ -0,0 +1,809 @@
...
+        } catch {
+            keyError = true
+        }
+    }
+
+    private func loadKeyStatuses() {
+        var statuses: [UUID: Bool] = [:]
+        do {
+            for connection in settings.connections {
+                statuses[connection.id] = try keyStore.key(for: connection.id)?.isEmpty == false
```
abue-ammar: This reads every API key out of the Keychain only to answer "is one stored?". The secret then sits in memory for as long as the string lives, and this runs on each `onAppear` and after each save or removal.

The Keychain answers the same question without handing over the data. `APIKeyStore` already builds the query:

```swift
// APIKeyStore.swift
func hasKey(for connectionID: UUID) throws -> Bool {
    let status = SecItemCopyMatching(
        query(for: connectionID, returningData: false) as CFDictionary, nil)
    if status == errSecItemNotFound { return false }
    guard status == errSecSuccess else { throw StoreError.keychain(status) }
    return true
}
```

Tinycast/Features/AI/Service/ChatHistoryStore.swift
```diff
@@ -0,0 +1,354 @@
...
+            try FileManager.default.createDirectory(
+                at: databaseURL.deletingLastPathComponent(), withIntermediateDirectories: true)
+        } catch {
+            isAvailable = false
+            return false
+        }
+        guard
+            sqlite3_open_v2(
+                databaseURL.path, &database,
+                SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE | SQLITE_OPEN_FULLMUTEX, nil) == SQLITE_OK,
```
abue-ammar, line 194: The store is `@MainActor`, so one thread ever touches this handle. `SQLITE_OPEN_FULLMUTEX` puts a lock around every call for no gain.

```suggestion
                SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE | SQLITE_OPEN_NOMUTEX, nil) == SQLITE_OK,
```

Tinycast/Features/AI/Service/ChatHistoryStore.swift
```diff
@@ -0,0 +1,354 @@
...
+        bind(summary.id.uuidString, to: statement, at: 1)
+        bind(summary.title, to: statement, at: 2)
+        bind(summary.preview, to: statement, at: 3)
+        sqlite3_bind_double(statement, 4, summary.createdAt.timeIntervalSince1970)
+        sqlite3_bind_double(statement, 5, summary.updatedAt.timeIntervalSince1970)
+        sqlite3_bind_int64(statement, 6, Int64(summary.messageCount))
+        return sqlite3_step(statement) == SQLITE_DONE
+    }
+
+    private func replaceMessages(in session: ChatSession, database: OpaquePointer) -> Bool {
```
abue-ammar: Each save deletes every message of the conversation and writes them all again, image blobs included, on the main actor. `AIChatState` saves at the start of a turn and at the end of one, so a long chat with pictures rewrites all of its blobs twice per turn.

The messages already carry a stable `id` and a `position`. An upsert on the rows that changed, plus a delete of the rows past the new end, would touch two rows in the usual case:

```sql
INSERT INTO messages(id, conversation_id, position, role, text, state, sent_at)
VALUES(?, ?, ?, ?, ?, ?, ?)
ON CONFLICT(id) DO UPDATE SET
  position = excluded.position, text = excluded.text, state = excluded.state;
```

The images then need writing only for a message the database has not seen.

Tinycast/Features/AI/Model/AIModelDiscovery.swift
```diff
@@ -0,0 +1,203 @@
...
+            var queryItems = components.queryItems ?? []
+            queryItems.removeAll { $0.name == "limit" }
+            queryItems.append(URLQueryItem(name: "limit", value: "1000"))
+            components.queryItems = queryItems
+        } else if provider == .openRouter {
+            components.path = path + "/models/user"
+        } else {
+            components.path = path + "/models"
+        }
+        return components.url!
```
abue-ammar: Two force unwraps in this function — this one and `URLComponents(url:resolvingAgainstBaseURL:false)!` on line 117.

Both are safe today, because `AIEndpointPolicy.validate` runs first and the rebuilt path always starts with `/`. But this is a `Model/` file, the one place the project holds to Foundation only and to no surprises, and a later edit to the path maths would turn a bad base URL into a crash.

```swift
guard var components = URLComponents(url: baseURL, resolvingAgainstBaseURL: false) else {
    return baseURL
}
...
return components.url ?? baseURL
```

Tinycast/Features/AI/Model/ChatSession.swift
```diff
@@ -0,0 +1,69 @@
...
+        return Self.summary(text, limit: 120)
+    }
+
+    var summary: ChatConversation {
+        ChatConversation(
+            id: id, title: title, preview: preview, createdAt: createdAt,
+            updatedAt: updatedAt, messageCount: messages.count)
+    }
+
+    var requestMessages: [AIMessage] {
```
abue-ammar, line 37: Every turn sends the whole conversation again, images and all, with no cap. A long chat gets slower and costlier with each message, and a chat with several pasted screenshots will reach a provider's request limit and fail with a plain HTTP error that says nothing useful.

A cap on what goes out — the last N messages, or a byte budget with the images of older turns dropped first — would keep a long chat usable. Worth a line in `docs/features/ai.md` under the invariants either way, so the next reader knows the choice was made on purpose.

PR #335, opened 2026-08-24
Add vim-style context menu navigation

### LINKED ISSUE #334: [Feature]: Easy navigation in context menu
### What should it do?

On keyboards without arrow keys or when using the home row, navigating between context menu (cmd+k) items is difficult. Not having text filtering is not a big issue (as there's only a few items anyway) but using j/k for navigation can come in handy.

Image

It simplifies menu navigation and increases efficiency by needing less keystrokes/hand movements.

kublaios:
## Related issue

Closes #334

## What changed

Adds Vim-style menu navigation to Tinycast’s palette menus:

- Press `J` to move the highlighted menu item down.
- Press `K` to move the highlighted menu item up.
- Key repeat is supported.
- The bindings apply only while a palette menu is open.
- Command, Control, and Option-modified J/K remain available for shortcuts.

The menu-navigation key set was extracted into `MenuNavigation` so AppKit’s input freeze and SwiftUI’s key handlers share the same definition. Outside an open menu, J and K remain normal search input.

## Memory footprint

TODO

## Demo

TODO

## Drawbacks

n/a

## Tests & validation

- Added `Tests/palette-menu-nav-test.swift`.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -96,6 +96,7 @@ run hover-arming-test      Tinycast/Palette/HoverArming.swift \
                            Tinycast/Features/Clipboard/Model/ClipboardStore.swift \
                            Tinycast/Features/Clipboard/Model/ClipboardFilter.swift \
                            Tinycast/Features/Quicklinks/Model/Quicklink.swift
+run palette-menu-nav-test  Tinycast/Palette/MenuNavigation.swift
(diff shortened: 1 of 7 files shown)


abue-ammar:
@kublaios hey sorry but i do not want  vim key bindings is tinycast. if this is really important to you then you can have it on your fork.

PR #340, opened 2026-08-24
feat: add shared shortcuts for launcher favorites and pinned clipboard items

### LINKED ISSUE #291: [Feature]: Favorites - allow shortcuts with other keyboard layouts

mathiasdotdev:
## Related issue

Close #291 #292 

Related to #315 (previous PR)

## What changed

This PR adds a shared `⌘`-number shortcut system for both launcher favorites and pinned clipboard entries.

### 1. Physical number-row shortcut mapping

`FavoriteSlots` now maps physical macOS number-row key codes to ten slots:

- `⌘1` through `⌘9`
- `⌘0` as the tenth slot

The mapping uses hardware key codes rather than the character produced by the keyboard layout. This makes the shortcuts work consistently on layouts such as QWERTY and AZERTY.

The existing digit-based mapping remains responsible for displaying slot labels, while the new key-code mapping resolves actual keyboard events.

### 2. AppKit handles the shortcuts

The shortcut detection was moved from SwiftUI's `onKeyPress` handling into `PaletteWindowController`.

When a non-repeated command-only number-row event is received:

1. The key code is converted into a favorite slot index.
2. The event is consumed by the palette window.
3. The resolved index is published through `PaletteState`.

This avoids keyboard-layout-dependent behavior and ensures the shortcuts are intercepted before AppKit or the field editor handles them.

### 3. `PaletteState` forwards shortcut events

`PaletteState` now contains:

- `favoriteSlotToken`, a change token used to notify SwiftUI.
- `favoriteSlotIndex`, the resolved slot index.
- `noteFavoriteSlot(_:)`, which stores the index and generates a new token.

`RootPaletteView` observes that token and dispatches the action to the currently active screen.

This preserves the existing architecture: AppKit captures the low-level keyboard event, while the active SwiftUI screen decides what the slot means.

### 4. Launcher favorite behavior

On `LauncherScreen`, `⌘1…⌘0` still launches favorites by position.

The previous SwiftUI key handler was removed. The behavior remains available in both palette sizes, subject to the existing rule that favorites must be enabled in compact mode.

Favorites beyond the tenth item remain visible and reorderable, but do not receive a keyboard slot.

### 5. Pinned clipboard shortcuts

`ClipboardScreen` now supports the same ten shortcuts through `activatePinned(at:)`.

A new `ClipboardStore.pinnedItem(at:in:filter:)` method resolves the requested item from the visible result set:

- The current search query is applied.
- The current clipboard type filter is applied.
- Only the visible pinned section is considered.
- Slot numbering follows the visible pinned order.
- Missing slots are ignored without producing an action.

For example, after filtering for `beta`, `⌘1` selects the first visible pinned result containing `beta`, rather than the first pinned item in the entire history.

### 6. Clipboard row UI

Pinned clipboard rows now receive an optional slot label. When the Command key is held, the first ten visible pinned rows display shortcuts such as:

- `⌘ 1`
- `⌘ 2`
- …
- `⌘ 0`

The labels are hidden otherwise, so the normal clipboard list remains visually compact.

## Memory footprint

| Step | Memory |
| --- | ---: |
| Idle after launch | 34.4 MB |
| Palette open | 40.1 MB |
| Feature in use (peak) | 44.3 MB |
| Palette closed, settled | 41.8 MB |

Leak-tested:

### Before

2026-08-22 at 23 57 35

https://github.com/user-attachments/assets/19acd787-ee49-4d8b-a892-465dec193538

As you can see the previous version was not compatible with multiple keyboard layout (espacially with French Layout (Azerty).

### After : using AppKit instead of SwiftUI's `onKeyPress`

The mapping uses hardware key codes rather than the character produced by the keyboard layout. This makes the shortcuts work consistently on layouts such as QWERTY and AZERTY.

https://github.com/user-attachments/assets/2f993a91-f453-4908-942a-391850bcc829

## Drawbacks

None known.

### Added tests

The commit adds coverage in two standalone harnesses.

#### `Tests/favorites-test.swift`

The favorites harness now verifies that:

- Physical key code `18` resolves to the first slot.
- Physical key code `25` resolves to the ninth slot.
- Physical key code `29` resolves to the tenth slot (`⌘0`).
- An unknown key code does not resolve to a slot.

The existing tests still verify the ten-slot limit, the `1…9,0` display order, the absence of an eleventh slot, negative-index handling, and the round-trip agreement between displayed digits and slot indices.

#### `Tests/clipboard-test.swift`

The new `pinnedSlotResolutionUsesVisiblePins()` case verifies that:

- The first and third slots resolve to the corresponding pinned rows.
- A slot beyond the visible pinned rows returns `nil`.
- A search query is applied before slot numbering.
- The visible order is preserved after filtering.
- A type filter is applied before slot numbering.
- Filtering can make a previously available slot unreachable.

Together, these tests protect the contract that shortcuts operate on the visible pinned block, rather than on the complete unfiltered clipboard history.

### Commands

```sh
./Scripts/run-tests.sh favorites-test
./Scripts/run-tests.sh clipboard-test
./Scripts/lint.sh
```

The tests exercise the shipped model sources through the repository's standalone harnesses. Manual validation should also confirm that `⌘1…⌘0` launches the expected favorite or pastes the expected visible pinned clipboard item, including with a query active and with the palette in compact mode.

I have taken your comments into account (https://github.com/abue-ammar/tinycast/pull/315#pullrequestreview-5001884414).

I made all those tests and it worked with both AZERTY and QWERTY.

## Manual Keyboard Validation

Run every keyboard test with both layouts:

- QWERTY: `Cmd+1` through `Cmd+9`, then `Cmd+0`
- AZERTY: the same physical number-row keys, including `Shift` where required

The physical key position must determine the slot, not the character produced by the layout.

### 1. Launcher Favorites

- [x] Create 10 launcher favorites.
- [x] QWERTY: physical keys 1 through 9 launch favorites 1 through 9.
- [x] QWERTY: physical key 0 launches favorite 10.
- [x] AZERTY: the same physical keys launch the same favorites.
- [x] The displayed labels are `1` through `9`, then `0`.
- [x] With fewer than 10 favorites, unused slots do nothing.
- [x] With 10 favorites, `Add to Favorites` is disabled.
- [x] Remove a favorite, then confirm a new favorite can be added.
- [x] Reorder favorites and confirm the slots follow the new order.

### 2. Launcher Compact Mode

- [x] Enable `Show favorites in compact mode`.
- [x] Open the launcher in compact mode.
- [x] QWERTY: test all 10 physical number-row keys.
- [x] AZERTY: test the same 10 physical keys.
- [x] Confirm visible favorites launch correctly.
- [x] Confirm slots 6 through 0 launch favorites not visible in the compact bar.
- [x] Disable `Show favorites in compact mode` while keeping favorites configured.
- [x] Repeat all 10 physical keys on QWERTY and AZERTY.
- [x] Confirm favorites still launch even though they are hidden.
- [x] Expand the palette and confirm the slot mapping is unchanged.

### 3. Clipboard Pinned Slots

- [x] Add and pin at least 3 clipboard entries.
- [x] QWERTY: each physical slot pastes the matching pinned entry.
- [x] AZERTY: the same physical keys paste the same entries.
- [x] Confirm slot order matches visible pin order.
- [x] Search for a subset of pinned entries and retest the slots.
- [x] Test text, link, image, and mixed clipboard filters.
- [x] Test an unavailable slot and confirm nothing unrelated is pasted.
- [x] Confirm pasting does not reorder pinned entries.
- [x] Hold `Cmd` and confirm the slot labels appear; release it and confirm they disappear.

### 4. Other Palette Screens

For each screen below, test all 10 physical number-row keys with QWERTY and AZERTY:

- [x] Snippets
- [x] Quicklinks
- [x] Quicklink argument form
- [x] Emoji picker
- [x] Calculator history
- [x] File search
- [x] Uninstall screen

### 5. Modifier and Existing Shortcut Behavior

- [x] QWERTY: test `Cmd+Shift+1`.
- [x] AZERTY: test the equivalent physical combination.
- [x] Confirm `Shift` does not prevent a favorite slot from working.
- [x] Test `Cmd+Option+1` and `Cmd+Control+1`; neither launches a favorite.
- [x] Hold a slot key and confirm key repeat does not cause unwanted repeated launches.
- [x] Test `Cmd+,` and confirm Settings opens.
- [x] Test `Cmd+.` and confirm the selected item is pinned where applicable.
- [x] Test `Cmd+W` and confirm the palette closes.
- [x] Test `Escape` and confirm existing escape behavior is unchanged.
- [x] Switch between QWERTY and AZERTY without restarting Tinycast.
- [x] Repeat a launcher slot and clipboard slot after switching layouts.

### 7. Final Regression and Shell Smoke Tests

- [x] Close and reopen Tinycast, then retest one launcher slot in both layouts.
- [x] Confirm favorite persistence after relaunch.
- [x] Confirm clipboard pin persistence after relaunch.

I found a bug during those tests (idk if it's intended) :
1. when you open a note, and use the clipboard history to paste something in a window, it's always pasted into the note (the Note always has the focus)

### DIFF
--- a/Tests/clipboard-test.swift
+++ b/Tests/clipboard-test.swift
@@ -13,6 +13,7 @@ struct ClipboardTests {
         pasteLeavesPinsAlone()
         pinsSurvivePruningAndTheWindow()
         pinsLeadFilteredSearches()
+        pinnedSlotResolutionUsesVisiblePins()
         textFormClassification()
         typeFilterSplitsTheHistory()
         typeFilterJoinsTheSearchMemo()
@@ -140,6 +141,44 @@ struct ClipboardTests {
             expect(
                 short.first?.text == "needle in the haystack",
                 "the pinned match leads the fallback search too")
+        }
+    }
+
+    /// Slot picks are taken from the visible pinned block after query/filter, not from all pins.
+    static func pinnedSlotResolutionUsesVisiblePins() {
+        withStore { store, _ in
+            store.addText("alpha one", sourceBundleID: nil)
+            store.addText("beta two", sourceBundleID: nil)
+            store.addText("beta three", sourceBundleID: nil)
+            store.addText("plain text", sourceBundleID: nil)
+
+            store.togglePinned(item(store, "alpha one"))
+            store.togglePinned(item(store, "beta two"))
+            store.togglePinned(item(store, "beta three"))
+
+            expect(
+                store.pinnedItem(at: 0, in: "", filter: .all)?.text == "alpha one",
+                "slot 1 maps to the first pinned row")
+            expect(
+                store.pinnedItem(at: 2, in: "", filter: .all)?.text == "beta three",
+                "slot 3 maps to the third pinned row")
+            expect(
+                store.pinnedItem(at: 3, in: "", filter: .all) == nil,
+                "missing pinned slot returns nil")
+
+            expect(
+                store.pinnedItem(at: 0, in: "beta", filter: .all)?.text == "beta two",
+                "query narrows the pinned block before slot mapping")
+            expect(
+                store.pinnedItem(at: 1, in: "beta", filter: .all)?.text == "beta three",
+                "slot mapping follows visible pinned order under query")
+            expect(
+                store.pinnedItem(at: 2, in: "beta", filter: .all) == nil,
+                "query can remove pinned slots from reach")
+
+            expect(
+                store.pinnedItem(at: 0, in: "", filter: .link) == nil,
+                "filter applies before pinned slot mapping")
         }
     }
 
--- a/Tests/favorites-test.swift
+++ b/Tests/favorites-test.swift
@@ -23,6 +23,14 @@ struct FavoritesTest {
         check("⌘0 is the tenth, not the first", FavoriteSlots.index(for: "0") == 9)
         check("a letter is not a slot", FavoriteSlots.index(for: "a") == nil)
 
+        let numberRowKeyCodes: [UInt16] = [18, 19, 20, 21, 23, 22, 26, 28, 25, 29]
+        for (index, keyCode) in numberRowKeyCodes.enumerated() {
+            check(
+                "physical number-row keycode \(keyCode) maps to slot \(index)",
+                FavoriteSlots.index(forKeyCode: keyCode) == index)
+        }
+        check("unknown keycode is not a slot", FavoriteSlots.index(forKeyCode: 0) == nil)
+
         check("the first row shows 1", FavoriteSlots.digit(at: 0) == "1")
         check("the tenth row shows 0", FavoriteSlots.digit(at: 9) == "0")
         check("the eleventh row shows nothing", FavoriteSlots.digit(at: 10) == nil)
--- a/Tinycast/Features/Clipboard/Model/ClipboardStore.swift
+++ b/Tinycast/Features/Clipboard/Model/ClipboardStore.swift
@@ -299,6 +299,12 @@ final class ClipboardStore {
         search(query, filter: filter).firstIndex { $0.id == item.id }
     }
 
+    /// The Nth visible pinned entry under `query` and `filter`, where 0 is the first pinned row.
+    func pinnedItem(at index: Int, in query: String, filter: ClipboardFilter) -> ClipboardItem? {
+        guard index >= 0 else { return nil }
+        return search(query, filter: filter).prefix(while: \.isPinned).dropFirst(index).first
+    }
+
     private func unfiltered(_ q: String) -> [ClipboardItem] {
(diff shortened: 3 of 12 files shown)


abue-ammar:
hey thanks for the pr. review it again and let you know my feedbacks asap

PR #347, opened 2026-08-25
Tint an extension action's icon in the ⌘K panel

list-jonas:
## Related issue

None filed — hit it with the Hugeicons UI extension.

## What changed

An `Action`'s `icon` is a full `ImageLike`, so it can name a source and tint it — Hugeicons builds its colour picker from `{source: Icon.Circle, tintColor: Color.Red}` rows. The panel's rows were `PopoverMenuItem`, whose glyph is a bare `.symbol(name)` or `.file(path)` and carries no tint, so `ExtensionActionsMenu` pulled the SF Symbol name out of each resolved icon and dropped the rest. Every row painted `Color.secondary`: ten grey circles where Raycast shows ten colours.

Rows are `ExtensionActionItem` now, carrying a resolved `ExtensionImage`, so themed `{light, dark}` tints, raw hex and `mask` behave as they do in a list. It stays inside `Features/Extensions/` per `AGENTS.md` — duplicating `PopoverMenuItem`'s shape is what keeps it there.

`ExtensionImage.actionIcon` holds the two panel-only rules: an action with no usable icon draws a glyph rather than an empty slot, and a destructive one with no tint takes red. Red is symbols only — a tint masks artwork, so a destructive PNG would otherwise draw as a red silhouette.

Two smaller things: the resolve call passed `assetsPath: nil`, so an action naming a bundled asset got the generic bolt. And `ExtensionCommandScreen.actions(at:)` returns nil by design now, since a command's panel isn't expressible as `PopoverMenuContent` — `menuRowActions` is the one row order the panel, ↑/↓ and ↵ share.

## Demo
Before:
file-a20e3022b8100e21a7c00bcbdee5537d
After:
file-50efead2624e5ef28effa33d8d5cb59a
Raycast:
image

## Drawbacks

A tinted artwork icon is masked rather than recoloured. That matches Raycast and the existing list behaviour, but an extension tinting a multi-colour PNG loses its colours.

## Tests & validation

- `ext-test` gains `actionIconChecks` — tint kept, raw hex, a themed tint picking the right side, asset names, the destructive fallback, an explicit tint beating it, and destructive artwork staying untinted. Needs `ExtensionImage` in the harness's source list.
- All 39 harnesses pass, Debug builds with no new warnings, `lint.sh` clean.
- Ran the installed Hugeicons bundle through the harness: `Red #FF4245`, `Orange #FF9230`, `Green #30D158` — distinct where they were one grey.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -197,15 +197,21 @@ run ext-store-test         $E/Model/ExtensionRegistry.swift \
                            $E/Model/ExtensionPackageManager.swift \
                            $E/Model/ExtensionStoreResponse.swift
 run slow ext-test          -parse-as-library \
+                           Tinycast/Platform/Appearance.swift \
+                           Tinycast/Platform/Images/IconCache.swift \
+                           Tinycast/DesignSystem/Theme.swift \
                            $E/Model/ExtensionBootConfig.swift \
                            $E/Model/ExtensionManifest.swift \
                            $E/Model/RenderNode.swift \
                            $E/Service/ExtensionCatalog.swift \
                            $E/Service/ExtensionFetcher.swift \
+                           $E/Service/ExtensionIconCache.swift \
                            $E/Service/ExtensionNodeShims.swift \
                            $E/Service/ExtensionOAuthKeychain.swift \
                            $E/Service/ExtensionOAuthSession.swift \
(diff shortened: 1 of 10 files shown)


abue-ammar:
hey nice pr man!. this is a loat cleaner than the spaghetti that i had before. imma merge it

PR #349, opened 2026-08-26
Open the join preview on a live camera, and drop finished events from the launcher

### LINKED ISSUE #348: [Bug]: Calendar Feature - meetings
### Existing issues

- [ ] I searched open and closed issues and didn't find this already reported

### What happened?

I have tested the calendar feature and this is my recap :

# what is working

The event in the menu bar is correctly showing, you can see a demo :
Image

the camera preview and the confirmation before joining is fine.

# what is not working / should be changed
1. When an event is finished, the event stop showing in the menu bar but should also disappearing from the launcher: you can still see "Test" event, even though it's already finished some time ago.

Image

---

Image

2. I made a video available here : https://streamable.com/hokpeee. It's showing the camera preview which works fine, but when you launch the event it's flickering I don't know if it's on your part or if it's Apple the culprit 😅...

3. You should also keep an option to open the calendar event to show more informations like the description or the location ... it's only available when the event is nearly starting.

Image

### Steps to reproduce

Configure the Calendar Extension like this :

Image

# Hide current event

1. set an event a little forward (7 min later) and like 3 to 5 min duration with facetime
2. wait for the event to show in the menu bar 
3. wait the end of the event (the event in the menu bar disappear, but not in the launcher)

# Event flickering
1. open the event you have created
2. verify that the facetime is flickering (it's weird it's not always the case sometimes it's perfect and sometimes it's flickering like 1 second or 2)

### Screenshots or recording

already shown

0.9.21-beta.66

### macOS version

26.5.2

abue-ammar:
## Related issue

Closes #348

> **Scope note:** this covers items 1 and 2 of #348. Item 3 — `Open in Calendar` on a meeting
> **launcher entry**, which falls through to `AppActionsMenu` instead of `MeetingActionsMenu` — is
> not in here. Worth splitting into its own issue before merging this one closed.

### The join preview flickered (#348, item 2)

Four defects stacked on the same 300 ms:

- **The panel opened before the camera ran.** `show()` faded the panel in and *then* kicked off
  `Task { await session.start() }`. `startRunning` blocks for a few hundred milliseconds, so the
  panel was always fully visible over an empty stage first, with the feed popping in afterwards.
- **The stage branched on state nothing observed.** It read `session.session`, which was
  `@ObservationIgnored`. The placeholder→feed swap happened only because `start()` re-assigned
  `access` and `@Observable` fires on every set, unchanged value or not — an accident holding up
  the one transition the user actually sees.
- **`updateNSView` rebuilt the preview connection on every update.** `preview.session = session`
  ran unconditionally, and that setter tears down and re-establishes the connection, blanking the
  layer. It fired at least twice per present (the `access` mutation, then `PanelEntrance.appeared`)
  — this is the blank that lands *after* frames are already flowing.
- **The hosting view was layer-backed, not layer-hosting.** `wantsLayer` was set before
  `view.layer`, the opposite of the order [`NSView.layer`](https://developer.apple.com/documentation/appkit/nsview/layer)
  documents for a custom layer. AppKit owns the layer that way, may replace it, and applies
  implicit animations to its bounds.

`CameraPreviewSession` now resolves access, configures and blocks on `startRunning` **before**
anything is shown, then hands back a settled `Feed` — `.live` / `.denied` / `.noCamera`. The panel's
first frame is live video, the view is a pure function of one value, and `@Observable` goes with the
three fields it published. `Starting the camera…` is unreachable and deleted.

Two consequences fall out of the same shape:

- `requestCameraAccess()` ran from `start()` **after** the panel was up, so on a first join the TCC
  prompt took key, `windowDidResignKey` fired, and the join was **silently dropped**. The prompt now
  comes first.
- `session.stop()` ran the instant the user answered, tearing the camera down under a still-visible
  panel. It now runs from a new `fadeOut(duration:done:)` completion — the symmetric twin of the
  closure `fadeIn` already takes — skipped when a preview raised inside the fade already owns the
  camera. The two existing `fadeOut` callers take the default and are untouched, and `cancelFade()`
  is HUD-only, so the rescue path can never strand a running camera.

### A finished event stayed in the launcher (#348, item 1)

`UpcomingWindow.agenda` is documented as *the* place that says which events count, but it had no
notion of `now` — so an event that ended an hour ago was still "timed, not declined". It now takes
one and filters `end > now`. Every caller already had a `now` in hand, and **none of them change
behaviour**: `carded`, `joinable` and `MenuBarSummary` already excluded finished events with their
own predicates, which is exactly why the menu bar was right and nothing else was.

The filter alone isn't enough for the launcher slice, because those entries are a *published
snapshot*: `CalendarStore.publish` short-circuits on an unchanged array, and an event ending changes
nothing in EventKit. So `CalendarCoordinator` republishes on the minute boundary and on every summon
— the second matters because with the menu bar and auto join both off, no clock runs at all while
the palette is closed.

## Memory footprint

Leak-tested:

No new long-lived state: `CameraPreviewSession` loses three stored properties and its `@Observable`
conformance, and the `AVCaptureSession` it caches for reuse is unchanged from before. The one new
closure (`fadeOut`'s `done:`) captures `self` weakly.

## Drawbacks

**The panel now appears ~200–500 ms after ↵ rather than immediately**, because it waits on
`startRunning`. This was a deliberate choice over the alternative — opening instantly on a stable
stage and cross-fading the video in — because waiting *deletes* the swap rather than animating over
it: no Observation, no second animation on top of the window fade, and the view becomes a pure
function of one settled value. The palette dismissing is the immediate feedback that ↵ registered.
If the delay reads badly in practice, the cross-fade is the fallback.

`My Schedule` now also drops a meeting when it ends, which follows from `agenda` being one rule
rather than several. That is beyond what #348 asked for, but the alternative is a second rule.

## Tests & validation

- `./Scripts/run-tests.sh` — all 45 harnesses pass.
- `Tests/calendar-test.swift` (`agendaFiltering`) gains two cases: a meeting is off the agenda the
  moment it ends, and one still running stays however long ago it started. Existing `agenda(from:)`
  call sites in `agendaFiltering` / `cardWindow` take the new `now:`.
- Debug build: no new warnings. `./Scripts/lint.sh`: clean. The `Model/` AppKit/SwiftUI grep from
  `AGENTS.md` still returns nothing.

**Not yet exercised by hand** — the flicker is presentation code no harness reaches, so these were
read out of the code rather than watched on screen:

- [ ] `tccutil reset Camera com.tinycast.app.dev`, then join from the launcher: the prompt appears
      **before** any panel, and granting completes the join rather than dropping it.
- [ ] Second and third joins in the same launch: the panel fades in already showing video — no
      `Starting the camera…`, no black frame, no mid-life blank.
- [ ] ↵ to join: video stays up through the fade, and the camera light goes out as the panel leaves.
- [ ] A 3-minute event that ends leaves the launcher results and `My Schedule` on the same minute
      boundary it leaves the menu bar, with the palette open and closed over the end.

### DIFF
--- a/Tests/calendar-test.swift
+++ b/Tests/calendar-test.swift
@@ -145,8 +145,16 @@ struct CalendarTests {
             event(id: "allday", start: 30, isAllDay: true),
             event(id: "declined", start: 30, isDeclined: true)
         ]
-        let agenda = UpcomingWindow.agenda(from: events)
+        let agenda = UpcomingWindow.agenda(from: events, now: at(0))
         expect(agenda.map(\.id) == ["early", "late"], "the agenda is timed, accepted and in start order")
+
+        let over = event(id: "over", start: 0, minutes: 30)
+        expect(
+            UpcomingWindow.agenda(from: [over], now: at(30)).isEmpty,
+            "a meeting is off the agenda the moment it ends, the way it leaves the menu bar")
+        expect(
+            UpcomingWindow.agenda(from: [over], now: at(29)).map(\.id) == ["over"],
+            "one still running stays, however long ago it started")
     }
 
     static func cardWindow() {
@@ -179,7 +187,7 @@ struct CalendarTests {
             window.carded(from: [linkless], now: start) == nil,
             "a meeting with no link is never carded")
         expect(
-            UpcomingWindow.agenda(from: [linkless]).map(\.id) == ["linkless"],
+            UpcomingWindow.agenda(from: [linkless], now: start).map(\.id) == ["linkless"],
             "but it stays on the agenda, so it is still listed and searchable")
     }
 
--- a/Tinycast/DesignSystem/Interaction/PanelTransition.swift
+++ b/Tinycast/DesignSystem/Interaction/PanelTransition.swift
@@ -21,8 +21,9 @@ extension NSWindow {
         }
     }
 
-    /// Safe to interrupt: the handler checks opacity, so `cancelFade` can rescue it.
-    func fadeOut(duration: TimeInterval) {
+    /// Safe to interrupt: the handler checks opacity, so `cancelFade` can rescue it. `done` runs
+    /// once the panel is actually off screen, which is where teardown the user can see belongs.
+    func fadeOut(duration: TimeInterval, done: (@MainActor () -> Void)? = nil) {
         NSAnimationContext.runAnimationGroup { context in
             context.duration = duration
             context.timingFunction = CAMediaTimingFunction(name: .easeIn)
@@ -31,6 +32,7 @@ extension NSWindow {
             MainActor.assumeIsolated {
                 guard let self, self.alphaValue == 0 else { return }
                 self.orderOut(nil)
+                done?()
             }
         }
     }
--- a/Tinycast/Features/Calendar/Model/MenuBarSummary.swift
+++ b/Tinycast/Features/Calendar/Model/MenuBarSummary.swift
@@ -13,7 +13,7 @@ struct MenuBarSummary: Sendable {
 
     /// The earliest event still inside its window, so one hiding hands the space to the next.
     func event(from events: [MeetingEvent], now: Date) -> MeetingEvent? {
-        UpcomingWindow.agenda(from: events).first {
+        UpcomingWindow.agenda(from: events, now: now).first {
             (!linkedOnly || $0.link != nil) && now >= $0.start - lead && now < hidesAt($0)
         }
     }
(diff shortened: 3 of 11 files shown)


mathiasdotdev:
> Item 3 — Open in Calendar on a meeting
> launcher entry, which falls through to AppActionsMenu instead of MeetingActionsMenu — is
> not in here. Worth splitting into its own issue before merging this one closed.

@abue-ammar do you want me to open a new issue ?

abue-ammar:
> > Item 3 — Open in Calendar on a meeting
> > launcher entry, which falls through to AppActionsMenu instead of MeetingActionsMenu — is
> > not in here. Worth splitting into its own issue before merging this one closed.
> 
> @abue-ammar do you want me to open a new issue ?

i am releasing all these in Tinycast 0.9.22 · beta . test them lemme know what still need to be fixed. in discord maybe

mathiasdotdev:
ok i'll try this tonight i'll tell you

PR #363, opened 2026-08-27
Draw a raycast-* colour named inside an extension's own SVG, and keep accessories in a detail list

list-jonas:
## Related issue

None filed — hit it with a quota extension's usage ring.

## What changed

A `raycast-*` colour name is legal wherever a Raycast tint is, so an extension drawing its own SVG writes one straight into `stroke`. No renderer knows the keyword, so the shape draws nothing at all: a usage ring came out as an empty slot rather than an arc.

The name resolves to `rgba(…)` during the decode, in `ExtensionIconCache.loadInlineAsync` — already detached, already holding the bytes. Doing it in `resolve` instead would put it in a `body`, re-decoding and re-encoding every inline SVG on the main actor on each hover and selection move: 50 rows of a 6 KB chart cost 0.45 ms a pass at decode versus 12.9 ms there.

Decoding is also the only place that reaches the markdown images a Detail draws, which never pass through `resolve` and so kept drawing nothing.

The palette arrives as `[name: css]` from `ExtensionImage.svgPalette(isDark:)`, flattened once per appearance into two `static let`s — a `Color` only resolves to sRGB on the main actor, which is exactly what the decode must not touch. `ExtensionImage.palette` is now the one source both the SwiftUI tint and the CSS rewrite read from, replacing the switch that listed the same nine names.

The rewrite walks whole names rather than searching for known ones: `raycast-red` sits inside `raycast-red-invented`, and substituting it there would corrupt a name we can't resolve. It early-outs on a `.literal` scan, since the default search is Unicode-canonical and costs ~3x on a long payload. The media-type guard covers a percent-encoded payload too, so a photo is never walked as text.

Both call sites key their `.task` on `ExtensionImage.LoadKey` rather than the source alone — the URL no longer says what will be drawn, and without it a dark/light flip leaves the old ink on screen.

Separately, a List with `isShowingDetail` dropped each row's accessories. The API only *advises* an extension against sending them there; Raycast draws the ones it's given, and a quota row's entire signal is in them. They're drawn now, with `layoutPriority(1)` on the title so a 290pt detail row doesn't truncate it.

## Demo
Before:
image
image

After:
image
image

Raycast:
image
image

## Drawbacks

The rewrite is textual, not attribute-scoped, so a name is substituted wherever it appears in an SVG that names one. `<text>raycast-green</text>` becomes the rgba string, and a `raycast-blue` gradient `id` is rewritten along with its `url(#…)` ref — the pair still matches, but reads oddly. `class="raycast-red-box"` and an uppercase `raycast-Green` are both correctly left alone. Scoping it properly means parsing XML in the decode path, which costs more than the bug is worth.

A long title in a detail list squeezes its accessories out: unchanged to ~20 chars, 47pt → 15pt at 24, effectively gone past 26. Giving the accessory the priority instead truncates titles on every row rather than a few, which reads worse.

## Tests & validation

- `ext-icon-test` gains `paletteColorsInSVGResolve`, asserting on drawn pixels rather than the rewritten string — the string proves nothing about whether the shape draws. Covers both encodings, a trailing `?raycast-width=` query, an unknown name shadowing a known one, the light and dark ramps landing on white and black ink, and a PNG passing through untouched. Needs `ExtensionImage` and its dependencies in the harness's source list.
- All 39 harnesses pass, Debug builds with no new warnings, `lint.sh` clean.
- Merges into current `main` without conflict, and `ext-icon-test` passes on the merged tree — `main` has since touched both `ExtensionImage.swift` and this harness.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -114,7 +114,20 @@ run entry-icon-test        Tinycast/Platform/Appearance.swift \
                            Tinycast/Platform/Images/IconCache.swift
 run ext-icon-test          Tinycast/Platform/Appearance.swift \
                            Tinycast/Platform/Images/IconCache.swift \
-                           Tinycast/Features/Extensions/Service/ExtensionIconCache.swift
+                           Tinycast/Platform/Compression/Zlib.swift \
+                           Tinycast/DesignSystem/Theme.swift \
+                           Tinycast/Features/Extensions/Model/ExtensionBootConfig.swift \
+                           Tinycast/Features/Extensions/Model/ExtensionManifest.swift \
+                           Tinycast/Features/Extensions/Model/RenderNode.swift \
+                           Tinycast/Features/Extensions/Service/ExtensionCatalog.swift \
+                           Tinycast/Features/Extensions/Service/ExtensionFetcher.swift \
+                           Tinycast/Features/Extensions/Service/ExtensionNodeShims.swift \
+                           Tinycast/Features/Extensions/Service/ExtensionOAuthKeychain.swift \
+                           Tinycast/Features/Extensions/Service/ExtensionOAuthSession.swift \
+                           Tinycast/Features/Extensions/Service/ExtensionRuntime.swift \
+                           Tinycast/Features/Extensions/Service/ExtensionIconCache.swift \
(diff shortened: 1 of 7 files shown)


abue-ammar:
this looks like a good pr!  i will review it ASAP

Issue #353, opened 2026-08-26
[Feature]: Business days & workdays arithmetic in calculator


pazhanir:
### What should it do?

Extend `CalcDateTime` to support adding and subtracting business days (weekdays / workdays), skipping weekends (Saturday & Sunday):

- Support duration phrases:
  - `today + 5 business days` / `today - 3 business days`
  - `tomorrow + 10 work days` / `5 weekdays from now`
  - `august 26 2026 + 15 workdays`
- Support flexible 3-atom dates like `august 26 2026`, `26 august 2026`, `2026 august 26`.
- Accurately step across calendar day boundaries skipping Saturday and Sunday while keeping arithmetic DST-safe (`Calendar.startOfDay` iteration).

### Why does it belong in Tinycast?

- **Zero overhead / Pure Foundation**: Implemented entirely within `Tinycast/Features/Calculator/Model/CalcDateTime.swift` using existing `Calendar` APIs — zero third-party dependencies, zero binary bloat.
- **High daily utility**: Calculating deadlines, project turnarounds, shipment ETAs, and contract review dates in business days directly from the command palette without needing dedicated websites or Python scripts.
- **Strictly tested**: Covered by pure model unit tests (`calc-test`).

abue-ammar:
i dont want it because it is depended on calendar. in tinycast any feature can be turned off. so if someone turns of calendar. it wont work. so that's why this issue is invalid to me. but you can add other calculations that raycast has but we dont.

PR #379, opened 2026-08-28
Answer time zones, business days and the spelled-out phrasings the calculator was missing

list-jonas:
## Related issue

Addresses **#353**, closed as invalid on the grounds that business days depend on the Calendar feature and would break when it's switched off. That isn't the case here.

## What changed

**Time zones.** `CalcTimeZone` answers `time in Tokyo`, `5pm ldn in sf` and `diff paris`, built on **Apple's Foundation**: `TimeZone.knownTimeZoneIdentifiers` already carries the whole IANA database, so nothing is generated and no copy of tzdata is committed — the table updates with macOS. The city index is a **lazy `static let`**, built on the first zone query and never rebuilt (443 identifiers, ~18 ns a lookup); a query that names no zone never touches it. `aliases` is the hand-written half: abbreviations, nicknames (`sf`, `nyc`, `ldn`) and IATA codes. The home zone comes off the **injected `Calendar`**, so `Model/` performs no environment read.

**Dates.** Business days skip weekends, plus dotted and ordinal spellings (`19.2.27`, `28. aug`), bare dates, and `monday in 3 weeks`. A date answer now badges its weekday instead of `Result`.

**Everything else.** Slashed rates, timespans, the rest of the trig set, `square root of`, `power`, tip/ratio/aggregate phrasings, and a mid-expression `to` so `100 km/h to mph + 3mph` no longer needs parentheses.

## No dependencies, no permissions, no calendar events

Two unrelated things share the word: `Foundation.Calendar` is pure date math compiled into the system, EventKit is your events and the thing that's TCC-gated. Only the first is used — a grep for `EventKit`, `EKEventStore` or `calendarEnabled` across `Features/Calculator/` returns nothing, so business days work with the Calendar feature **off** and nothing ever prompts. Public holidays are deliberately not modelled, since EventKit is their only source and the answer would then differ between two Macs. The one import across the whole diff is `Foundation`.

## Examples

| Query | Answer |
| --- | --- |
| `today + 5 business days` | 4 September · Friday |
| `5 weekdays from now` | 4 September · Friday |
| `august 26 2026 + 15 workdays` | 16 September · Wednesday |
| `28. aug + 3` | 31 August · Monday |
| `time in vie` | 2:00 PM · UTC → Vienna |
| `5pm ldn in sf + 2h` | 11:00 AM |
| `100 km/h to mph + 3mph` | 65.13711922 mph |
| `ratio of 1920 to 1080` | 16 : 9 |

## Drawbacks

`1.2.24` is both a plausible semver and a valid 1 February 2024, and nothing separates them — a one- or three-digit tail is rejected, but a two-digit patch reads as a date. Requiring a four-digit year would close it and cost `19.2.27`.

`20 eur to usd * 30` still wants parentheses: it has two honest readings where `+` and `-` have one.

## Tests & validation

- `calc-test` gains ~170 assertions: every grammar above, the ambiguity guards (`10 in in cm` stays a conversion, `mad` stays the dirham), weekend boundaries from Friday, Monday and Saturday anchors, and the malformed inputs that used to trap.
- All 49 harnesses pass, Debug builds with no new warnings, `lint.sh` clean, `Model/` stays Foundation-only.
- Hot path measured against `main` over 146 app-name prefixes: **2.04 µs → 2.08 µs per keystroke**, so a search that isn't a calculation costs the same. Release bundle 9296 KB → 9396 KB.

### DIFF
--- a/Tests/calc-test.swift
+++ b/Tests/calc-test.swift
@@ -821,6 +821,12 @@ struct CalcTests {
         expectNil("is what % of")
         expectNil("tip on")
 
+        // A pasted non-breaking space still reaches the zone table, since the gate scans for
+        // whitespace rather than a literal space. Arithmetic splits on " + " and never did.
+        expectDisplayAt("time\u{a0}in\u{a0}tokyo", "9:18 AM")
+        expectDisplayAt("time\u{9}in\u{9}tokyo", "9:18 AM")
+        expectDisplayAt("time\u{2009}in\u{2009}tokyo", "9:18 AM")
+
         // The ordinal dot German and Austrian dates write after the day
         expectDisplayAt("28. aug + 3", "31 August")
         expectDisplayAt("28. august + 3", "31 August")
--- a/Tests/calc-test.swift
+++ b/Tests/calc-test.swift
@@ -610,8 +610,12 @@ struct CalcTests {
         expectDisplay("10$", "835.00 INR", region: "INR")
         expectDisplay("1 btc", "5,010,000.00 INR", region: "INR")
         expectCopy("1 usd", "83.50 INR", region: "INR")
-        // Nothing to say: the region names the currency written, one nobody quotes, or none at all
-        expectDisplay("1 usd", "1.00 USD", region: "USD")
+        // The region names the currency written, so the dollar pairs with the euro instead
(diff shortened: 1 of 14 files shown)


list-jonas:
The calculator still isn't 100% on par with the Raycast one, but quite close to it.

PR #392, opened 2026-08-30
Resolve palette shortcuts through the ASCII-capable keyboard layout

JaeSeoKim:
## Related issue

N/A — no directly matching issue was found.

### Problem

Palette-internal shortcuts were matched against the character produced by the active input source.

For example, with a non-ASCII input source or IME active, opening the palette and pressing
Command-K did not open Actions. Switching to an English input source made the same physical shortcut
work. Global shortcuts were unaffected because Carbon registers them with hardware key codes and
modifier masks. Internal shortcuts instead passed through the focused field editor and SwiftUI
`KeyEquivalent` matching, so the input source could change the character before a filtered
`.onKeyPress(keys: ["k"])` handler saw it.

### Resolution

Added `ASCIIKeyboardLayout`, which resolves a hardware key code through macOS's current
ASCII-capable keyboard layout.

Character shortcut handlers now accept every key press and compare the logical ASCII key inside the
callback, using `NSApp.currentEvent.keyCode`. Panel-owned chords and extension action shortcuts use
the same translation. This keeps Command-K and the other Command/Control palette shortcuts working
while a non-ASCII input source or IME is active.

The translation:

- Applies only to Command and Control chords, leaving ordinary typing, Option-only input, dead keys,
  and IME composition unchanged.
- Falls back to SwiftUI's original `KeyEquivalent` when an ASCII translation is unavailable.
- Uses the current ASCII-capable layout, so Dvorak, Colemak, and similar layouts retain their logical
  letter positions.
- Does not synthesize or redispatch replacement `NSEvent` instances.

`KeyShortcut` also uses the shared translator when rendering shortcut glyphs, keeping the displayed
shortcut and the key used for matching consistent. The hotkey harness and manual regression
documentation now cover this behavior.

## Memory footprint

Measured with `/usr/bin/footprint` on the ad-hoc-signed Debug build on macOS 26.5.2.

| Idle after launch       | 42 MB  |
| Palette open            | 41 MB  |
| Feature in use (peak)   | 45 MB  |
| Palette closed, settled | 44 MB  |

Leak-tested: Repeated Command-K / Backspace 11 times and then closed the palette. Peak footprint
stayed below 53 MB. `leaks` reported no retained synthetic `NSEvent` and no Tinycast-owned allocation
chain; its remaining 15.2 KB was in macOS AppIntents XPC objects also present in the baseline process.

## Drawbacks

- Modified character handlers receive every key press and immediately ignore non-matching keys.
- Resolving a Command or Control chord queries the current ASCII-capable layout on the main actor.
  This avoids caching a layout that can change while Tinycast is running.
- Matching uses `NSApp.currentEvent` while SwiftUI invokes the key-press callback. If that event is
  unavailable or cannot be translated, matching falls back to SwiftUI's original key.

## Tests & validation

- All 50 test harnesses passed.
- Extended `Tests/hotkey-test.swift` to cover ASCII-capable keyboard layout translation.
- SwiftLint passed with only existing repository warnings.
- Pure-layer import validation passed.
- Debug configuration built successfully.
- Manually verified Command-K with a non-ASCII input source active.
- Verified ordinary IME text entry remains unchanged.
- Verified global shortcuts continue to work.
- Repeated Command-K 11 times and confirmed that replacement `NSEvent` objects do not accumulate.

### DIFF
--- a/Tests/hotkey-test.swift
+++ b/Tests/hotkey-test.swift
@@ -87,11 +87,19 @@ struct DoubleTapDetectorTests {
     }
 
     static func layoutCharacters() {
-        let character = ASCIIKeyboardLayout.character(for: kVK_ANSI_K)
-        expect(character?.isEmpty == false, "the ASCII-capable layout translates an ANSI letter key")
+        let keyCodes = [kVK_ANSI_K, kVK_ANSI_X, kVK_ANSI_Q, kVK_ANSI_Comma, kVK_ANSI_Period]
+        let characters = keyCodes.compactMap { ASCIIKeyboardLayout.character(for: $0) }
         expect(
-            character?.unicodeScalars.allSatisfy(\.isASCII) == true,
+            characters.count == keyCodes.count,
+            "the ASCII-capable layout translates every ANSI key a palette chord uses")
+        expect(
+            characters.allSatisfy { $0.unicodeScalars.allSatisfy(\.isASCII) },
             "the shortcut character stays ASCII while a non-ASCII input source is active")
+        expect(
+            keyCodes.allSatisfy {
+                ASCIIKeyboardLayout.character(for: $0, modifiers: UInt32(cmdKey >> 8)) != nil
+            },
+            "a layout's Command table resolves the same keys, so ⌘ chords never lose their letter")
     }
 
     // MARK: - Built-in command mappings
--- a/Tests/hotkey-test.swift
+++ b/Tests/hotkey-test.swift
@@ -56,6 +56,7 @@ struct DoubleTapDetectorTests {
     static func main() {
         modifierGlyphs()
         commandActions()
+        layoutCharacters()
         hyperChord()
         hyperRetargeting()
         firing()
@@ -83,6 +84,14 @@ struct DoubleTapDetectorTests {
             DoubleTapModifier.allCases.map(\.rawValue)
                 == ["control", "option", "shift", "command"],
             "raw values are the persisted spelling and stay in canonical ⌃⌥⇧⌘ order")
+    }
+
+    static func layoutCharacters() {
+        let character = ASCIIKeyboardLayout.character(for: kVK_ANSI_K)
+        expect(character?.isEmpty == false, "the ASCII-capable layout translates an ANSI letter key")
+        expect(
+            character?.unicodeScalars.allSatisfy(\.isASCII) == true,
+            "the shortcut character stays ASCII while a non-ASCII input source is active")
     }
 
     // MARK: - Built-in command mappings
--- a/Tinycast/Palette/PalettePanel.swift
+++ b/Tinycast/Palette/PalettePanel.swift
@@ -64,7 +64,9 @@ final class PalettePanel: NSPanel {
         else { return nil }
         let arrow: (key: KeyEquivalent, code: Int)
         // Character chords, not key codes: Dvorak transposes the two.
-        switch event.charactersIgnoringModifiers?.lowercased() {
+        switch ASCIIKeyboardLayout.character(for: Int(event.keyCode))?.lowercased()
+            ?? event.charactersIgnoringModifiers?.lowercased()
+        {
         case "n": arrow = (.downArrow, kVK_DownArrow)
         case "p": arrow = (.upArrow, kVK_UpArrow)
         case "f": arrow = (.rightArrow, kVK_RightArrow)
--- a/Tinycast/Palette/PalettePanel.swift
+++ b/Tinycast/Palette/PalettePanel.swift
@@ -63,8 +63,8 @@ final class PalettePanel: NSPanel {
         guard event.modifierFlags.intersection([.command, .option, .control, .shift]) == .control
         else { return nil }
         let arrow: (key: KeyEquivalent, code: Int)
-        // Character chords, not key codes: Dvorak transposes the two.
-        switch ASCIIKeyboardLayout.character(for: Int(event.keyCode))?.lowercased()
+        // Through the ASCII-capable layout: an IME must not move ⌃N off its physical key.
+        switch ASCIIKeyboardLayout.character(for: event)?.lowercased()
             ?? event.charactersIgnoringModifiers?.lowercased()
         {
         case "n": arrow = (.downArrow, kVK_DownArrow)
--- a/Tinycast/Palette/PaletteWindowController.swift
+++ b/Tinycast/Palette/PaletteWindowController.swift
@@ -266,9 +266,10 @@ final class PaletteWindowController: NSObject, NSWindowDelegate {
                 self.core.palette.prepare(mode: .launcher)
                 return true
             }
-            // Character chords, not key codes: Dvorak transposes the two.
-            guard let character = ASCIIKeyboardLayout.character(for: Int(event.keyCode))?
-                .lowercased() ?? event.charactersIgnoringModifiers?.lowercased()
+            // Through the ASCII-capable layout: an IME must not move ⌘W off its physical key.
+            guard
+                let character = ASCIIKeyboardLayout.character(for: event)?.lowercased()
+                    ?? event.charactersIgnoringModifiers?.lowercased()
             else { return false }
             switch character {
             case ",":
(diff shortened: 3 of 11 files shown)


abue-ammar:
please create issue what's the issue why do we need to solve this, attach related issues. and how reliable is this solution? this entire project is based of keybaord shortcut. you have tested it with IME stuffs have you tested all the keybaord shortcut based stuffs not just Emacs but all the others . our FIRST class users are `US-QWERTY` how reliable is this pr that QWERTY users wont face any kinda issue.  and state all the drawbacks properly were they present before this changes or no. your pr will be closed without explanation if an AI reply this.

image

Tinycast/Platform/ASCIIKeyboardLayout.swift
```diff
@@ -0,0 +1,50 @@
...
+        return String(utf16CodeUnits: characters, count: length)
+    }
+
+    /// SwiftUI exposes the input-source character; recover the logical ASCII key from AppKit.
+    @MainActor static func keyEquivalent(fallingBackTo key: KeyEquivalent) -> KeyEquivalent {
+        guard let event = NSApp.currentEvent,
+            !event.modifierFlags.isDisjoint(with: [.command, .control]),
+            let character = character(for: Int(event.keyCode))?.lowercased().first,
+            character.unicodeScalars.allSatisfy(\.isASCII)
+        else { return key }
```
abue-ammar: **Shifted chords break whenever this fallback is taken, on every layout including US QWERTY.**

`matches` compares against a lowercased character, but the fallback returns `key` untouched. With Shift held, SwiftUI hands you `"X"`, so `matches(press.key, character: "x")` is false and the shortcut is dropped.

The old code covered this explicitly with `keys: ["x", "X"]`, `["f", "F"]` and `["q", "Q"]`, so this PR loses **⌃⇧X, ⇧⌘F and ⌃⇧Q** on the fallback path.

Lowercasing the fallback closes it. `.first` rather than `Character(_:)` because a few uppercase scalars lowercase to more than one character.

```suggestion
        else { return key.character.lowercased().first.map(KeyEquivalent.init) ?? key }
```

Tinycast/Platform/ASCIIKeyboardLayout.swift
```diff
@@ -0,0 +1,50 @@
...
+            CFDataGetBytePtr(layoutData), to: UnsafePointer<UCKeyboardLayout>.self)
+        var deadKeyState: UInt32 = 0
+        var length = 0
+        var characters = [UniChar](repeating: 0, count: 4)
+
+        let error = UCKeyTranslate(
+            keyLayout,
+            UInt16(keyCode),
+            UInt16(kUCKeyActionDisplay),
+            0,
```
abue-ammar: **This regresses the "Dvorak - QWERTY ⌘" input source.** That layout exists to give QWERTY letter positions *while Command is held*, and it encodes that in the layout's `cmdKey` modifier table. Passing `0` here skips that table, so ⌘K moves back to the Dvorak K position, which is the one thing those users picked the layout to avoid. `charactersIgnoringModifiers` honoured it before.

The `0` is right for the glyph path (`KeyShortcut.keyGlyph` wants the base character) and wrong for the chord path, so the two want different modifier state:

```swift
@MainActor static func character(for keyCode: Int, modifiers: UInt32 = 0) -> String? { … }

// in keyEquivalent, when the event carries Command:
character(for: Int(event.keyCode), modifiers: UInt32(cmdKey >> 8))
```

(`UCKeyTranslate` wants the Carbon flags shifted right by 8.)

Separately, the move dropped the `// no modifiers: the glyph is the key's base character` line that used to sit on this argument. Worth keeping — it is the only thing that explains the `0`.

Tinycast/Palette/PalettePanel.swift
```diff
@@ -64,7 +64,9 @@ final class PalettePanel: NSPanel {
         else { return nil }
         let arrow: (key: KeyEquivalent, code: Int)
         // Character chords, not key codes: Dvorak transposes the two.
```
abue-ammar: This comment now says the opposite of what the line under it does: the switch reads a key code, not a character. Dvorak still lands right because the ASCII-capable source *is* Dvorak, but the reasoning has changed and the comment should say why the key code is now safe.

```suggestion
        // Key codes through the ASCII-capable layout: an IME must not move ⌃N off its physical key.
```

Tinycast/Palette/PaletteWindowController.swift
```diff
@@ -267,7 +267,9 @@ final class PaletteWindowController: NSObject, NSWindowDelegate {
                 return true
             }
             // Character chords, not key codes: Dvorak transposes the two.
```
abue-ammar: Same stale comment as in `PalettePanel` — it describes the `charactersIgnoringModifiers` version that this PR replaced.

```suggestion
            // Key codes through the ASCII-capable layout: an IME must not move ⌘W off its key.
```

Tests/hotkey-test.swift
```diff
@@ -85,6 +86,14 @@ struct DoubleTapDetectorTests {
             "raw values are the persisted spelling and stay in canonical ⌃⌥⇧⌘ order")
     }
 
+    static func layoutCharacters() {
+        let character = ASCIIKeyboardLayout.character(for: kVK_ANSI_K)
+        expect(character?.isEmpty == false, "the ASCII-capable layout translates an ANSI letter key")
+        expect(
+            character?.unicodeScalars.allSatisfy(\.isASCII) == true,
+            "the shortcut character stays ASCII while a non-ASCII input source is active")
```
abue-ammar: **This can't fail on a US-layout runner**, which is every runner, so it does not guard the bug it was added for. The IME case is untestable from a harness — that is fine, but then the assertion should at least cover the whole key range the chords use rather than one letter, so a layout that translates letters but not punctuation is caught.

```suggestion
        let keyCodes = [kVK_ANSI_A, kVK_ANSI_K, kVK_ANSI_Z, kVK_ANSI_Comma, kVK_ANSI_Period]
        let characters = keyCodes.compactMap { ASCIIKeyboardLayout.character(for: $0) }
        expect(
            characters.count == keyCodes.count,
            "the ASCII-capable layout translates every ANSI key a palette chord uses")
        expect(
            characters.allSatisfy { $0.unicodeScalars.allSatisfy(\.isASCII) },
            "the shortcut character stays ASCII while a non-ASCII input source is active")
```

Tinycast/Palette/RootPaletteView.swift
```diff
@@ -361,8 +361,10 @@ struct RootPaletteView: View {
                 screen: menuOpen ? nil : screen as? ExtensionCommandScreen, selection: sel)
         )
         // ⌘K toggles the actions panel for the current selection.
-        .onKeyPress(keys: ["k"], phases: .down) { press in
-            guard press.modifiers.contains(.command) else { return .ignored }
+        .onKeyPress(phases: .down) { press in
```
abue-ammar, line 364: Not a defect I can point at, but the one change here I'd want confirmed on a device before merge: dropping `keys:` turns six handlers into unfiltered ones, so every keystroke in the palette now walks all six before reaching the field editor.

The guards check modifiers first, so plain typing returns `.ignored` without touching TIS and costs nothing measurable. The question is dispatch, not cost — that unfiltered `onKeyPress` sits above a focused `TextField`, and it is worth typing a fast query and confirming no character is eaten or reordered.

While you're there: a ⌘ chord now queries `TISCopyCurrentASCIICapableKeyboardLayoutInputSource` once per handler that reaches `matches`, so ⌘K costs about four copies. Deliberate per the PR description, and I agree caching a layout that can change mid-session is worse. Just noting it is four, not one.

abue-ammar (commented):
Read this on Linux, so no build or manual run behind it. The translation itself is a no-op on US QWERTY (the current ASCII-capable layout *is* US QWERTY, so key code to character is the identity), and the four panel chords, the six palette chords and the extension dispatch all resolve to the same characters they did before. Five notes below, one of which is a real regression that also hits QWERTY.

abue-ammar:
i fixed the issue it had. merging it

PR #412, opened 2026-08-31
Fix snippet expansion in web and Electron editors

### LINKED ISSUE #410: [Bug]: Snippet keyword expansion fails in some web/Electron editors
### Existing issues

### What happened?

Snippet keyword expansion works in some text fields, but fails in some web/Electron-style editors.

Observed working examples:

- Apple Books text fields
- Helium's browser chrome/search input

Observed failing examples:

- ChatGPT's conversation composer inside Helium
- Some VSCodium inputs, including editor-style/custom text surfaces

The keyword itself is detected, but the expansion is not applied in the failing editors.

This appears to depend on the type of text control rather than on a specific application.

### Steps to reproduce

1. Enable Snippets and configure a snippet with a keyword.
2. Type the keyword in an Apple Books text field or Helium's browser search input.
3. The snippet expands normally.
4. Type the same keyword in the ChatGPT conversation composer inside Helium, or in an affected VSCodium input.
5. The keyword remains and the snippet does not expand.

### Expected behavior

Snippet keyword expansion should work consistently in editable text controls that accept normal keyboard input, including web/Electron editors such as the ChatGPT composer and VSCodium editor-style inputs.

### Additional context

The failing controls are still visible through macOS Accessibility and report themselves as editable/focused. For example, the ChatGPT composer is exposed as an editable text area and reports writable text and selection attributes.

A likely timing issue is that Tinycast observes the final keyword keystroke before Chromium/Electron has propagated that keystroke through its editor/DOM model into the Accessibility text state. Tinycast may therefore validate the focused control while its Accessibility value is briefly one keystroke behind the event stream and reject the expansion as a mismatch.

This would explain why native/simple controls work while renderer-based editors such as ChatGPT and some VSCodium surfaces fail. It also means the earlier focus-restoration explanation for this issue was probably incorrect.

The important safety constraint remains that Tinycast should not blindly delete text when the focused control genuinely no longer contains the matched keyword. A fix should distinguish a short-lived Accessibility synchronization delay from a real text/caret mismatch.

### Screenshots or recording

_No response_

Current main (also observed after migrating from 0.10.9-beta.77 to Tinycast Dev)

### macOS version

27.0

### Measured diagnostics

A local diagnostic comparison reproduced the behavior on current main without changing the delivery logic.

- **Helium browser chrome / address field:** the full 10-character test keyword was already present in the Accessibility value and caret state at the first sample (about 0.3 ms after the listener observed the final key event). Normal Tinycast keyword expansion succeeded.
- **ChatGPT conversation composer in Helium:** at the first sample, Accessibility exposed only 9 of the 10 keyword characters; the full keyword appeared about 4.5 ms later. This confirms a short renderer-to-Accessibility propagation delay.
- **ChatGPT Accessibility write behavior:** after selecting the keyword, writing replacement text through the editable Accessibility text interface returned success, but the composer's actual value did not change at all. Tinycast can therefore receive a false-positive delivery result from this control.
- **VSCodium Monaco editor:** the real editor accepted the 10-character keyword and saved it to disk, while Accessibility continued to report an empty value and caret position 0 from the immediate sample through at least 320 ms. The focused Accessibility element also reports that the editor is unavailable unless screen-reader optimization is enabled. Normal Tinycast keyword expansion failed and left the keyword untouched.

These results show two distinct compatibility gaps in renderer-based editors:

1. some controls expose authoritative text state only a few milliseconds after the key event; and
2. some controls expose a writable-looking Accessibility surface whose state is not authoritative, or report a successful write without changing the underlying editor.

A fix should preserve Tinycast's fail-closed protection against deleting unrelated user text, while distinguishing transient Accessibility synchronization from a genuinely mismatched caret/text state. It should also verify that an Accessibility write actually changed the target before treating the delivery as complete, and use the existing keyboard-event fallback only when the focused target can be established safely.

chiimagnus:
## Related issue

Closes #410

## What changed

- Distinguishes a briefly stale Accessibility snapshot from a genuine keyword mismatch instead of rejecting both immediately.
- Gives automatic delivery a bounded 35 ms convergence window for renderer Accessibility state that is still catching up with the key stream.
- Treats text-marker web/editor controls as event-delivery targets rather than trusting AX mutation setters that can report success without changing the editor model.
- Reads native AX replacements back before reporting delivery, so a false-success setter is not accepted as delivery.
- Stops using non-authoritative text-marker state to confirm long paste fallback delivery.
- Cancels an in-flight automatic delivery on the next real user keyboard or mouse activity; Tinycast's tagged synthetic events do not cancel their own fallback.
- Adds deterministic snippet harness coverage for renderer lag, genuine mismatch, non-authoritative empty AX state, AX false-success read-back, and stale automatic-delivery cancellation.
- Updates the Snippets delivery contract documentation.

## Memory footprint

Measured on a clean temporary channel using the Release build and macOS `footprint` (`phys_footprint`). The temporary channel contained one snippet and no migrated user data.

| Idle after launch       | 40 MB  |
| Palette open            | 43 MB  |
| Feature in use (peak)   | 45 MB  |
| Palette closed, settled | 43 MB  |

Leak-tested: yes. After the initial one-time warm-up, four consecutive additional palette open/close cycles all settled at 43 MB; the post-close floor did not continue to rise.

## Drawbacks

Automatic expansion can spend up to 35 ms waiting for renderer Accessibility text to converge before using the existing event fallback. A readable state that already contains enough text but disagrees with the captured keyword still rejects immediately rather than guessing.

Text-marker editors deliberately use keyboard-event delivery because their AX mutation channel is not authoritative enough to confirm replacement safely. The fallback is less atomic than a native AX replacement, but it keeps the existing permission, Secure Event Input, target-app and cancellation gates, and now aborts when the user produces new real input before delivery completes.

## Tests & validation

- `./Scripts/run-tests.sh` — all 52 harnesses pass.
- `./Scripts/run-tests.sh snippets-test` — passes after the final safety change.
- `./Scripts/lint.sh` — `lint-clean`; only repository-baseline warnings remain.
- Pure-layer check: no AppKit/SwiftUI/Cocoa imports under `Features/*/Model/`.
- `git diff --check` — clean.
- Clean unsigned Debug build — succeeds with zero `warning:` / `error:` lines in the build log.
- Release build used for the memory sweep — succeeds with zero `warning:` / `error:` lines.
- Deterministic coverage for pending → matched convergence, genuine mismatch rejection, non-authoritative empty AX state, false-success AX writes, and real-vs-synthetic input cancellation.
- Manual: snippet keyword expansion succeeds in the ChatGPT composer in Helium.
- Manual: snippet keyword expansion succeeds in the affected VSCodium editor surface.
- Manual: ordinary/simple Helium text input continues to expand correctly.

### DIFF
--- a/Tests/snippets-test.swift
+++ b/Tests/snippets-test.swift
@@ -1573,8 +1573,9 @@ struct SnippetsTests {
             now: { Date(timeIntervalSince1970: 1_000) },
             syntheticEventTag: 123,
             logsTapFailures: false)
+        var activityCount = 0
 
-        listener.start { _, _, _, _ in }
+        listener.start(onUserActivity: { activityCount += 1 }, onMatch: { _, _, _, _ in })
         check(
             "real listener waits without permissions and does not install",
             listener.status == .needsAccessibility && tap.installCount == 0)
@@ -1593,7 +1594,25 @@ struct SnippetsTests {
                 && tap.installCount == 2
                 && tap.state == .active)
 
-        listener.start { _, _, _, _ in }
+        listener.processEvent(
+            typeRaw: CGEventType.keyDown.rawValue,
+            keyCode: 0,
+            flagsRaw: 0,
+            text: "x",
+            eventUserData: 0,
+            secureEventInputEnabled: false)
+        listener.processEvent(
+            typeRaw: CGEventType.keyDown.rawValue,
+            keyCode: 0,
+            flagsRaw: 0,
+            text: "x",
+            eventUserData: 123,
+            secureEventInputEnabled: false)
+        check(
+            "real user input invalidates pending automatic delivery while Tinycast events do not",
+            activityCount == 1)
+
+        listener.start(onUserActivity: { activityCount += 1 }, onMatch: { _, _, _, _ in })
         check(
             "real listener repeated start does not install a second tap",
             listener.status == .active && tap.installCount == 2)
@@ -1632,7 +1651,7 @@ struct SnippetsTests {
         check(
             "real listener stop is authoritative",
             listener.status == .off && tap.state == .absent)
-        listener.start { _, _, _, _ in }
+        listener.start(onUserActivity: { activityCount += 1 }, onMatch: { _, _, _, _ in })
         listener.stop()
         check(
             "real listener rapid on and off leaves no tap",
--- a/Tests/snippets-test.swift
+++ b/Tests/snippets-test.swift
@@ -636,6 +636,44 @@ struct SnippetsTests {
             "a rejected AX keyword replacement fails closed instead of deleting by events",
             !AccessibilityReplacement.rejected.fallsBackToEvents)
         check(
+            "an AX keyword one character behind the event stream remains pending",
+            AccessibilityReplacementPolicy.keywordState(
+                value: "!tcaxprob",
+                selectedRange: NSRange(location: 9, length: 0),
+                keyword: "!tcaxprobe") == .pending)
+        check(
+            "a converged AX keyword resolves to its exact replacement range",
+            AccessibilityReplacementPolicy.keywordState(
+                value: "prefix !tcaxprobe",
+                selectedRange: NSRange(location: 17, length: 0),
+                keyword: "!tcaxprobe") == .matched(NSRange(location: 7, length: 10)))
+        check(
+            "an AX state with enough text but the wrong suffix is a genuine rejection",
+            AccessibilityReplacementPolicy.keywordState(
+                value: "prefix !tcaxwrong",
+                selectedRange: NSRange(location: 17, length: 0),
+                keyword: "!tcaxprobe") == .rejected)
+        check(
+            "an empty editor AX snapshot remains pending instead of becoming a false mismatch",
+            AccessibilityReplacementPolicy.keywordState(
+                value: "",
+                selectedRange: NSRange(location: 0, length: 0),
+                keyword: "!tcaxprobe") == .pending)
+        check(
+            "AX replacement confirmation requires the observable text to actually change",
+            AccessibilityReplacementPolicy.confirmsReplacement(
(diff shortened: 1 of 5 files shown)


abue-ammar:
Superseded by #422, now merged. Thanks @chiimagnus — both of your mechanisms shipped, with `Co-authored-by`: the `AXSelectedTextMarkerRange` gate, the bounded keyword-convergence window, the AX write read-back, and the cancellation on real user input.

Folded in rather than merged alongside because #420 and this PR both rewrote the same tier decision in `replaceUsingAccessibility`; reconciling them separately would have left two overlapping verification paths with no way to tell which was load-bearing. Two things changed on the way in:

- The read-back returns `.unavailable` when the value is unchanged (events are safe) and `.rejected` only when it moved into something we did not write — a plain `.rejected` was a silent dead end with no fallback.
- The event tier had to be fixed first, in #420. Blink caps one key event's text at four UTF-16 units, so routing renderer editors to events without chunking would have turned "keyword stays put" into "keyword deleted, `pers` typed".

Closing this one as merged-by-proxy, not rejected.

chiimagnus:
Small attribution issue: the `Co-authored-by` trailer in #422 currently uses `chiimagnus@gmail.com`, which GitHub does not associate with my account.

`chii_magnus@outlook.com` is associated with `@chiimagnus` and can be used instead. My original commits in #412 also use the GitHub noreply address `151696847+chiimagnus@users.noreply.github.com`, which would work as well.

abue-ammar:
> Small attribution issue: the `Co-authored-by` trailer in #422 currently uses `chiimagnus@gmail.com`, which GitHub does not associate with my account.
> 
> `chii_magnus@outlook.com` is associated with `@chiimagnus` and can be used instead. My original commits in #412 also use the GitHub noreply address `151696847+chiimagnus@users.noreply.github.com`, which would work as well.

got it

PR #418, opened 2026-08-31
Honour an extension's popToRootType, so a finished command returns to the root

### LINKED ISSUE #404: [Bug]: popToRootType is ignored, so a finished command is restored instead of the root
### Existing issues

### What happened?

With Pop to Root Search set to a delay, an extension command that finishes its work and closes the window is restored instead of the root when I summon Tinycast again.

Picking a project in Zed Recent Projects opens the project in Zed, but re-summoning inside the delay puts me back in the project list rather than at a fresh root search.

The cause is in the host, not the extension. `closeMainWindow` and `showHUD` accept `{ clearRootSearch, popToRootType }` per the [Raycast API](https://developers.raycast.com/api-reference/window-and-search-bar#closemainwindow).

`ExtensionHostBridge` reads only `clearRootSearch`, and `ExtensionManager.closeMainWindow` then drops that too, so every close takes the `Default` path and honours the delay. `PopToRootType.Immediate` and `Suspended` are unreachable — an extension has no way to express either.

Zed Recent Projects calls bare `closeMainWindow()`, so it gets `Default`. Bitwarden's `create-login`, `create-send` and `search-sends` do pass `popToRootType: Immediate` to `showHUD`, and it is discarded.

### Steps to reproduce

1. Settings → General → Pop to Root Search → After 90 seconds
2. Install the Zed Recent Projects extension and open Search Recent Projects
3. Select a project — Zed opens and the palette closes
4. Summon Tinycast again within 90 seconds
5. The project list is restored. Expected: a fresh root search

0.1.0 dev

### macOS version

27.0

list-jonas:
## Related issue

Closes #404

## What changed

`closeMainWindow` and `showHUD` accept `clearRootSearch` and `popToRootType`
([Window & Search Bar](https://developers.raycast.com/api-reference/window-and-search-bar)), but the
host parsed neither past `clearRootSearch` and then dropped that too. Every close took the `Default`
path, so with Pop to Root Search on a delay, re-summoning restored the command instead of the root —
picking a project in Zed Recent Projects left you back in the list.

The option is now read in `ExtensionHostBridge` and passed inward as a `PaletteReset`, and the
pending reset becomes four states instead of an optional `Timer`. One deliberate divergence:
a command that closes its own window resolves `Default` to `Immediate`, since a finished command
leaves the delay no screen to hold. Zed calls bare `closeMainWindow()`, so wiring the option alone
did not fix it.

## Drawbacks

`clearRootSearch: false` cannot survive a reset — `prepare(mode:)` clears the query either way.

## Tests & validation

All 52 harnesses pass, Debug builds with no new warnings, `lint.sh` clean. `ext-test` covers the
wire mapping, the divergence and the HUD ownership rule.

Exercised against installed extensions: Zed Recent Projects (the reported symptom), Bitwarden (the
three commands passing `popToRootType` to `showHUD`), and TinyPNG/Translate for the no-view HUD path.


abue-ammar:
after reviewing this PR what i understood is 
Before: the user's Pop to Root Search delay decided everything. 
After: extension JS decides, in both directions.

tinycast wont depend on 3rd party. if you have any first party solution feel free to create a new pr. i am closing it for this reason

Issue #431, opened 2026-09-01
Add window workspaces


1tuz:
## Problem

A work setup often uses the same applications in the same positions, but Tinycast currently moves one window at a time. Rebuilding a multi-window layout after a restart, display change, or context switch requires repeating those commands.

## Example

A user saves a `Writing` workspace with a text editor on the left and a reference app on the right. One launcher command or hotkey restores that arrangement.

## Proposal

Window Workspaces would save a named snapshot of eligible Accessibility windows and restore it only when the user explicitly runs the workspace. Frames would be stored relative to each display's visible area so the layout can adapt to resolution changes. Missing applications may be launched once during that explicit restore.

## Boundaries

This would not watch windows, poll displays, run timers, or create a background automation service. It would reuse Tinycast's existing launcher, hotkey, Accessibility, and window-management paths.

Would this be a good fit for Tinycast?

abue-ammar:
related to #433

PR #432, opened 2026-09-01
Add window workspaces (#431)

1tuz:
## Summary
- save named Accessibility-window layouts by application and display
- restore the Settings-selected layout through Launcher or a global shortcut
- keep frames resolution-independent by storing visible-frame fractions; launch missing apps once when available

## Idle impact
No polling, timers, observers, dependencies, Electron, or JavaScript runtime. Accessibility work occurs only on explicit Save or Restore gestures.

## Verification
- workspace and extension harnesses pass
- `./Scripts/lint.sh`
- `./Scripts/format.sh --check`
- Debug `xcodebuild` succeeds

A parallel full-suite run hit a pre-existing flaky `ext-test`; its isolated rerun passed.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -185,6 +185,7 @@ run volume-test            Tinycast/Features/SystemActions/Model/VolumeLevel.swi
 run window-command-test    Tinycast/Features/WindowManagement/WindowCommand.swift \
                            Tinycast/Features/WindowManagement/WindowLayout.swift \
                            Tinycast/Features/WindowManagement/WindowActionMemory.swift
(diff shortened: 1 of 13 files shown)


abue-ammar:
@1tuz hey kindly open issue i would like to know weather it should belongto tinycastor no. with why do we need it and attaching demo much preferred

abue-ammar:
after consideration i am sorry to say i do not think i would want it in tinycast. i will close the pr but in future if lots of people demands it a lot maybe i will add it thanks a lot. if it's something really important u have already built it so you can have it in your fork.

Issue #256, opened 2026-08-18
[Bug]: Applications that are in subfolders of /Applications are not shown


vexcited:
### Existing issues

- [x] I searched open and closed issues and didn't find this already reported

### What happened?

I have those installed and I can't find them.

<img width="357" height="115" alt="Image" src="https://github.com/user-attachments/assets/4f07288b-d164-4f3a-9751-092145fb27b0" />

I suspect it's because inside a subfolder in /Applications, the same issue happens with DaVinci Resolve notably.

### Steps to reproduce

1. Have an app in a subfolder of /Applications
2. Try to search that app in Tinycast
3. Woops, nothing...

### Screenshots or recording

<img width="753" height="471" alt="Image" src="https://github.com/user-attachments/assets/147a31a6-3aed-4af8-a022-9bd097932981" />

<img width="752" height="485" alt="Image" src="https://github.com/user-attachments/assets/c2cbc73e-09b9-4731-a3a7-ece1a4ff36ef" />

### Tinycast version + channel

0.9.5 stable

### macOS version

27.0 beta 6

vexcited:
NOTE: I know I can add those folders to lookup Applications in the app settings but nobody wants to add all those subfolders manually one by one for each apps, it works by default on any other launcher and should probably be the expected behavior.

abue-ammar:
> NOTE: I know I can add those folders to lookup Applications in the app settings but nobody wants to add all those subfolders manually one by one for each apps, it works by default on any other launcher and should probably be the expected behavior.

yeah will fix it might go one layer more to look

PR #458, opened 2026-09-04
Descend through symlinked folders when scanning application scopes

shelken:
## Related issue

Follow-up to #256 (PR #268)

## What changed

PR #268 enabled one-level-deep descent to discover apps located inside vendor subfolders. However, if a subfolder or scope is a symbolic link, `FileManager.default.contentsOfDirectory` throws POSIX error 20 (`ENOTDIR`) and `item.resourceValues.isDirectory` returns `false`.

- **First choice (fast path)**: Retains batched attribute prefetching (`includingPropertiesForKeys: [.isDirectoryKey, .isSymbolicLinkKey]`). Normal directories and bundles follow the original path with zero extra stat calls.
- **Fallback**: When `contentsOfDirectory(at: url)` fails with `ENOTDIR` (e.g. `url` is a symlinked directory), it falls back to `url.resolvingSymlinksInPath()`. During enumeration, symlinks are checked to see if their destination is a directory.
- **Logical paths**: Reconstructs items using `url.appendingPathComponent(item.lastPathComponent)` so indexed app paths remain stable under their configured scope rather than leaking transient physical targets.
- Prunes broken `.app` symlinks so dead targets are not indexed.

## Memory footprint

| Step                    | Memory       |
| ----------------------- | ------------ |
| Idle after launch       | not measured |
| Palette open            | not measured |
| Feature in use (peak)   | not measured |
| Palette closed, settled | not measured |

Not profiled. The change only performs a bounded directory traversal returning temporary `[URL]` arrays dropped immediately after enumeration. No cache, no long-lived state, and no background tasks.

Leak-tested: no — allocations are local to the function call.

## Drawbacks

None. Bounded depth of 1 and leaf treatment of `.app` bundles remain strictly unchanged.

## Tests & validation

- Extended `Tests/scopes-test.swift` with assertions for symlinked directories, logical path preservation, and broken symlink pruning.
- Passed `./Scripts/run-tests.sh scopes-test`.
- Passed `./Scripts/run-tests.sh` (all 54 harnesses pass).
- Passed `./Scripts/lint.sh`.
- Verified on macOS that `SearchScopes.appBundles(in: ["~/Applications"])` discovers apps under symlinked subfolders with logical paths preserved.

### DIFF
--- a/Tests/scopes-test.swift
+++ b/Tests/scopes-test.swift
@@ -2,7 +2,7 @@ import Foundation
 
 @main
 struct ScopesTest {
-    static func main() {
+    static func main() throws {
         let fm = FileManager.default
         let root = fm.temporaryDirectory
             .appendingPathComponent("tinycast-scopes-\(UUID().uuidString)")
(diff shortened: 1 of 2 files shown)


abue-ammar:
there will always be these nix user i have to make exceptions for. how about adding the nix folder location application path to the scope? i don't want to put this in tinycast.
and why it's not in fallback put it in fallback why is it in the first choice.

shelken:
Thanks for taking a look and sharing your thoughts!

To clarify the design decisions and why manually adding the path doesn't work well:

1. **Why manual addition isn't viable for Nix / Home Manager:**
   - **Dynamic hash paths**: Nix packages live in `/nix/store/<hash>-.../Applications`. Every time a user runs a rebuild or garbage collection, the old hash path is deleted and replaced with a new one. A manually configured store path would permanently break on the next update with an invalid path warning.
   - **macOS `NSOpenPanel` resolves symlinks**: When a user picks `~/Applications/Home Manager Apps` in the file picker, macOS out-of-process file panel automatically resolves symlinks and returns the raw `/nix/store/...` physical path. Tinycast would store that ephemeral path rather than the stable symlink.
   - Supporting manual addition cleanly would actually require significantly more code (~50+ lines) to reverse-engineer and map physical paths back to symlinks in the UI.

2. **No Nix-specific exceptions or hardcoded logic:**
   - This PR contains **zero** Nix-specific code or path checks. It is purely generic POSIX symlink handling for directory enumeration.
   - It completes the contract originally established in #256 (PR #268): descending one subfolder deep to find applications. The existing implementation used `isDirectoryKey` (which returns `false` for symlinked directories) and called `contentsOfDirectory` directly on the link (which throws POSIX error 20 `ENOTDIR`). This PR simply allows that same 1-level descent to work transparently on symlinked directories.

3. **Regarding "Fallback":**
   - In Tinycast, `Fallback` refers to post-search actions (AI chat, file search, web search) when a query yields no app matches.
   - Application discovery, on the other hand, is a one-pass background scan (`AppIndex.scan`). When enumerating a folder, inspecting whether a subfolder is a directory or a directory symlink is handled directly during that single traversal, rather than a separate query-time fallback.

In total, this adds only ~16 lines of core logic to `SearchScopes.swift` while keeping the strict 1-level descent limit and `.app` leaf boundary completely intact. It keeps things simple and clean with zero UI overhead.

Co-Authored-By: gemini-3.8-flash-high

shelken:
Updated the PR with a commit separating the fast path from the symlink fallback:

1. **Fast path**:
   - Retains batched attribute prefetching (`includingPropertiesForKeys: [.isDirectoryKey, .isSymbolicLinkKey]`).
   - Regular directories and bundles follow the original path with zero extra stat calls.

2. **Fallback**:
   - `contentsOfDirectory(at: url)` is called first. When it fails with `ENOTDIR` (url is a symlinked directory), it falls back via `??` to `url.resolvingSymlinksInPath()`.
   - During iteration, destination checks only run if `values?.isSymbolicLink == true`.

3. **Manual scope addition in existing code**:
   - `contentsOfDirectory(at: url)` in the current main branch throws `POSIX error 20 (ENOTDIR)` when pointed directly at a symlinked directory. Manually adding the path to scopes currently yields zero results.
   - The fallback resolves this for both manually added scopes and 1-level subfolder discovery, while keeping logical paths stable.

abue-ammar:
performance degraded i dont want it in tinycast. if it's a deal breaker for you, you can have it in your fork. sorry to say

image

PR #484, opened 2026-09-07
Treat a lone x between operands as multiplication

qeude:
## Related issue

Closes #

## What changed

- `CalcTokenizer` folds a lone `x` / `X` between operands into `*` (`3x3`, `3 x 3`, `2xpi`, `$5 x 2`), so the scalar and typed-quantity paths agree with no parser change. It only fires at a token boundary with a value on the left, which is why `x`, `x3`, `max`, `hex` and a half-typed `0x` prefix stay as they were.
- `docs/signing.md`: two-step identity creation (`req -new` + `x509 -req`). The one-shot `req -x509` prepends a duplicate `CA:true` basicConstraints on OpenSSL 1.1, producing an identity `codesign` cannot use. Verification is now a trial sign instead of `find-identity`.

## Memory footprint

No new state — a tokenizer-only change plus docs.

| Idle after launch       | —      |
| Palette open            | —      |
| Feature in use (peak)   | —      |
| Palette closed, settled | —      |

Leak-tested: n/a — no new long-lived objects.

## Drawbacks

- The card echoes what was typed (`3 x 3`, not `3 × 3`).
- An attached trailing `3x` stays silent until the next operand arrives, which is what keeps a half-typed hex prefix card-free.

## Tests & validation

- `calc-test`: new cases for attached / spaced / uppercase `x`, a unary right side, `2xpi`, `6/2x(1+2)`, the trailing partial (`10 x` → `10`), quantity and currency (`2(3)kg x 2`, `$5 x 2`), and nil guards (`x`, `x3`, `3x`).
- Full suite: 58/58 harnesses pass; `lint.sh` is clean.
- By hand: the new two-step identity trial-signed a scratch binary, and a Debug `Tinycast Dev` build signed with it succeeded.

### DIFF
--- a/Tests/calc-test.swift
+++ b/Tests/calc-test.swift
@@ -74,6 +74,19 @@ struct CalcTests {
         expectDisplay("e^2", "7.389056099")
 
         // Implicit multiplication
+        // `x` reads as `*` between operands, attached or spaced, like Raycast
+        expectDisplay("3x3", "9")
+        expectDisplay("3 x 3", "9")
+        expectDisplay("3X3", "9")
+        expectDisplay("3 x -2", "-6")
+        expectDisplay("2xpi", "6.283185307")
+        expectDisplay("6/2x(1+2)", "9")
+        expectDisplay("10 x", "10")  // trailing operator keeps the prefix while typing
+        expectDisplay("$5 x 2", "10.00 USD")
+        expectDisplay("2(3)kg x 2", "12 kg")
+        expectNil("x")
+        expectNil("x3")
+        expectNil("3x")  // half-typed, not an operator yet
         expectDisplay("4(2+3)", "20")
         expectDisplay("(2+3)(2+3)", "25")
         expectDisplay("2pi", "6.283185307")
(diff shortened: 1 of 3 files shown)


abue-ammar:
Closing in favour of #483, which already lands this — same approach, same behaviour.

#483 folds a lone `x` / `X` into `*` in the tokenizer rather than the parser, for the reason given here (so the scalar and typed-quantity paths agree with no parser change), and it carries all twelve assertions from this PR verbatim:

`3x3`, `3 x 3`, `3X3`, `3 x -2`, `2xpi`, `6/2x(1+2)`, `10 x`, `$5 x 2`, `2(3)kg x 2`, and the `x` / `x3` / `3x` nil guards — plus the `docs/features/calculator.md` paragraph.

I ran every one of those against #483's branch and the answers match this PR exactly, including the guards that keep `max`, `hex` and a half-typed `0x` prefix card-free.

Two notes on why they aren't merged together:

- This PR edits `CalcParser.swift`, which #483 deletes when it splits that file into `CalcTokenizer.swift` and `CalcExpressionParser.swift`. It is already showing as conflicting with `main`.
- #483 also introduces `xor` as a bitwise operator, so its version of the check carries one extra guard so that `5 xor 3` is not split into `5 * or 3`. That guard isn't needed on this branch, which has no `xor` — just noting why the two spellings of the check differ.

Thanks @qeude — the `x`-as-multiply behaviour ships either way.

Issue #486, opened 2026-09-07
[Feature]: Focus an extension command's first argument after Space


xeric:
### What should it do?

When a launcher result is a Raycast extension command with declared arguments, typing the command's exact title and then pressing Space should move focus directly from the search field to the command's first inline argument.

Partial matches and ordinary edits should continue filtering. Commands without arguments should keep the existing search behavior. The title comparison can remain case- and diacritic-insensitive, matching launcher search.

This matches Raycast's keyboard flow and reuses Tinycast's existing inline argument fields, focus state and submission path; it adds no new surface or persistent state.

### Why does it belong in Tinycast?

Tinycast already renders extension command arguments inline, but reaching them currently requires Tab after finding the command. Raycast users expect the trailing Space after an exact command title to hand input to the first argument, so commands such as search, translate and status lookup can be run as one continuous typed flow.

The implementation is narrow: detect one appended Space for the selected extension command, verify an exact title match, and focus its first declared argument. Model coverage pins exact, case-insensitive, partial-title, non-Space and no-argument behavior.

abue-ammar:
i dont want to implement it or have this in tinycast. sorry reason raycast is a 3rd party it stays 3rd party and i dont want raycast extension to tightly coupled with tinycast.

PR #491, opened 2026-09-07
Focus extension command arguments after Space

xeric:
## Summary

- Focus the first declared argument when a selected extension command's exact title is followed by Space.
- Keep ordinary launcher filtering unchanged for partial titles, non-Space edits, and commands without arguments.
- Reuse the existing inline argument field and submission flow; no new persistent state or UI surface is introduced.

## Motivation

Raycast extension commands frequently require an argument. After locating an exact command, users currently need to press Tab before typing that argument. This change allows one continuous keyboard flow: type the command title, press Space, then enter the first argument.

## Implementation

- Add a pure matcher for the exact-title plus trailing-Space transition.
- Resolve the selected extension command's first argument in `LauncherScreen`.
- Route that focus target through the existing palette query-change path.
- Cover exact, case-insensitive, partial-title, non-Space, and no-argument behavior in the standalone harness.

## Verification

- `./Scripts/run-tests.sh`: 58 harnesses passed.
- `./Scripts/lint.sh`: passed.
- Debug build: succeeded with no new warnings.
- Model-layer import purity check: passed.
- Manual Dev-build smoke test: entering `Ask AI`, pressing Space, and typing `hello` kept the launcher query at `Ask AI` and placed `hello` in the inline `Question` field.
- Physical footprint remained below the documented ceiling: 50 MB idle, 54 MB while exercising the flow, and 53 MB after closing the palette.

Related to #486.

### DIFF
--- a/Tests/ext-test.swift
+++ b/Tests/ext-test.swift
@@ -250,6 +250,26 @@ struct ExtensionTests {
         check(
             "provided arguments survive completion",
             manifest.commands[3].completeArguments(["q": "hi"]) == ["q": "hi"])
+        check(
+            "Space after an exact command title focuses its first argument",
+            manifest.commands[3].firstArgumentNameAfterSpace(
+                previousQuery: "Args", newQuery: "Args ") == "q")
+        check(
+            "the exact command title comparison ignores case",
+            manifest.commands[3].firstArgumentNameAfterSpace(
+                previousQuery: "args", newQuery: "args ") == "q")
+        check(
+            "a partial command title keeps filtering",
+            manifest.commands[3].firstArgumentNameAfterSpace(
+                previousQuery: "Arg", newQuery: "Arg ") == nil)
+        check(
+            "ordinary edits do not steal argument focus",
+            manifest.commands[3].firstArgumentNameAfterSpace(
+                previousQuery: "Args", newQuery: "Args!") == nil)
+        check(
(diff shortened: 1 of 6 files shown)


abue-ammar:
@xeric put demo video

abue-ammar:
it breaks in a terrible way . how am i supposed to access search file command? if you want space to access ? i dont want this in tinycast. if it is really a deal breaker. you can have this in your fork. thanks

https://github.com/user-attachments/assets/841ef580-4f4b-425c-bb97-514f00d84d00

xeric:
Hi @abue-ammar, yes, you are right, I did not consider some use cases that really need space. Do you think keeping it working as using the "tab" key is okay? Raycast is using the tap to navigate to the input box of argument.
You can have a try again, with this change it has two ways to focus argument input, space and tab, if you think it is ok to use tab, I can raise another PR.

sermelipharo:
Implement it in my fork, been testing it for a few days, feels great, also fixed the collisions when alias is matching with same non alias command. All kudos to Abue Ammar, the fork is completely depends on the main rep and syncs automatically. 

[✨ Tinycast Space Edition](https://github.com/sermelipharo/tinycast-space#alias--space)

abue-ammar:
@sermelipharo you can keep it in your fork and use it the way you like. i am not putting it. space will behave space. space will not behave enter or tab

PR #495, opened 2026-09-07
Background refresh for no-view extension commands

qeude:
## Related issue

Closes #

## What changed

- `no-view` commands declaring `interval` (`90s` / `1m` / `12h` / `1d`, clamped to 1 min) re-run headlessly on schedule with `launchType: Background`.
- `updateCommandMetadata({ subtitle })` implemented (was a no-op): per-command override persisted in `extension-data`, shown in launcher search; `null` clears back to the manifest; a subtitle restating the extension name is dropped since the row already shows it.
- Launcher row: antenna dot while refresh is on (slashed twin when off), orange warning with the error as tooltip on background failure; the ⌘K menu gains Enable / Disable Background Refresh plus Refresh Now.
- Settings › Extensions › command: Background refresh toggle with last refresh + last error.
- Scheduler: a single loop, batched close ticks, deterministic per-command phase, exponential backoff to 24 h, timeout inside the interval, never preempts a running command (a manual run aborts a tick in flight), no toasts / HUDs / alerts from background runs.
- New `ext-refresh-test` harness; `ExtensionRefreshPolicy` holds every pure decision (parsing, due dates, backoff, subtitle fallback, indicator state).
- Tolerant `extension-data` decoding (new `ext-storage-test`): files predating a key keep their data instead of resetting the extension — this also fixes upgrades, where the pre-`metadata` files would otherwise have been wiped on first launch.

## Memory footprint

Verified with `footprint` (Activity Monitor's metric), same protocol on both sides.

| Profile | before (main) | after (PR) |
| ------- | ------------- | ---------- |
| Normal user, extensions off | 30 MB | 30 MB |
| Power user, extensions on + refresh ticking | 30 MB | 32 MB |

Nothing for normal users; ~2 MB transient during a tick, discarded with it.

## Demo

Background refresh in launcher search

*Caffeinate Status showing its live `✖ Decaffeinated` subtitle plus the refresh dot. After-state still; no before clip (headless test run).*

## Drawbacks

- `menu-bar` intervals parse but never schedule (menu-bar commands don't run).
- A tick never preempts a running command; it waits for the next due slot.
- Minimum interval clamped to 60 s (Raycast documents 10 s in one place, 1 m in the manifest reference).
- No Extension Diagnostics equivalent; background errors surface in Settings and the row tooltip.

## Tests & validation

- New `Tests/ext-refresh-test.swift` (interval parsing, due dates, backoff, jitter, timeout, subtitle fallback/dedup, indicator state, `LaunchType` raw values); full suite 60/60 green, `lint.sh` clean, Debug build with zero new warnings, `Model/` purity grep empty.
- Live E2E on the Dev channel with the real Coffee extension: the `1m` manifest interval parsed, background ticks rewrote the subtitle (`✔ Caffeinated` / `✖ Decaffeinated`), dot + warning states appear in the launcher row (see Demo). Genericness proven with a synthetic second extension (counter + subtitle advancing across ticks, then removed).


abue-ammar:
hey i got some questions and discussion 
1. why do we need it? 
2. what problem you had without it and what issue it solves?
3. I am not happy with the ram usage at all and looks like you put a background job in the main app entry point and it is running in background and ram usage is high. 
4. i Strictly do not want raycast extension to to tightly coupled with Tinycast that's i have have separated everything no dependency even popovers are different 
5. if it's really just 1 extension related issue eg: caffeine. this don't belong to Tinycast and it's more of fixing personal issue. 
6. for extension if it works that's it i don't want major Polish that degrades Tinycast and make it a bloated.

qeude:
Thanks for these feedback

1, 2, 4, 5 I've been using Tinycast as a replacement of Raycast and was just expecting extensions to work the same way. While I understand the idea of not being coupled tightly with Raycast, I think it's important to have feature parity and benefit from the huge Raycast extensions store the same way we used to from Raycast.
In my case it's just caffeine, but since this is a documented thing in Raycast, I think there are multiple of these across extensions, so it can be a problem for any user of Tinycast to have such a no-op command while it used to work in Raycast.
3. When building the app with the same Debug profile before and after on my side, I don't see any difference in RAM usage. The background task is owned by `ExtensionManager` and not the entry point

Don't hesitate if you have any questions

abue-ammar:
@qeude Thanks for clarifying. Do you know how many other extensions have this kind of behavior, and whether these changes should work fine for them too?

Since Tinycast mainly targets normal users, most users will have extensions off and auto-refresh off by default. It’d be great to check the RAM impact for:

1. **Normal user:** Extensions off, basic app search + clipboard usage.
2. **Power user:** Extensions + auto-refresh on, using extensions normally.

Could you compare both cases against main vs your PR? Tinycast is extremely strict about RAM usage, so I want to make sure normal users aren’t affected at all.

qeude:
@abue-ammar Sure!

There are 75 extensions using this pattern across the Raycast extensions repo

Verified with `footprint` (Activity Monitor's metric), same protocol on both sides.

| Profile | before (main) | after (PR) |
| ------- | ------------- | ---------- |
| Normal user, extensions off | 30 MB | 30 MB |
| Power user, extensions on + refresh ticking | 30 MB | 32 MB |

Nothing for normal users; ~2 MB transient during a tick, discarded with it.

PR #496, opened 2026-09-07
Make Escape walk back through the screens it opened

notsoshu:
Escape now walks back through the screens it opened, instead of every screen
having its own idea of "back".

## The four rules

Taking the four points as written:

1. Escape with text in the search field clears the field — unchanged.
2. Escape in a view opened directly by its own hotkey closes the palette — unchanged.
3. Escape in a view reached from root search goes back to the launcher — new.
4. ⌘Escape from any view goes to root search — this already worked, undocumented
   and untested; it now has both.

> **A quirk on ⌘⎋ worth a decision.** macOS binds ⌘Esc to Game Overlay — symbolic
> hotkey `260`, [documented by Apple][overlay] — and WindowServer consumes it before
> any application receives it. The handler in `PaletteWindowController` is correct,
> and I confirmed the chord end to end with a hardware press against a traced build,
> but only after switching that shortcut off. The quirk is how it fails: on a stock
> macOS ⌘⎋ does nothing at all here, with no beep and no error, and the keystroke
> never reaches the process — so it is invisible from inside the app and easy to
> mistake for a bug in the palette. This predates the PR; I raise it because this is
> the first change to document and surface the chord, so it may be worth rebinding
> to something macOS does not already claim.

[overlay]: https://support.apple.com/en-au/105118

## How it works

`PaletteState` gains a `PaletteNavigationStack` of frames, each holding the mode,
query, selection and clipboard filter of the screen it covers. Escape pops one
frame, and at the bottom of the stack popping means hiding — so rules 2 and 3 are
the same rule seen from different depths.

Where a screen sits is decided by the summon and nothing else, in one line of
`PaletteCoordinator.showPalette`: a hidden palette starts a new stack, a visible
one pushes. No `origin:` parameter threaded through the call sites. That means a
screen opened by its own hotkey is a root and closes, while the same screen opened
from a launcher row goes back to that row with the query still typed and the row
still selected.

A third motion sits beside push and pop: `replace(mode:)` swaps the current screen
and leaves the stack alone. Starting a new chat uses it, so the launcher that
opened the old conversation is still there to go back to; so does one extension
command fired over another's screen, since stacking those would leave Escape
restoring a frame whose command had already been stopped.

The three back edges converge: the header chevron, bare Backspace and Escape all
call `RootPaletteView.goBack()`. An extension screen still leaves through
`exitExtensionScreen`, which pops the extension's own navigation before the
palette's — and the chevron and Backspace now honour that too, where before only
Escape did.

The chevron is drawn only where that step leads somewhere. Its old condition, "any
mode but the launcher", used to mean the same thing as "has a screen underneath",
and with a stack it no longer does — on a hotkey-opened screen it would have become
a back arrow that dismissed the window. A screen with nothing under it now shows
its own icon, and an extension counts its own navigation depth, so a command with
sub-views keeps the chevron.

**The setting:** Escape Key Behavior, under General beside Pop to Root Search.
"Pop back or close" is the default and the behaviour above; "Close and pop to root"
hides and resets at once, leaving Backspace and the chevron as the way back. It is
searchable in Settings and rides a settings backup like `popToRootTimeout` does —
it grants no capability and names nothing machine-local, so it belongs in
`mirrored` rather than the exclusions, and the coverage test enforces that.

**The import:** an imported settings file carries `escapeKeyBehavior` in the same
`general` object as `popToRootTimeout`, so it arrives under that same option,
relabelled "Pop to root & Escape". Exact-match only — an unrecognised value is
skipped rather than clamped to the nearest, the same rule the timeout follows.

**The timeout still applies** under "Pop back or close": Pop to Root Search is
already a setting here, and having Escape quietly ignore it would break it. Under
"Close and pop to root" the reset is immediate, since that option's name promises
it, and that immediate reset honours the same authorizing exception the timer does,
so an extension waiting on a browser round trip still finds its screen when it
returns. Nothing new is bound to go back under that option — bare Backspace already
does exactly that.

## What changes for existing use

- **A chat opened by its own hotkey now closes on Escape** rather than falling back
  to the launcher. Chat loses its special case and joins the stack like every other
  screen. This is the one place existing muscle memory changes.
- **Tab stays lateral.** The mode ring replaces the current screen rather than
  descending, so Escape after a Tab behaves exactly as it did.
- **A different mode's hotkey pressed while the palette is open pushes rather than
  re-roots**, which is the one case deriving provenance from the summon gets wrong.
  Escape then returns to the screen you were looking at.

## Testing

`palette-escape-test` is rewritten against the new resolver and covers the whole
decision table — both behaviours × `canGoBack` × seven modes × empty/typed ×
menu-open. New `palette-navigation-test` covers push/pop/replace/reset round trips
and that going back restores the frame it stored, including that a replaced screen
never adds a second frame. The import harness gains the new value's mapping,
including that an unrecognised one is skipped.

All 59 harnesses pass, the Debug build has no new warnings, and swift-format and
swiftlint are clean.

Manually exercised on a Debug build: the launcher→clipboard→back round trip with
the query and row restored, hotkey-opened screens closing, chat and chat history
two deep, uninstall, the clipboard type filter and ⌘K menu each swallowing the
first Escape, ⌘⎋ from depth, Backspace under both settings, the chevron appearing
only over a real back step, and a settings import. Two sub-checks I could not drive
from the outside: an extension's inner sub-view popping before the palette's stack,
and Escape closing an extension form's dropdown without navigating — both are
pre-existing code paths this change does not touch.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -138,6 +138,7 @@ run ai-instructions-test   Tinycast/Features/AI/Model/AIInstructions.swift \
                            Tinycast/Features/AI/Model/AIPreamble.swift
 run hover-arming-test      Tinycast/Palette/HoverArming.swift \
                            Tinycast/Palette/PaletteState.swift \
+                           Tinycast/Palette/PaletteNavigationStack.swift \
                            Tinycast/Palette/PaletteMode.swift \
                            Tinycast/Features/Clipboard/Model/ClipboardStore.swift \
                            Tinycast/Features/Clipboard/Model/ClipboardFilter.swift \
@@ -149,6 +150,19 @@ run hover-arming-test      Tinycast/Palette/HoverArming.swift \
                            Tinycast/Features/CustomCommands/Model/CustomCommand.swift
 run palette-escape-test    Tinycast/Palette/PaletteMode.swift \
                            Tinycast/Palette/PaletteEscapeAction.swift \
+                           Tinycast/Features/Settings/EscapeKeyBehavior.swift \
+                           Tinycast/Features/Quicklinks/Model/Quicklink.swift \
+                           Tinycast/Features/Quicklinks/Model/QuicklinkDestination.swift \
+                           Tinycast/Features/CustomCommands/Model/CustomCommand.swift
+run palette-navigation-test Tinycast/Palette/PaletteNavigationStack.swift \
+                           Tinycast/Palette/PaletteState.swift \
+                           Tinycast/Palette/HoverArming.swift \
+                           Tinycast/Palette/PaletteMode.swift \
+                           Tinycast/Features/Clipboard/Model/ClipboardStore.swift \
+                           Tinycast/Features/Clipboard/Model/ClipboardFilter.swift \
+                           Tinycast/Features/Clipboard/Model/ColorValue.swift \
+                           Tinycast/Features/Clipboard/Model/ColorFormat.swift \
+                           Tinycast/Features/Clipboard/Model/ColorSpaces.swift \
                            Tinycast/Features/Quicklinks/Model/Quicklink.swift \
                            Tinycast/Features/Quicklinks/Model/QuicklinkDestination.swift \
                            Tinycast/Features/CustomCommands/Model/CustomCommand.swift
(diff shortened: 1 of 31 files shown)


abue-ammar:
it carried forward in #504 with commit credit and authorship

PR #504, opened 2026-09-08
Make Escape walk back through the screens it opened

abue-ammar:
## Make Escape walk back through the screens it opened

The palette had no memory of how it got to a screen. `PaletteState.mode` was a flat enum and every
transition went through `prepare(mode:)`, which resets everything — so Escape had to guess, with
hard-coded per-mode rules. Typing "Clipboard History" at the root, pressing ↵, then pressing Escape
closed the window instead of returning to the search that found it.

### The rule

**The summon decides provenance, not the mode.** `PaletteCoordinator.navigate(to:)` is the single
rule: a palette already on screen is being *navigated*, so the current screen is pushed and becomes
the step back; a hidden one is being *summoned*, so the new screen is a root with nothing behind it.
Every mode command and every global hotkey already funnels through `showPalette`, so this needs no
per-feature plumbing.

`PaletteState` holds the screens below `mode` as `[PaletteFrame]` — mode, query and selection, enough
that returning looks like never having left — with four motions over it:

| Motion | Meaning |
| --- | --- |
| `prepare(mode:)` | become the root: open fresh, drop the stack |
| `replace(mode:)` | swap the screen, keep what it was opened over (a new chat, not a new root) |
| `push(mode:)` | open over the current screen, which a back step returns to |
| `pop()` | restore the screen underneath; `false` when this one is the root |

`pop()` bumps `followToken` rather than `resetToken` — the reset token exists to snap a list to the
top, which would throw away the very selection being restored.

### What Escape does now

Clear a non-empty query → an extension screen exits itself (it keeps a stack the palette cannot see)
→ a pushed screen pops → a root hides the palette. A focused inline argument field is still a rung
above the query. Bare ⌫ in an empty field takes the same step, and all three back edges — Escape,
backspace, and the header chevron — now agree, where they previously disagreed.

### ⌘⎋ needs an event tap

⌘⎋ pops to the root search from any depth with the window still open. It cannot be done in the
responder chain: macOS binds the chord before any app sees it, so a handler there compiles, never
runs, and looks like a palette bug. `CommandEscapeTap` takes it at the head of the HID stream
instead — the one point earlier than the system's own binding.

The tap is enabled only while the palette is on screen, watches `keyDown` alone, and returns the
event untouched whenever the panel is not the key window, so nothing else on the system loses ⌘⎋ to
it. It is a modifying tap, so it rides on Accessibility — already required and prompted for
elsewhere in the app; no new permission class, and nothing here asks for Input Monitoring.

### The header chevron

Drawn on every screen but the launcher, so clipboard and emoji read the same as quicklinks and
snippets — a slot that changed shape with provenance would look like two different controls. Where
the click lands still follows the stack, and its tooltip says which rather than promising a step back
that is really a close. It lights to `textPrimary` under the pointer over a new `Theme.Duration.hover`,
with the hover state held inside `HeaderBackButton` so the header around it never re-renders.

### Setting

`Escape Key Behavior` in General, beside Pop to Root Search:

- **Navigate back or close window** (default) — the behaviour above.
- **Close window and pop to root** — an empty field closes the window and resets it immediately,
  whatever Pop to Root Search says.

Clearing the query is still the first press under either. Mirrored into settings backups.

### Behaviour changes worth calling out

- **Chat opened by its own global hotkey now closes on Escape** rather than falling back to the
  launcher — it is a root like any other directly summoned screen. Reached from the launcher, it
  goes back.
- **A global hotkey pressed while the palette is already open pushes**, so Escape returns to what was
  on screen. One rule rather than two, and it matches Raycast.

### Testing

- `palette-navigation-test` (new) — push/pop round-trips, the token discipline, `prepare` emptying
  the stack, `replace` preserving it.
- `palette-escape-test` — rewritten for the new decision table, plus the ⌘⎋ chord predicate
  (⌥⌘⎋ is Force Quit and must pass through; Caps Lock and fn riding along must not disqualify it).
- All 59 harnesses pass, Debug builds with no new warnings, `lint.sh` clean, and no AppKit/SwiftUI
  has leaked into `Features/*/Model/`.

### Still to verify by hand

⌘⎋ can only be confirmed on a real keyboard, and needs Accessibility granted to the build under
test — `Tinycast Dev.app` is its own bundle ID with its own TCC entry, so a grant to an installed
copy does not carry over. Worth confirming too that ⌘⎋ still reaches macOS normally while the
palette is closed.

The manual sweep in `docs/testing.md` has not been run yet.

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -149,6 +149,18 @@ run hover-arming-test      Tinycast/Palette/HoverArming.swift \
                            Tinycast/Features/CustomCommands/Model/CustomCommand.swift
 run palette-escape-test    Tinycast/Palette/PaletteMode.swift \
                            Tinycast/Palette/PaletteEscapeAction.swift \
+                           Tinycast/Features/Settings/EscapeKeyBehavior.swift \
+                           Tinycast/Features/Quicklinks/Model/Quicklink.swift \
+                           Tinycast/Features/Quicklinks/Model/QuicklinkDestination.swift \
+                           Tinycast/Features/CustomCommands/Model/CustomCommand.swift
+run palette-navigation-test Tinycast/Palette/PaletteState.swift \
+                           Tinycast/Palette/PaletteMode.swift \
+                           Tinycast/Palette/HoverArming.swift \
+                           Tinycast/Features/Clipboard/Model/ClipboardStore.swift \
+                           Tinycast/Features/Clipboard/Model/ClipboardFilter.swift \
+                           Tinycast/Features/Clipboard/Model/ColorValue.swift \
+                           Tinycast/Features/Clipboard/Model/ColorFormat.swift \
+                           Tinycast/Features/Clipboard/Model/ColorSpaces.swift \
                            Tinycast/Features/Quicklinks/Model/Quicklink.swift \
                            Tinycast/Features/Quicklinks/Model/QuicklinkDestination.swift \
                            Tinycast/Features/CustomCommands/Model/CustomCommand.swift
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -149,6 +149,7 @@ run hover-arming-test      Tinycast/Palette/HoverArming.swift \
                            Tinycast/Features/CustomCommands/Model/CustomCommand.swift
 run palette-escape-test    Tinycast/Palette/PaletteMode.swift \
                            Tinycast/Palette/PaletteEscapeAction.swift \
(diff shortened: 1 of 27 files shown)


softwarefollower:
There is still no way to completely reset the palette on close, for example: launching clipboard history (either indirectly or directly), there is no way to close the palette and then on relaunch have a completely fresh root search state (currently it always opens back up to the last search/extension).

This is where the ⌘W suggestion comes in, where it closes the palette and resets it to root search in the background, so the next time you launch the palette with ⌘SPACE it is a fresh root search.

PR #508, opened 2026-09-08
Navigate inline command arguments with arrow keys

list-jonas:
## Related issue

None filed — inline command arguments required Tab to leave the query field.

## What changed

Bare Left/Right now cross inline command-argument fields only when the caret reaches a text boundary, matching Raycast’s behavior.

Returning from an argument selects the query. Right first collapses that selection to the query’s end; Right again enters the first argument. The remaining argument order reuses the same focus ring as Tab/Shift-Tab.

Normal caret movement, text selections, modified arrow shortcuts, menus, and grid navigation stay unchanged.

## Tests & validation

All 60 harnesses pass. Debug build succeeds and `./Scripts/lint.sh` is clean.

### DIFF
--- a/Tinycast/Palette/PalettePanel.swift
+++ b/Tinycast/Palette/PalettePanel.swift
@@ -4,12 +4,19 @@ import SwiftUI
 
 /// Borderless floating panel that hosts the SwiftUI command palette.
 final class PalettePanel: NSPanel {
+    enum HeaderFieldBoundary {
+        case leading
+        case trailing
+    }
+
     /// Bare backspace, which the field editor swallows before `onKeyPress` could see it.
     var onBareBackspace: (() -> Bool)?
     /// Command chords the field editor swallows, plus the ones no main menu handles.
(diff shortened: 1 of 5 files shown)


abue-ammar:
@list-jonas i could not get it can you show mw with video? like what are we trying to do here. if it is argument related. do check we have argument for extension eg: brew search extension command, also quicklinks as well, and command argument

list-jonas:
https://github.com/user-attachments/assets/9f856eaf-8a73-411d-afec-6df4011f7f1a

applies to inline arguments in extension commands (e.g. Brew Search) and templated quicklinks.

abue-ammar:
@list-jonas nice nice.  so this is applies to all quicklinks and custom commands?

abue-ammar:
@list-jonas  i rebased it

list-jonas:
> @list-jonas nice nice. so this is applies to all quicklinks and custom commands?

Quicklinks: yes, when they have inline placeholders. Custom commands: no, they have a separate argument screen.

PR #515, opened 2026-09-09
fix(extensions): render search bar dropdowns

### LINKED ISSUE #511: [Bug]: searchBarAccessory is never rendered, so dropdown-filtered extensions list nothing
### Existing issues

### What happened?

The Raycast **Visual Studio Code** extension's `Search Recent Projects` command renders an empty list in Tinycast even though VS Code has recent projects. There is also no project-type dropdown in the search bar, which is the same defect seen from the other side.

This is **not** the upstream Raycast bug. VS Code 1.118 moved `history.recentlyOpenedPathsList` from `globalStorage/state.vscdb` to the shared store at `~/.vscode-shared/sharedStorage/state.vscdb` (raycast/extensions#27440, #27537, #28311). That was fixed in raycast/extensions#28477 (merged 2026-06-02), and the bundle installed here already contains the fix, so the extension does read the right database:

- `~/.vscode-shared/sharedStorage/state.vscdb` holds `history.recentlyOpenedPathsList`; the legacy `~/Library/Application Support/Code/User/globalStorage/state.vscdb` no longer has the key.
- `sqlite3 --json --readonly <shared db> "<the extension's own query>"` returns 1 row / 234 entries.

Running the installed bundle through the repo's own harness shows the data arriving while the list stays empty:

```
$ ext-test "<extensions>/visual-studio-code" index
  proc.run: sqlite3 --json --readonly ~/.vscode-shared/sharedStorage/state.vscdb …
3 render(s); host calls: proc.run
root: list  rows: 0
```

Instrumenting the bundle's `useRecentEntries` shows the query result is fine — `rows=1, entries=20080 chars, parsed=234, merged=235, loading=false` on the final render — so the empty list is the filter, not the data.

**Where it breaks.** `src/index.tsx` keeps the type filter in `useState<EntryType | null>(null)` and only ever sets it from the search-bar dropdown:

```tsx
const [type, setType] = useState<EntryType | null>(null);
…
searchBarAccessory={<EntryTypeDropdown onChange={setType} />}
…
data?.filter(filterUnpinnedEntries(pinnedEntries))?.filter(filterEntriesByType(type))
```

`filterEntriesByType(null)` falls to `default: () => false`, so while `type` is still `null` every entry is filtered out. `type` never changes because the dropdown never reaches the screen:

- the runtime does pass it through — `Scripts/raycast-runtime/src/api/components.js:135` emits `searchBarAccessory` as a `__slot` child, and `Scripts/raycast-runtime/fixtures.mjs:461` asserts it survives;
- `ExtensionScreen.swift:91` parses it into `searchBarAccessory`;
- **nothing reads that property.** `grep -rn searchBarAccessory` over the repo matches only `ExtensionScreen.swift` (declaration and initialiser), the runtime source and that fixture — there is no `List.Dropdown` / `Grid.Dropdown` rendering anywhere under `Tinycast/Features/Extensions/`.

So the accessory is dropped, `onChange` never fires, and the predicate rejects everything. Patching only the initial state in a `/tmp` copy (`useState<EntryType | null>("All Types")`) renders 151 real recent projects on the same bundle, which pins the cause.

A second, independent gap sits behind the first: a minimal probe extension rendering `List.Dropdown` with `defaultValue`, `storeValue` and `onChange` gets **zero** `onChange` calls in the harness. Rendering the accessory alone would therefore still leave this extension filtered to nothing, because Raycast reports the initial selection and the extension is written against that.

**Impact:** any extension that filters its rows from a `searchBarAccessory` dropdown shows an empty or wrong list.

### Steps to reproduce

1. Install the Visual Studio Code extension (`thomas/visual-studio-code`) and make sure VS Code (1.136.2 here) has recent projects — `File > Open Recent` is populated.
2. Run `Search Recent Projects` from Tinycast.
3. Expected: the recent projects list. Actual: `Pinned Projects` and `Recent Projects` are both empty, and the search bar shows no project-type dropdown.

### Screenshots or recording

None — the harness output above reproduces it headlessly.

0.10.15 stable (84)

### macOS version

26.6.2

### Where it breaks / suggested fix

1. Render `searchBarAccessory` in the extension palette's search bar — `List.Dropdown` / `Grid.Dropdown` with their `Item` / `Section` children — and dispatch the selected item's `onChange` handler. `ExtensionScreen` already carries the node; nothing consumes it.
2. Report the initial selection once on mount (`defaultValue`, or the stored value when `storeValue` is set), so a dropdown-filtered extension is not empty before the user touches the control. `storeValue` has no implementation either — `grep storeValue Tinycast/` returns nothing.
3. The runtime bundle is generated, so the change belongs in `Scripts/raycast-runtime/src/api/components.js` followed by `node Scripts/raycast-runtime/build.mjs`, not in the committed `RaycastRuntime.generated.js`.

OiAnthony:
## Summary

- render `List.Dropdown` and `Grid.Dropdown` search-bar accessories in extension commands
- dispatch the initial selection and persist `storeValue` independently from extension `LocalStorage`
- support mouse, keyboard, and `⌘P` interaction using the extension-owned picker surface
- add a JavaScriptCore regression probe that starts with an empty filtered list and verifies the initial dropdown dispatch reveals its rows

## Verification

- `./Scripts/run-tests.sh` — 63/63 harnesses passed
- `xcodebuild build -project Tinycast.xcodeproj -scheme Tinycast -configuration Debug CODE_SIGNING_ALLOWED=NO` — passed
- `./Scripts/lint.sh` — clean
- model-layer AppKit/SwiftUI/Cocoa import check — clean
- regression probe observed red without the initial value dispatch and green with it restored

Closes #511

### DIFF
--- a/Scripts/run-tests.sh
+++ b/Scripts/run-tests.sh
@@ -185,6 +185,11 @@ run palette-navigation-test Tinycast/Palette/PaletteState.swift \
                            Tinycast/Features/Quicklinks/Model/Quicklink.swift \
                            Tinycast/Features/Quicklinks/Model/QuicklinkDestination.swift \
                            Tinycast/Features/CustomCommands/Model/CustomCommand.swift
+run palette-filter-test    Tinycast/Palette/PaletteMode.swift \
+                           Tinycast/Palette/PaletteFilterAction.swift \
+                           Tinycast/Features/Quicklinks/Model/Quicklink.swift \
+                           Tinycast/Features/Quicklinks/Model/QuicklinkDestination.swift \
+                           Tinycast/Features/CustomCommands/Model/CustomCommand.swift
 run palette-tab-test       Tinycast/Palette/PaletteMode.swift \
                            Tinycast/Palette/PaletteTabAction.swift \
                            Tinycast/Features/Quicklinks/Model/Quicklink.swift \
@@ -319,6 +324,10 @@ run ext-form-test          $E/Model/ExtensionFormMetrics.swift \
                            $E/Model/ExtensionDateExpression.swift \
                            $E/UI/ExtensionListKey.swift \
                            Tests/ext-list-key-test.swift
+run ext-accessory-test     $E/Model/RenderNode.swift \
+                           $E/Model/ExtensionPickerItem.swift \
+                           $E/Model/ExtensionSearchAccessory.swift \
+                           $E/Service/ExtensionStorage.swift
 run slow ext-test          -parse-as-library \
                            Tinycast/Platform/Appearance.swift \
                            Tinycast/Platform/Images/IconCache.swift \
@@ -331,6 +340,8 @@ run slow ext-test          -parse-as-library \
                            $E/Model/ExtensionRefreshPolicy.swift \
                            $E/Model/ExtensionRefreshState.swift \
                            $E/Model/RenderNode.swift \
+                           $E/Model/ExtensionPickerItem.swift \
+                           $E/Model/ExtensionSearchAccessory.swift \
                            $E/Service/ExtensionCatalog.swift \
                            $E/Service/ExtensionFetcher.swift \
                            $E/Service/ExtensionIconCache.swift \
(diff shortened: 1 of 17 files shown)


OiAnthony:
image

abue-ammar:
image
this placement just doesnt meet tinycast ui quality sorry. take hint from entire app. have a look on clipboard history filter and its placement

OiAnthony:
@abue-ammar Thanks for the reminder. I’ve made some styling adjustments.
image

Issue #402, opened 2026-08-30
[Feature]: Decouple calendar menu bar display from launcher icon


matheus-apolonio:
## Problem

Hiding Tinycast’s launcher icon currently removes the calendar display as well, because both use the same menu-bar item.

## Proposal

- Give the calendar its own menu-bar item, independent of the launcher icon.
- When both menu-bar preferences are on, show only the calendar item; hiding the launcher icon does not hide the calendar item.
- Add calendar display modes: Disabled, Meeting Icon, and Meeting Title (title • in X min).
- Preserve existing enabled calendar menu-bar preferences as Meeting Icon.
- Add a **Today** option to Show Upcoming Events:
  - show an event anywhere later today;
  - still show an event that starts within 30 minutes after midnight;
  - show `No upcoming events` only after there are no events remaining today and none imminent after midnight.

## Implementation

The implementation is ready on my fork:

- Fork: https://github.com/matheus-apolonio/tinycast
- Branch: `calendar-menu-bar-display`
- Latest commit: `5c2dcb9` Refine calendar menu bar display

The branch has focused coverage for the Today horizon, including a 15:00 event at 10:00 and the 23:29/23:30 midnight boundary. The full local harness suite and lint checks pass.

matheus-apolonio:
Implementation is available in my fork and is ready for a pull request once a maintainer gives the go-ahead required by CONTRIBUTING.md.

Fork: https://github.com/matheus-apolonio/tinycast
Branch: `calendar-menu-bar-display`

Included:
- independent launcher and calendar menu-bar controls;
- calendar display modes: Disabled, Meeting Icon, and Meeting Title;
- meeting title format: `Title • in X min`;
- `No upcoming events` when no events remain today;
- migration for existing calendar menu-bar preferences and updated documentation.

The branch contains two focused commits:
- `d07f9c3` Decouple calendar menu bar display
- `ad853eb` Show empty calendar menu bar state

The automated test suite and lint checks pass locally, and the app has been launched locally for verification.

abue-ammar:
hey sorry about that! and thanks for bringing it. could you open a pr? i will take a look at it. could you tell me little more about it how are you solving this issue. did you put a complete new menubar icon? i would really appreciate that? and if you could attach some screenshot. thanks a lot

matheus-apolonio:
[...]

PR #521, opened 2026-09-09
Recover Calendar access after TCC reset

matheus-apolonio:
## Related issue

Follow-up to #402, whose calendar menu-bar work has already merged.

## What changed

When Calendar permission has been reset in macOS and EventKit reports `Not asked yet`, Settings ▸ Calendar now shows `Allow Calendar Access…`. It reuses the existing confirmation and EventKit request flow even if the Calendar feature preference is already enabled. The denied state still directs the user to System Settings.

## Memory footprint

| Idle after launch       | Not measured |
| Palette open            | Not measured |
| Feature in use (peak)   | Not measured |
| Palette closed, settled | Not measured |

Leak-tested: Not run; this is a small settings recovery path with no new long-lived owner, timer, or observer.

## Demo

Not included. The change only exposes the existing permission flow in a previously unreachable state.

## Drawbacks

The fix is effective only in builds that contain it. An already-installed release without this code cannot trigger the macOS prompt from its `Not asked yet` state.

## Tests & validation

- All 62 harnesses passed, including `calendar-test`.
- `./Scripts/lint.sh` passed (existing warnings only).
- Debug build completed successfully with `CODE_SIGNING_ALLOWED=NO`.
- The manual Calendar sweep now covers resetting permission, retrying from Settings, and the denied-state Settings shortcut.

### DIFF
--- a/Tinycast/Features/Calendar/Settings/CalendarSettingsView.swift
+++ b/Tinycast/Features/Calendar/Settings/CalendarSettingsView.swift
@@ -30,7 +30,18 @@ struct CalendarSettingsView: View {
             }
             .settingsEnabled(settings.calendarEnabled && settings.calendarShowInLauncher)
 
-            if store.access == .denied {
+            if settings.calendarEnabled, store.access == .notDetermined {
+                Section {
+                    SettingsRow(
+                        title: "Calendar access is needed",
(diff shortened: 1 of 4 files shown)


abue-ammar:
what issue it solves care to explain with easy words? why do we need it idk

Issue #451, opened 2026-09-04
[Feature]: Search Screenshots


softwarefollower:
### What should it do?

Built-in feature of Raycast that allows you search quickly through your specified screenshot folder in a gallery view of thumbnails WITH OCR fuzzy searching.

### Why does it belong in Tinycast?

Helps the user quickly find information and references.

abue-ammar:
this has drawback weather we want it or no. because needs gpu acceleration, and heavy cpu usage and potential ram usage. raycast  has this means we don't have to do everything they offer please do understand that

softwarefollower:
No worries, I appreciate how lightweight Tinycast is. Wouldn't want to impact that core design. Just thought I would ask but I didn't know how resource heavy it was.

abue-ammar:
> No worries, I appreciate how lightweight Tinycast is. Wouldn't want to impact that core design. Just thought I would ask but I didn't know how resource heavy it was.

i will  analyze it and see how  big of an impact it will do to the core design and  definitely update here. i am currently traveling so  i will off 2 days

abue-ammar:
for now i am not pushing tinycast to do this. maybe in future if we work on it will update here. closing it for now

abue-ammar:
@softwarefollower  it is now available just merged

softwarefollower:
Thank you, OCR in clipboard is a very useful feature!

This thread was mostly about a specific screenshots folder (specified in the extension settings) with OCR enabled, but I understand if its not part of Tinycast's scope to add the "Search Screenshots" command.

PR #522, opened 2026-09-09
Add opt-in OCR search for clipboard images and PDFs

### LINKED ISSUE #518: [Feature]: Background OCR and PDF text indexing for clipboard search
### What should it do?

Would you welcome a PR adding **on-device image OCR and PDF text indexing to clipboard-history search**? I have a working local prototype and would like to confirm whether this fits Tinycast before preparing an upstream PR.

For example, after copying a receipt screenshot or a PDF file, searching for an invoice number or a phrase inside it would find the original clipboard entry.

Proposed behavior:

- Use Apple's Vision `RecognizeTextRequest` for images, including referenced image files.
- Read embedded PDF text with PDFKit; OCR pages that have no embedded text.
- Retain extracted text as separate SQLite/FTS search metadata across restarts. Include existing history, short queries and pinned entries.
- Preserve the original image/file when copying or pasting. Deleting, pruning or clearing an entry also deletes its extracted text.
- Process one item at a time at background priority, starting after two seconds without input while the palette is hidden. Disabling clipboard history cancels the worker.
- Keep all extraction local, using Apple frameworks without an external service or third-party dependency.

Clipboard history already holds screenshots and file references, but remembering a filename or visually scanning thumbnails is often harder than remembering words in the content. This would make the existing clipboard search useful for those entries without introducing another browser or search surface.

Related requests: #40 asks for pasting an image as OCR text; #451 asks for screenshot-folder search. This proposal focuses on searching content already captured in clipboard history.

### CPU and memory: preliminary measurements

Measured on an **Apple M4 Pro, 12 logical CPUs, 24 GiB RAM, macOS 26.6.2**, using Swift 6.3.3 in Swift 6 mode with `-Onone`.

Five consecutive fresh-process runs of the local `clipboard-text-test` harness, measured with `/usr/bin/time -l`. Each run creates a synthetic 1000 × 300 image with one line of text, extracts it both as a clipboard image and a referenced file, and processes a two-page PDF containing one embedded-text page and one scanned page. The harness also exercises cancellation and idle scheduling with synthetic work. Every run passed 13 checks.

| Whole harness process | Median | Range across 5 runs |
| --- | ---: | ---: |
| Wall time | 1.11 s | 1.08–1.20 s |
| CPU time, user + system | 0.39 s | 0.39–0.46 s |
| Peak resident memory (RSS) | 76.9 MiB | 72.7–81.4 MiB |
| macOS-reported peak memory footprint | 119.5 MiB | 115.2–121.0 MiB |

These are **whole standalone-harness measurements**, including fixture creation, framework initialization and test sleeps. They are neither per-item costs nor the incremental cost inside the running app. Fresh processes do not imply cold OS/framework caches. Separate system-process activity and GPU/Neural Engine utilization are not measured.

In particular, the reported peak footprint is above Tinycast's 100 MB budget even though RSS is lower. **The prototype has not established that the full app stays within its memory budget or returns to baseline after OCR.** Real screenshots, dense/multilingual documents, a large backlog, retained framework memory and battery impact still need profiling before a shipping PR.

### Resource controls and current limits

- One worker; no extraction on the capture or search path. The coordinator and extraction task both use background priority.
- The idle gate is checked before each item, with a two-second wait between items. An active document can finish after input resumes; background priority is not a hard CPU cap.
- While waiting, the worker rechecks the gate every two seconds. Once it observes an empty queue, its task exits.
- Prototype bounds: files up to 32 MB, first 64 PDF pages, images/renders at most 2048 pixels on the longest side, and 32 KB of extracted UTF-8 text per entry.
- At that text cap, 1,000 fully populated resident entries could add about 30.5 MiB of raw text before string/container overhead; pins and SQLite/FTS storage add further costs. This is an upper-bound calculation, not a measured allocation.
- Completed attempts are retained to avoid repeated OCR. Referenced-file edits are not monitored. Images within a PDF page that already has embedded text are not separately OCR'd.

### Prototype status and maintainer input

The local prototype has regression coverage for persistence, search beyond the in-memory window, pinned/short-query matching, deletion, cancellation, and embedded/scanned PDF extraction. The Debug build and lint pass.

Is this feature wanted, and would you prefer it to be opt-in? Feedback on acceptable CPU/memory limits and whether to start with images only or include PDFs would help define a focused PR.

kmplngj:
## Related issue

Closes #518. Also addresses the resource concerns raised in #451.

## What this adds

**"Search text in images and PDFs" in Clipboard History settings, off by default.** The switch is
excluded from settings backups, so importing settings can never turn background processing on for
another Mac. When it is on, Apple's Vision recognizes text in clipboard images and scanned PDF pages,
and PDFKit's embedded text is preferred wherever a page has it. Everything stays on this Mac.

Recognized text is search metadata and nothing else. It lives in its own `item_text` table behind a
trigram FTS index — never on `ClipboardItem`, never in the resident window — so no surface can paste
it, copy it, or classify an entry by it. Original text and path search still answers immediately, and
OCR-only matches arrive afterwards from a cancellable off-main query, with no spinner or skeleton.

## How recognition runs

**In a separate process.** `ClipboardTextWorker` spawns one bundled `ClipboardTextHelper` per item
from `Contents/Helpers` and reaps it before returning, so Vision's and PDFKit's allocations leave
with a process that exits instead of accumulating in a menu-bar app that runs all day. In-process
recognition was the first attempt here and retained too much.

The helper is handed an input path and answers with bounded text down a pipe. It has no database,
clipboard or settings access. The parent accepts at most 32 KB, propagates cancellation, and
terminates and reaps a helper that runs past 60 seconds; its blocking read and wait sit on their own
`DispatchQueue` rather than the cooperative pool.

Scheduling is one item at a time, starting after two seconds without input — including while the
palette is open — with 250 ms between items and no idle polling once the queue empties. Failures go
to a separate `item_text_failures` table for up to three attempts 30 seconds apart, which never block
another item, and a new capture wakes a pending retry wait.

Work is bounded: 32 MB files, the first 64 PDF pages, a 4,194,304-pixel bitmap budget with a
4096-pixel maximum edge, and 32 KB of text per item. Long bitmaps are recognized in overlapping
2048-pixel tiles, with Vision's relative minimum text-height cutoff disabled so small text on a tall
screenshot or page survives.

## Build and release changes

Shipping a second executable needs a second target, so `project.yml` gains a `ClipboardTextHelper`
tool target, an `excludes:` that keeps its two sources from also compiling into the app, and a
dependency that builds, embeds and signs it into `Contents/Helpers`. **No third-party dependencies:**
the helper links only Foundation, CoreGraphics, ImageIO, PDFKit and Vision, and the app binary ends
up with zero Vision/PDFKit links.

`EXECUTABLE_NAME` and `PRODUCT_MODULE_NAME` are pinned on that target because `release.yml` passes
`PRODUCT_NAME` on the xcodebuild command line, which applies to every target in the project —
without them, a channel build renames the executable the app looks for.

`release.yml` now asserts architecture slices on **both** shipping binaries. Both jobs previously
checked `Contents/MacOS` alone, and a bundled helper now sits beside it. A thin helper inside a
universal build is the quiet form of the bug that guard exists to prevent: the app still boots on
Intel, `posix_spawn` fails with `EBADARCH`, the indexer burns three retries per item into private
OSLog, and clipboard OCR simply never answers.

## Verification

- **67/67 harnesses pass**, including new coverage for real Vision recognition, embedded and scanned
  PDFs, retry backoff and exhaustion, cancellation, deadline termination, output bounds, child
  reaping, and an independent ordering oracle over 2500 entries across every type filter.
- `./Scripts/lint.sh` clean, `./Scripts/format.sh --check` clean, Debug build with no new warnings,
  and all Model files pass the import-purity grep.
- **Release build with the channel override** (`PRODUCT_NAME="Tinycast Beta"`): the helper embeds
  under its fixed name and the bundle passes `codesign --verify --deep --strict`.
- **Universal release build** (`ARCHS="arm64 x86_64"`): both binaries carry both slices at a macOS
  26.0 floor. The two new release guards were run against real thin and universal bundles in both
  directions — each passes its own bundle and catches the helper in the other.
- **Manual, on a real build:** searching a freshly copied image containing text. Parent RSS stays at
  60–70 MB, and no helper process survives the run.

`ci.yml` has no `xcodebuild` step by design, so the `release.yml` guards are not exercised by this
PR's CI — hence the local verification above.

## Drawbacks

- **Combined parent + helper memory during the ~1 second a helper is alive is not characterised.**
  The parent staying flat is what was checked; the transient overlap of the two processes was not.
- One process per item repeats startup, trading some throughput for deterministic release of
  extraction memory. Large-document throughput is unmeasured.
- **Raycast quality parity is unverified.** English, German, small-font and tall-image cases passed a
  small corpus; Japanese missed a glyph. Noisy scans, rotation and broader language coverage need
  testing.
- Persistent errors stop after three attempts. Enabling text search retries earlier failed and empty
  attempts, including on a later launch, so an empty input may be reprocessed — but nothing retries
  forever within a session.
- File, page and text limits can omit content. A page that already has embedded text does not
  separately OCR images on it, and editing a referenced file does not refresh text already recognized
  from it.
- Overlapping tiles are not deduplicated, so text falling in an overlap appears twice in the derived
  string. Harmless for substring matching, but the stored text is not a faithful transcript.
- Ordinary matches deliberately precede OCR-only unpinned matches, rather than interleaving by
  recency.


abue-ammar:
it does not work. i have tested it. also ram goes high and never goes down. at this rate i might not want this in tinycast as i have discussed this matter in https://github.com/abue-ammar/tinycast/issues/451
https://github.com/user-attachments/assets/a8660b9a-b671-4188-bc1f-124eb8119309

kmplngj:
Strange works really nice in my local build. Any ideas what the issue caused?

Did you have an very very big history in the clipboard that perhaps needs so long to work on?

Did you let the Apple Vision Framework let the time to work on the clipboard contents? It perhaps take some time.

abue-ammar:
> Strange works really nice in my local build. Any ideas what the issue caused?
> 
> Did you have an very very big history in the clipboard that perhaps needs so long to work on?
> 
> Did you let the Apple Vision Framework let the time to work on the clipboard contents? It perhaps take some time.

i dont really have much history like 20-30 at most. and in raycast it works fine. in tinycast i noticed higher ram usage but output is 0. you can take a look and i am in macos 27 latest beta. i will see. but i am not okay with the ram usage. and this is not a very well demanded feature

PR #540, opened 2026-09-10
Add a palette background transparency control

hkay-dev:
**I used Codex to help write and check this change.**

## Related issue

There isn't a linked GitHub issue yet.

## What changed

General settings now have a five-position Background transparency slider and a Reset button. The slider adjusts the tint over the palette's existing system blur in Light and Dark. The center position and Reset use the original background color and window shadow.

CleanShot 2026-09-10 at 3 40 34 AM

Custom positions add a white border. The two more transparent Dark positions turn off the system shadow and its black outline, and use a one-point gradient border with a brighter top edge. The setting saves when a drag ends and on keyboard adjustments, and participates in settings search and backups.

The background observes the setting in its own view. Colors and the slider width live in Theme, and the settings row allows the description to wrap onto two lines. The palette and UI documentation describe the new behavior.

## Memory footprint

Debug build on macOS 27.0 beta, measured with `footprint`. These figures are physical footprint in decimal MB from a fresh process.

| Step | Memory |
| --- | --- |
| Idle after launch | 38.62 MB |
| Palette open with query `c` | 59.48 MB |
| Process peak during Settings and slider checks | 304.79 MB |
| Palette and Settings closed, settled | 85.97 MB |

The process peak exceeds the repository's 100 MB limit, and closing the windows did not return memory to the idle reading. This run included opening Settings and changing the slider. I haven't compared the same sequence with an unchanged build, so these numbers don't show whether the transparency change increases memory use or caused the spike.

The leak check did not pass. `leaks --noContent --nostacks` reported 430 allocations totaling 20,608 bytes, with three `NSXPCConnection` root cycles. An earlier check of unchanged `main` at `1855c81` also reported three `NSXPCConnection` root cycles, but it used a different sequence of interactions.

## Demo

The two videos show the same query `c`, window dimensions, and desktop background in Light and Dark. Each clip lasts 10 seconds and shows each position for two seconds, labeled -2, -1, 0, +1, and +2. Position 0 keeps the original appearance.

### Dark

[tinycast-dark-transparency-detents-10s.webm](https://github.com/user-attachments/assets/5809ea90-d43d-4cf7-a294-b5e5d485f51f)

### Light

[tinycast-light-transparency-detents-10s.webm](https://github.com/user-attachments/assets/d59810d7-6f26-421f-a19a-6672a8dbac32)

## Drawbacks

This changes tint opacity over the existing blur. The system material still controls blur strength. The two more transparent Dark positions lose the native drop shadow.

The available videos show the five settings in sequence. They do not meet the template's side-by-side before/after format.

## Tests & validation

- All 64 harnesses passed.
- The appearance harness checks the unchanged default in both themes, tint direction, range endpoints, extreme inputs, and border colors.
- The Debug build succeeded with no new source warnings, and lint passed with existing warnings.
- Model import purity, settings search checks, and `git diff --check` passed.
- Checked all five positions in Light and Dark, Reset, saved settings, and the border appearance in the running app.
- Read the full 12-file diff against `main` at `1855c81`.
- The backup harness checks setting coverage. I checked export and import handling in the source; this field does not have a round-trip test.


hkay-dev:
I'm going to continue to research that leak issue before I decide this is ready.

abue-ammar:
> I'm going to continue to research that leak issue before I decide this is ready.

i think that's  probably not a potential leak it could be short burst process which is okay. but still better to check. short burst are okay because chasing 100mb hard cap and then quality  degrades that will be bad too.

hkay-dev:
I've checked to the best of my ability but it looks like this only happens during the initial changing of the setting. At rest and during normal operations it doesn't look like there are any memory leaks caused directly by this commit. 

I'm cautiously submitting this for your approval to see what you think.

abue-ammar:
> I've checked to the best of my ability but it looks like this only happens during the initial changing of the setting. At rest and during normal operations it doesn't look like there are any memory leaks caused directly by this commit.
> 
> I'm cautiously submitting this for your approval to see what you think.

it's good i have checked it! nice work i am merging it

PR #551, opened 2026-09-10
Let the Permissions pane ask for calendar access

### LINKED ISSUE #538: [Bug]: Unable to access calendar and can not request access as well
### Existing issues

### What happened?

I can not access calendars and i can not request access as well

### Steps to reproduce

Open TinyCast
Go to Settings 
Go to Permissions

### Screenshots or recording

Image

Version 0.10.15 (84) + github latest release

### macOS version

 26.6.2

abue-ammar:
## Related issue

Closes #538

## What changed

The reported status was correct — TCC genuinely had no record for Tinycast, which is why the app is
absent from the Calendars list in the issue's screenshot. `EKEventStore.authorizationStatus(for:
.event)` returned `.notDetermined`, so "Not asked yet" was accurate. The bug was that **nothing in
the app could ask, and System Settings could not be used instead**:

- **Privacy & Security ▸ Calendars has no "+" button.** macOS lists only apps that have already
  prompted, so the pane's `Open…` button sent people to a screen where Tinycast cannot be added.
- **`setCalendarEnabled` wrote `settings.calendarEnabled = true` before `requestAccess()`.** A prompt
  that threw or was dismissed left the switch reading as on with no access.
- **The re-ask guard was `store.access == .notDetermined`** (and before #521, `enabled !=
  settings.calendarEnabled`), so with the switch already on, nothing could ask again. That is
  exactly "can not request access as well" on v0.10.15.

#521 fixed the Calendar pane only, and it is in `v0.10.17-beta.86` — not in stable. The **Permissions**
pane, which is where the issue's repro steps go, was still a dead end on `main`.

Three changes:

- `PermissionsSettingsView` offers **Grant Access…** when access is `notDetermined`, routed through
  `CalendarCoordinator.setCalendarEnabled(true)` so Tinycast's own confirm dialog still comes first.
  `.granted` and `.denied` keep the `Open…` button — a denial is the one state System Settings can
  undo.
- `calendarEnabled` is written **after** macOS grants, so a failed prompt can never strand the
  feature with the switch on.
- The re-ask guard widens from `access == .notDetermined` to `access != .granted`.

The `MenuPanel.swift` hunk is unrelated pre-existing drift that `Scripts/format.sh` corrected.

Ruled out along the way: signing and Info.plist are not involved. A probe app signed with the
`Tinycast Self-Signed` identity, carrying the same `NSCalendarsFullAccessUsageDescription`, prompts
correctly on macOS 27 and then reads `fullAccess` with 3 calendars.

## Memory footprint

Not measured — no allocation, no new state and no new long-lived owner. Two `Bool` branches in an
existing settings row, and one statement reordered in a coordinator that already existed.

| Step                    | Memory        |
| ----------------------- | ------------- |
| Idle after launch       | not measured  |
| Palette open            | not measured  |
| Feature in use (peak)   | not measured  |
| Palette closed, settled | not measured  |

Leak-tested: no — nothing is retained, observed or cancelled by this change.

## Demo

The Calendars row in the Permissions pane gains a `Grant Access…` button in the `notDetermined`
state; everything else is unchanged. I could not capture a side-by-side clip: `osascript` has no
assistive access on this machine, so the Settings window cannot be driven to that pane
programmatically. Happy to add one if it's wanted before merge.

## Drawbacks

- Pressing **Grant Access…** from the Permissions pane turns the calendar feature on as well, because
  `calendarEnabled` *is* the consent record. The confirm dialog says so before macOS prompts, and it
  is the same path as the Calendar pane's switch, so there is no second consent shape.
- Users already stuck with `calendarEnabled = true` plus no TCC record keep that state in their prefs.
  #521's recovery section in the Calendar pane still covers them, and the Permissions pane now does
  too, so it is recoverable rather than requiring a defaults reset.

## Tests & validation

- `./Scripts/run-tests.sh` — all 64 harnesses pass.
- Debug build: no new warnings.
- `./Scripts/lint.sh` — clean.
- Model-purity grep over `Features/*/Model/` — clean.
- By hand: built a probe app under the `Tinycast Self-Signed` identity to confirm
  `requestFullAccessToEvents()` prompts and that `authorizationStatus` then reports `fullAccess`
  (3 calendars) on macOS 27, and that a terminal-launched binary is attributed to its parent process
  rather than the bundle — which is why the diagnosis used a launched `.app`.

### DIFF
--- a/Tinycast/Features/Calendar/Service/CalendarStore.swift
+++ b/Tinycast/Features/Calendar/Service/CalendarStore.swift
@@ -39,10 +39,15 @@ final class CalendarStore {
         hiddenCalendarIDs = Set(defaults.stringArray(forKey: hiddenKey) ?? [])
     }
 
+    /// TCC sends nothing when a grant changes in Settings, so anything acting on `access` re-reads.
+    func refreshAccess() {
+        access = Permissions.calendarAccess()
+    }
+
     // MARK: - Lifecycle
 
     func start() {
-        access = Permissions.calendarAccess()
+        refreshAccess()
         guard access == .granted else { return }
         observeWake()
         // Deferred: the first EventKit query pays for its XPC warm-up, and launch protects itself.
@@ -61,7 +66,7 @@ final class CalendarStore {
     /// Tinycast's own consent dialog has already been accepted by the time this runs.
     func requestAccess() async -> Bool {
         let granted = await Permissions.requestCalendarAccess()
-        access = Permissions.calendarAccess()
+        refreshAccess()
         guard granted else { return false }
         // A store built before the grant never sees the new calendars; drop it and rebuild.
         changeObserver = nil
@@ -111,7 +116,7 @@ final class CalendarStore {
 
     /// `EKEventStore` is not `Sendable`, so this stays on main; only pure values leave.
     func reload() {
-        access = Permissions.calendarAccess()
+        refreshAccess()
         guard access == .granted else {
             publish([])
             return
--- a/Tinycast/Features/Calendar/Settings/CalendarSettingsView.swift
+++ b/Tinycast/Features/Calendar/Settings/CalendarSettingsView.swift
@@ -140,6 +140,8 @@ struct CalendarSettingsView: View {
         .formStyle(.grouped)
         .settingsScrollTarget(.calendar)
         .releasesFocusOnOutsideClick()
+        // The snapshot is only reloaded while the feature runs, so a grant made in Settings lands here.
+        .onAppear { store.refreshAccess() }
     }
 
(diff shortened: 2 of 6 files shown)


Sheldontao:
https://github.com/user-attachments/assets/721aabea-a177-4b75-afc2-75736b3638ee

still can't grant calendar access on v0.10.20(already restart tinycast)

macbookair M4 
macos27

abue-ammar:
> Capso.Recording.2026-09-13.at.08.23.31.mp4 
> still can't grant calendar access on v0.10.20(already restart tinycast)
> 
> macbookair M4 macos27

put it in #538 thanks will fix it

PR #553, opened 2026-09-10
Step back to the root search instead of closing the window

### LINKED ISSUE #529: [Bug]: Esc and Back Button Close TinyCast Instead of Navigating Back
### Existing issues

- [ ] I searched open and closed issues and didn't find this already reported

### What happened?

When opening Tinycast (main search field) and click Tab button (navigating to the Clipboard list), pressing Esc or the Back button closes Tinycast completely instead of navigating back to the previous screen/ main search field

### Steps to reproduce

1. Open TinyCast using Cmd + Space.
2. Click Tab to open the clipboard list.
3. Press Esc or press back button

### Screenshots or recording

_No response_

0.10.15

### macOS version

26.0

list-jonas:
## Related issue

Closes #529 — Esc and the back chevron close Tinycast instead of returning to the previous screen.

## What changed

`pop()` treated an empty back stack as "nothing behind this screen", but the root search sits below
every screen there is. Three ways in stack nothing — a hotkey summon, a Tab across the ring, and any
`prepare` — so Escape closed the window from a clipboard opened a keystroke earlier.

`pop()` now falls back to the root search when the stack is empty, and `canGoBack` is
`mode != .launcher` rather than a stack test. Escape, the chevron and a bare backspace all route
through `pop()`, so the three back edges agree without touching any of them. `navigate(to:)` no
longer pushes the launcher — it is the floor every stack stands on, so the toggle hotkey pressed over
a screen was stacking the root search on top of it and Escape walked *down* into what was just left.

`closeAndPopToRoot` is unaffected: the behavior is read before the stack is, so the fallback below it
is never reached. The chevron's tooltip reads that behavior now, since under it Escape closes while
the chevron still steps back.

## Tests & validation

All 64 harnesses pass, Debug builds with no new warnings, `lint.sh` clean. `palette-navigation-test`
gains the hotkey-summoned case and asserts the root-search fallback where it previously asserted a
dead end; `palette-escape-test` pins that close-and-pop-to-root outranks the fallback.

### DIFF
--- a/Tests/palette-escape-test.swift
+++ b/Tests/palette-escape-test.swift
@@ -111,6 +111,11 @@ struct PaletteEscapeTests {
             resolve(menuOpen: true, canGoBack: true, behavior: .closeAndPopToRoot),
             .closeMenu,
             "a menu outranks the behavior setting beneath it")
+        // The setting is read before the stack is, so the fallback below it never reaches this.
+        expect(
+            resolve(mode: .clipboard, behavior: .closeAndPopToRoot),
+            .hidePalette,
+            "a screen summoned by its own hotkey closes under close-and-pop-to-root")
 
         expect(
             resolve(menuOpen: true, mode: .ai),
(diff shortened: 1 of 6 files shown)


abue-ammar:
i am going this this https://github.com/abue-ammar/tinycast/pull/554 because it is smaller and less work

PR #615, opened 2026-09-12
Select the search field's text when the palette is re-summoned

### LINKED ISSUE #611: [Feature]: Auto select text from palette after reopening
### What should it do?

current behavior: I type something in Tinycast, click somewhere outside and then again click it, the previous text is still there. So I have to backspace and then type my new search. There is an option to immeadiately clear the root search but it would be much more convinient if the text got selected, thereby giving the option to always have the old text, and also immediately writing over it without any additional clicks

new behavior: after I click something outside, and then again open tinycast, keep the text in the search bar selected/highlighted so that I can quickly start typing my new search without any new clicks. and this way if someone wants the old search, they still have that too (as its just selected/highlighted and not deleted)

It saves quick time and additional clicks XD

### Contribution

- [x] I'm willing to implement this myself

shailantani:
## Summary
- Closing the palette (click-outside/Escape) and reopening it left the previous query sitting in the field with the caret wherever it was last, so starting a new search meant manually backspacing the old one out first.
- Reopening now selects the leftover query when it's still relevant to your Pop to Root Search setting: typing immediately replaces it, and pressing → moves the caret to the end and keeps it — both via standard `NSTextView` selection behavior, no extra key handling needed.
- This respects each existing Pop to Root Search option's own timing, with no new setting added:
  - **Immediately**: unaffected — the query is cleared the instant the window hides, same as before.
  - **After N seconds**: reopening *before* the delay elapses selects the preserved query; reopening *after* it has already fired shows an empty root search, same as before.

## Changes
- `Tinycast/Palette/PalettePanel.swift`: adds `selectAllFieldEditorText()`, selecting the shared AppKit field editor's full text.
- `Tinycast/Palette/PaletteWindowController.swift`: `consumePreservedState()` already reports whether a reopen beat the pending Pop to Root timer — it now also flags that the surviving query is worth selecting, and `windowDidBecomeKey` calls the new select-all only when that flag is set. `Immediately` never has a timer to survive, so it's never affected.

## Test plan
- [x] `xcodebuild -project Tinycast.xcodeproj -scheme Tinycast -configuration Debug -destination 'platform=macOS' build` succeeds
- [x] "Immediately": type a query, click outside, reopen — root search is empty, no leftover text
- [x] "After 5 seconds": reopening within 5s selects the preserved query; reopening after 5s shows an empty root search
- [x] Manually verified normal in-palette navigation (mode switches, tabbing) is unaffected

Closes #611

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Tinycast/Features/Settings/AppSettings.swift
+++ b/Tinycast/Features/Settings/AppSettings.swift
@@ -15,17 +15,11 @@ enum PopToRootTimeout: Int, CaseIterable, Identifiable, Sendable {
     case afterThirty = 30
     case afterSixty = 60
     case afterNinety = 90
-    /// Never pops: the next summon shows the last query selected, ready to be typed over.
-    case selectPreviousQuery = -1
 
     var id: Int { rawValue }
 
     var title: String {
-        switch self {
-        case .immediately: return "Immediately"
-        case .selectPreviousQuery: return "Select previous query"
-        default: return "After \(rawValue) seconds"
-        }
+        self == .immediately ? "Immediately" : "After \(rawValue) seconds"
     }
 
     var interval: TimeInterval { TimeInterval(rawValue) }
--- a/Tinycast/Features/Settings/AppSettings.swift
+++ b/Tinycast/Features/Settings/AppSettings.swift
@@ -15,11 +15,17 @@ enum PopToRootTimeout: Int, CaseIterable, Identifiable, Sendable {
     case afterThirty = 30
     case afterSixty = 60
     case afterNinety = 90
+    /// Never pops: the next summon shows the last query selected, ready to be typed over.
+    case selectPreviousQuery = -1
 
     var id: Int { rawValue }
 
     var title: String {
-        self == .immediately ? "Immediately" : "After \(rawValue) seconds"
+        switch self {
+        case .immediately: return "Immediately"
+        case .selectPreviousQuery: return "Select previous query"
+        default: return "After \(rawValue) seconds"
+        }
     }
 
     var interval: TimeInterval { TimeInterval(rawValue) }
--- a/Tinycast/Palette/PalettePanel.swift
+++ b/Tinycast/Palette/PalettePanel.swift
@@ -32,8 +32,6 @@ final class PalettePanel: NSPanel {
     /// SwiftUI's text fields all edit through the window's one shared field editor.
     private var fieldEditor: NSTextView? { firstResponder as? NSTextView }
 
-    /// A re-summon leaves the field editor's selection as it was; select its text outright
-    /// so a fresh search replaces last time's instead of requiring a manual clear first.
     func selectAllFieldEditorText() {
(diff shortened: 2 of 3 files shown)


list-jonas:
image

Your PR currently overrides the **Pop to Root Search: Immediately** behavior. When enabled, the query should be cleared and must not reappear selected after reopening.

Same applies to delayed options, e.g. 5 seconds: reopening within 5s may preserve/select the query, but reopening after Pop to Root fires should show an empty root search.

shailantani:
> image
> 
> Your PR currently overrides the **Pop to Root Search: Immediately** behavior. When enabled, the query should be cleared and must not reappear selected after reopening.
> 
> Same applies to delayed options, e.g. 5 seconds: reopening within 5s may preserve/select the query, but reopening after Pop to Root fires should show an empty root search.

Yep let me fix that. I'll add a new option in the pop to root search.
That'll help always have the previous query selected/highlighted

shailantani:
@list-jonas can you check it out now? Added a new option. That shouldnt be conflicting with the immediate or the other options. 

Thanks!

list-jonas:
I think the **Select previous query** option adds unnecessary confusion, since Raycast doesn’t have this setting and it can’t be meaningfully combined with the existing Pop to Root Search timeout options. It would be better to replicate Raycast’s behavior, as that’s what most users will already be familiar with:

- Immediately: query is cleared and not persisted.
- After 5 seconds: query is persisted and auto-selected when reopened within 5 seconds. After 5 seconds, it is cleared and returns to root.

The same behavior should apply to the other timeout options.

shailantani:
> I think the **Select previous query** option adds unnecessary confusion, since Raycast doesn’t have this setting and it can’t be meaningfully combined with the existing Pop to Root Search timeout options. It would be better to replicate Raycast’s behavior, as that’s what most users will already be familiar with:
> 
>     * Immediately: query is cleared and not persisted.
> 
>     * After 5 seconds: query is persisted and auto-selected when reopened within 5 seconds. After 5 seconds, it is cleared and returns to root.
> 
> 
> The same behavior should apply to the other timeout options.

Behavior of Raycast is exactly what I was going for, the default behavior in Raycast is select the previous query in a given time
Thats why I implemented it here as well

Let me do this:
I'll add the select/highlight previous query within the options of Tinycast (and it follows the time ie, 5s , 15s etc)
And if its immediate then it doesnt highlight anything and its just a new query.
This way theres no confusion, and its just a default behavior like Raycast

Thanks for the input

shailantani:
@list-jonas I think this commit should solve everything. It just adds the select query feature for existing options. And also respects the time (5s,15s so on) and also doesnt show any query when immediately is selected.

list-jonas:
Looks good to me @abue-ammar

abue-ammar:
@shailantani  remove claude's co authroship from commit message

image

PR #647, opened 2026-09-13
Fix System Monitor metrics in extensions

### LINKED ISSUE #631: [Extension]: System Monitor CPU metrics remain stuck loading
## Existing issues

- [x] I searched open and closed issues and did not find this already reported.

## Extension URL

https://www.raycast.com/hossammourad/system-monitor

## What happened?

Opening System Monitor and selecting CPU leaves most of the detail pane permanently on `Loading…` or `Collecting sample…`. The aggregate CPU accessory originally rendered `NaN %`. After supplying CPU totals, the aggregate became numeric, but chip/core/GPU metadata, top processes, per-core usage, fans, load averages, and uptime still did not start or complete.

The failures have three runtime causes:

1. Tinycast does not forward `List.onSelectionChange`, so the extension keeps its internal selected view at System Info and deliberately skips CPU-only collectors. This also occurs when local filtering replaces the visible row without changing the numeric selection index.
2. The Node `os` shim lacks real `cpus()`, `freemem()`, `uptime()`, and `loadavg()` values required by `os-utils`.
3. Extension downloads discard Git mode `100755`, leaving the bundled temperature and SMC fan helper binaries non-executable. Existing installs also need permission repair.

Expected: selecting CPU populates every field and keeps refreshing CPU/process/core readings.

## Steps to reproduce

1. Install the Raycast System Monitor extension in Tinycast.
2. Run **System Monitor**.
3. Select the **CPU** row.
4. Wait at least ten seconds.
5. Observe that CPU details remain on loading/collecting placeholders.

## Screenshots or recording

A reproduction screenshot was captured on the affected build: the CPU row reports a numeric percentage while Chip, CPU Cores, GPU Memory, Top Processes, Per-Core Usage, Fans, Average Load, and Uptime remain on `Loading…` / `Collecting sample…`. I can attach the original PNG in a follow-up if needed.

## Tinycast version + channel

0.10.20 stable (build 528 when reproduced)

## macOS version

27.0 (26A428)

## Extension limits

- [x] I understand this is closed if the extension depends on Raycast-only capabilities.

## Contribution

- [x] I am willing to implement this myself once this issue is labelled `approved`.

A tested implementation is ready on `wowlocal:fix/system-monitor-cpu-metrics`. It includes runtime fixtures, Swift harness coverage, a universal release build, and live verification in the installed notarized app.

wowlocal:
## Related issue

Closes #631

Branch: https://github.com/wowlocal/tinycast/tree/fix/system-monitor-cpu-metrics

## What changed

- Back Node OS shims with real macOS data for `os.cpus()`, `os.freemem()`, `os.uptime()`, and `os.loadavg()`; remove the static CPU-count boot value.
- Preserve helper executable metadata from the extension store and repair bundled Mach-O or shebang helpers that lost executable permission during installation.
- Forward List/Grid selection changes by visible item ID through the extension-owned `ExtensionSelectionForwarder`; `RootPaletteView` keeps one integration line.
- Regenerate the embedded Raycast runtime and add focused regression coverage for OS metrics, helper repair, and selection forwarding.

## Why selection forwarding is part of this fix

System Monitor keeps `selectedItemId` in React state, updates it through `List.onSelectionChange`, and passes flags such as `isActive={selectedItemId === "cpu"}` to its monitors. The lightweight CPU summary runs continuously, while top processes, per-core sampling, temperature, fans, load averages, and uptime only execute for the active item.

Tinycast previously moved its native highlight through `vm.selection` without dispatching the rendered `onSelectionChange` handler. The UI could show CPU as selected while the extension still considered System Info active. Fixing the OS shim therefore made the summary numeric, but the active-only detail collectors still remained on loading placeholders.

The forwarder observes the resolved item ID rather than only the numeric index because local filtering can replace row zero while `vm.selection` remains zero.

## Reproduction

1. On current upstream `main`, enable extensions and install System Monitor.
2. Open the System Monitor command; it starts on System Info.
3. Select the CPU row.
4. Confirm the native row highlight moves to CPU.
5. Observe that the extension does not receive `onSelectionChange("cpu")`; the CPU detail stays on `Loading…` or `Collecting sample…`. Before the OS-shim fix, the summary also reports `NaN %`.
6. Switch among CPU, Memory, and Disk. The native selection moves, but the extension-side active item remains unchanged.

## Memory footprint

Measured as resident memory from the installed notarized 0.10.20 (530) build during a long-running session:

| Step | Memory |
| --- | ---: |
| Idle after launch / palette closed | 384 MB |
| Palette open | 382 MB |
| Feature in use (peak observed) | 388 MB |
| Palette closed, settled | 384 MB |

Leak-tested: exercised an open, load, back, and close cycle. No sustained-growth loop was run.

## Drawbacks

- Extension catalog scans now inspect a bounded header from regular asset files when executable metadata is unavailable, so installation recovery has a small amount of additional file I/O.
- The synchronous OS calls run on the existing extension JS thread, matching the other Node shims.
- There is no intentional visual design change, so no demo video is included.

## Tests & validation

- `./Scripts/run-tests.sh` — all 72 harnesses passed
- `node Scripts/raycast-runtime/fixtures.mjs`
- `./Scripts/run-tests.sh ext-test`
- `./Scripts/format.sh --check`
- `git diff --check origin/main...HEAD`
- Pure-model import check
- Debug app build with code signing disabled
- Developer ID signed and notarized universal build 0.10.20 (530) installed over the existing app and accepted by Gatekeeper
- Live System Monitor verification: numeric total CPU, Apple M4 Max / 16 cores, shared 48 GB GPU memory, top processes, per-core usage, temperature, both fans, load averages, and uptime all populated

`swiftlint` is unavailable on this machine, so `Scripts/lint.sh` could not run its SwiftLint phase.


abue-ammar:
@wowlocal why does it need changes in `Tinycast/Palette/RootPaletteView.swift` ? it is already 1400+ lines

list-jonas:
> @wowlocal why does it need changes in `Tinycast/Palette/RootPaletteView.swift` ? it is already 1400+ lines

The changes in `Tinycast/Palette/RootPaletteView.swift` are required so extensions know which item is selected. 

I’d keep only 1–3 lines there and move the logic into e.g. `ExtensionSelectionForwarder` bellow `Features/Extensions/UI`.

wowlocal:
Thanks — the `RootPaletteView` change was not required for `os.cpus()` itself. It addressed a second bug exposed after the Node OS shim started returning real data.

System Monitor keeps `selectedItemId` in React state, updates it only through `List.onSelectionChange`, and passes flags such as `isActive={selectedItemId === "cpu"}` into each monitor. The lightweight CPU summary runs continuously, but top processes, per-core sampling, temperature, fans, load averages, and uptime are gated by that active flag.

Tinycast already moved the native highlight through `vm.selection`, but it did not dispatch the rendered `onSelectionChange` handler back into the extension runtime. The UI could therefore show the CPU row as selected while the extension still believed the initial System Info row was active. That is why fixing the OS shim changed `NaN %` into a numeric CPU value, while the CPU detail still remained on `Loading…` / `Collecting sample…`.

### Reproduction on current upstream `main`

1. Enable extensions and install System Monitor.
2. Open the System Monitor command; it starts on System Info.
3. Select the CPU row.
4. Observe that the native row highlight moves to CPU.
5. The extension does not receive `onSelectionChange("cpu")`, so active-only CPU collectors do not start. The detail pane stays on loading/collecting placeholders. Before the OS shim part of this PR, the summary also reports `NaN %`.
6. Switch between CPU, Memory, and Disk: Tinycast changes the selected native row, but the extension-side `selectedItemId` remains at its initial value.

The observer uses the resolved item ID rather than only the numeric index because local filtering can replace row zero without changing `vm.selection == 0`; that replacement must still emit a selection change.

I applied the suggested extraction in commit `14d3755`:

- `RootPaletteView` now has one integration line: `ExtensionSelectionForwarder(screen: extensionScreen, selection: vm.selection)`.
- Selection clamping, item-ID observation, `null` handling, and runtime dispatch now live in `Features/Extensions/UI/ExtensionSelectionForwarder.swift`.
- `ExtensionScreen` remains the single source of visible extension row order.

Validation after the refactor: all 72 harnesses pass, `ext-test` passes independently, formatting and diff checks pass, the pure-model import check is clean, and the Debug app build succeeds. SwiftLint is not installed on this machine, so `Scripts/lint.sh` could not run its SwiftLint phase.

list-jonas:
Looks good overall - nothing worth stripping down. Two things i noticed:
- restoreExecutablePermissions runs on every scan. It's called from ExtensionCatalog.scan(), so each refresh() re-opens every asset to sniff magic bytes - 405 opens and ~17ms warm on a 29-extension install, to find 7 executables.
- os.cpus() reports placeholder hardware. model: "Apple Silicon" and speed: 0 for every core, which is wrong on Intel, but ngl I dont think its worth fixing.

Also pushed a small fix on top: the initial onSelectionChange fired with row zero, overwriting an extension's own selectedItemId - System Monitor's "Default tab" got reset to System Info on every launch. Seeding the selection from selectedItemId before the first dispatch fixes it.

abue-ammar:
@list-jonas looks a lot cleaner

PR #739, opened 2026-09-15
Fix time-zone source parsing

### LINKED ISSUE #727: [Bug]: Unrecognized source text silently changes time-zone conversions to the Mac's zone
### Existing issues

- [x] I searched open and closed issues and didn't find this already reported.

### What happened?

Adding a space before `pm`, or an `in` before the source city, can turn a correct conversion into a wrong answer. An unrecognized source such as `PSTT` also produces a time instead of being rejected.

Reproduced by compiling the unmodified `CalcEngine` model on both v0.10.23 and main at `8f7aebc1492ad84dbea3386299315a5373e4f653`. With `now` pinned to **2026-09-15 12:00 UTC** and the injected calendar using **Asia/Shanghai**:

| Query | Actual result in both versions | Intended conversion |
| --- | --- | --- |
| `5:30pm SF to London` | 1:30 AM (tomorrow), source Los Angeles | Correct control |
| `5:30 pm SF to London` | 10:30 PM (yesterday), source Shanghai | 1:30 AM on September 16 |
| `5pm SF to London` | 1:00 AM (tomorrow), source Los Angeles | Correct control |
| `5pm in SF to London` | 10:00 AM, source Shanghai | 1:00 AM on September 16 |
| `5pm PSTT to London` | 10:00 AM, source Shanghai | Reject the unknown source |

The explicit SF-to-London conversion also changes when only the injected home zone changes. For `5:30 pm SF to London`, home UTC gives **6:30 AM**, home Shanghai gives **10:30 PM (yesterday)**, and home Los Angeles gives **1:30 PM**. The intended conversion is **1:30 AM on September 16** in all three contexts.

`sourceMoment` parses only the first whitespace-separated word as a clock, then looks up all remaining source words as a zone. If that lookup fails, it falls back to `calendar.timeZone`. Consequently, `pm SF` loses both its meridiem and its zone, and `in SF` loses its zone. The clock branch doesn't reject the leftover text as the `time`/`now` branch does. [Source](https://github.com/abue-ammar/tinycast/blob/8f7aebc1492ad84dbea3386299315a5373e4f653/Tinycast/Features/Calculator/Model/CalcTimeZone.swift#L192).

Expected behavior: either interpret the complete source phrase correctly or decline the conversion. A supplied source must not silently become the Mac's zone. This matters when copying a meeting time: a plausible wrong result is easy to trust.

### Steps to reproduce

1. Compile the calculator model files with a small harness that calls `CalcEngine.evaluate(query, now: now, calendar: calendar)`.
2. Set the injected `now` to `2026-09-15T12:00:00Z` and calendar zone to `Asia/Shanghai`.
3. Evaluate the queries in the table, comparing the payload and `sourceBadge`.
4. Repeat the spaced-PM query with calendar zones `UTC` and `America/Los_Angeles`.

v0.10.23 stable sources (`156972aee4da24ac4e0a992cb7333d97728551eb`); also reproduced on main at `8f7aebc1492ad84dbea3386299315a5373e4f653`.

### macOS version

26.5 (25F71), Swift 6.3.3. These are the model-harness machine's versions.

### Contribution

- [x] I'm willing to implement this myself once this issue is labelled `approved`.

caezium:
## Related issue

Closes #727

## What changed

`5:30 pm SF to London` retains both PM and the San Francisco source, and `5pm in SF to London` resolves the source connector. Only an omitted source uses the injected home zone; an unknown source or unconsumed source text returns no calculator result.

An `in`/`at` before a source place is consumed only when the target resolves to a zone. This keeps `time in sf in 4 hours` and `now in tokyo in 2h` rejected instead of answering in the Mac’s zone, while `time in 4 hours in sf` still works. The meridiem check directly checks `rest.first` and consumes it once.

The change stays in `CalcTimeZone.swift` and the existing calculator harness.

## Memory footprint

Same Debug profile and 14-query workload on unmodified main `a38e4fc` and this patch, macOS 27.0 (26A428), Xcode 27.0. RSS in MiB:

| Step | Main | Patched |
| --- | ---: | ---: |
| Idle after launch | 99.14 | 100.19 |
| Palette open | 131.88 | 134.72 |
| Feature in use (sampled peak) | 142.83 | 145.06 |
| Palette closed, five seconds settled | 142.06 | 144.67 |

Physical footprint, idle / peak / settled, as reported by `vmmap`: main **30.7 / 50.9 / 50.0 M**, patched **31.8 / 53.0 / 52.1 M**. Both builds exceed the RSS ceiling and retain allocations above their cold-launch baseline after use.

Leak-tested: the calculator model reports **zero leaks**. The full app reports the same three `com.apple.linkd.autoShortcut` / `LNDaemonApplicationInterface` connection cycles: main **416 allocations / 19,936 bytes**, patched **415 / 19,888 bytes**. The app-wide zero-leak requirement remains unmet on both builds.

## Demo

No visual layout change. All 14 real-app checks pass: compact/spaced PM, a source connector, unknown-source rejection, both reported duration-target queries and their `at` variants, supported duration-before-zone forms, canonical current time, app search, arithmetic, and unit conversion. Live clock results were checked against Python time-zone data and the relevant screenshots inspected.

## Drawbacks

Unknown or unsupported source phrases produce no calculator card instead of a guessed conversion from the home zone. The existing full-suite timing failure and app memory/leak failures remain outside this parser fix.

## Tests & validation

- **1,327 calculator assertions pass**, including 138 added assertions. The four new duration-target rejection cases failed on the prior PR implementation across three injected home zones before this correction; the SF control checks its clock and badges.
- All 15 original issue checks pass, including source/target badges and copy text. A separate 48-case review probe rejects the 18 affected contexts and preserves all 30 controls.
- Against the prior PR implementation, the 326-context corpus changes only `time in SF in 4 hours` from a home-zone answer to no result; the other 325 outputs are identical. Against current main, 274 are identical and 52 consume or reject source text as intended.
- Full suite: **72/73 harnesses pass**. The sole `custom-command-test` failure, `output arrives while the command is still running`, also reproduces on unmodified main. Lint, model-import purity, and `git diff --check` pass; no lint warnings are in changed files.
- Debug and Release builds pass with no new warnings relative to main. Release executable: **11,144,008 bytes**, versus **11,144,088 bytes** on main.
- Rebased on current main, kept one logical commit, and reviewed the complete two-file diff.


abue-ammar:
thanks, this is a nice small fix and it does solve what #727 describes.

one thing before i merge though. since `in`/`at` is now stripped from the source, a couple of queries that used to return nothing now return a wrong-looking answer:

- `time in sf in 4 hours` → shows the mac's time 4h from now instead of SF's
- `now in tokyo in 2h` → same, mac's time instead of tokyo's

the source becomes sf/tokyo, but the trailing `in 4 hours` gets read as the target, so the target falls back to the home zone. that's the same "plausible but wrong" thing the issue is about.

could you skip stripping `in`/`at` from the source when the target is a duration, and add a test for those two? `time in 4 hours in sf` should keep working as it does now.

small nit while you're there: `if let meridiem = rest.first, meridiem == "am" || meridiem == "pm"` can just be `if rest.first == "am" || rest.first == "pm"`.

whoisarpit:
One related Raycast-compatibility gap I hit: `7am ist to sgt` returns no result, while `7am ist to singapore` and `7am ist to sgp` work. Raycast accepts the `sgt` form. Would you be open to adding `sgt` to the `Asia/Singapore` aliases with a test while this timezone parser PR is open? Happy to file a separate issue if you’d prefer to keep this PR scoped to #727.

abue-ammar:
@caezium have a look at the reviews

PR #756, opened 2026-09-15
Search and run Apple Shortcuts from the launcher

### LINKED ISSUE #721: [Feature]: Search and run Apple Shortcuts from Tinycast
### What should it do?

Add support for discovering and running shortcuts saved in Apple's macOS Shortcuts app from Tinycast's launcher.

When the Tinycast palette is open, typing the name of a shortcut should fuzzy-search it alongside apps, built-in commands, custom commands, and Quicklinks. Pressing Return should run the selected shortcut.

A first version does not need to create or edit shortcuts. It only needs to:

- discover shortcuts available to the current user;
- search them by name;
- run the selected shortcut through macOS's existing Shortcuts mechanism.

The shortcut definitions, permissions, and personal data should remain owned by the Shortcuts app. Refreshing the list when the palette opens would be sufficient for an initial implementation; a background watcher or a second automation engine is not required.

Tinycast already provides one keyboard-driven launcher for apps, custom commands, and Quicklinks. Existing macOS Shortcuts are currently outside that search surface, so users must open the Shortcuts app or create a separate Quicklink for each shortcut manually.

Shortcuts are a native, user-owned part of macOS and are often used for small daily workflows. Making them searchable in Tinycast would provide one entry point for those workflows without duplicating shortcut definitions, reimplementing an automation engine, or adding a third-party sync service.

### Contribution

- [ ] I'm willing to implement this myself **once this issue is labelled `approved`**

abue-ammar:
## Related issue

Closes #721

## What changed

Shortcuts from Apple's Shortcuts app now appear in the launcher as their own **Apple Shortcuts** section, after Quicklinks. You can find them by name or alias and run them with ↵, a favorite slot or a global hotkey.

- **Discovery:** there is no public API that lists shortcuts, so `AppleShortcutRunner` runs `/usr/bin/shortcuts list --show-identifiers` through `ToolRunner`. It needs no TCC grant. The list is read again on every launcher open, when the pane appears, and when the switch turns on. The read takes about 10 ms, and an overlapping request is dropped. `AppleShortcut.parseList` reads the name from before the trailing `(UUID)`, so a name that contains parentheses still parses.
- **Identity:** the entry id is `apple-shortcut:<uuid>`, with no bundle id. An alias or hotkey follows the shortcut through a rename, and a shortcut never shares a `preferenceKey` with Shortcuts.app. Every row uses the Shortcuts app's icon.
- **Running:** `AppleShortcutCoordinator.run(id:)` is the single funnel for launcher rows and hotkeys. It hides the palette, hands focus back, and runs `shortcuts run <uuid>` with **no timeout**, because a shortcut can wait on its own dialog. Success is silent. A failure shows Tinycast's dialog with the tool's error.
- **Settings → Apple Shortcuts:** the feature ships **off**, with one switch. When it is on, the pane lists each shortcut with icon, alias, hotkey recorder and a hide checkbox.
- **Sweeping deleted shortcuts:** a read that differs from the last one sweeps, and so does the first read after launch. It removes the hotkey, alias, visibility, favorite and ranking of any shortcut the list no longer names. A failed read, an empty read, or turning the feature off frees nothing.
- **Kind wiring:** `AppEntry.Kind.appleShortcut` and `HotKeyAction.appleShortcut(id:)`, with a `boundAppleShortcutIDs` index. `appleShortcutsEnabled` is saved in settings backups.

Also in the diff:

- The table half of `LauncherItemsSection` is extracted into `LauncherItemsList`, so the new pane reuses it. The four existing launcher-category panes behave as before.
- `ToolRunner.run` accepts a `nil` timeout. Both existing callers keep 120 s.

## Memory footprint

Not measured. The feature keeps one array of names and UUIDs and one icon bitmap shared by every row. The tool is a short-lived child process, and with the switch off it never runs.

Leak-tested: no. The coordinator lives as long as the app and holds no observers or timers.

## Demo

None yet.

## Drawbacks

- An empty read never sweeps, because deleting every shortcut looks the same as a read that came back empty. After you really delete the last shortcut, its aliases and hotkeys stay stored until the library holds a shortcut again.
- Hotkey bindings for Apple Shortcuts are not in `HotkeyBackup`, the same as extension commands.
- Discovery depends on the output format of `shortcuts list`, which Apple doesn't document.

## Tests & validation

- `./Scripts/run-tests.sh`: all 73 harnesses passed before the empty-read guard was added. After it, only `apple-shortcut-test` was rerun, and it passes. The new harness covers the list parser (parentheses in names, stderr noise, duplicate UUIDs, CRLF), entry id round trips, and `staleIDs`, including the empty-read guard.
- `./Scripts/lint.sh` is clean, and the Debug build succeeds with no new warnings.
- Not exercised in the running app yet.


Wulgaren:
can we pass value to those shortcuts?

PR #858, opened 2026-09-17
Reuse the cached source when resizing row icons

### LINKED ISSUE #568: Reduce launcher icon cache memory while preserving Retina resolution
Would you be interested in a small PR to reduce cached launcher icon bitmaps while keeping Retina resolution?

The current rasterizer creates 96×96px bitmaps, while launcher rows display icons at 24pt. A 48×48px bitmap already covers 24pt at 2×. My local experiment requests 48px for these small rows, keeps 96px for larger consumers, and includes pixel size in the cache key. This reduces raw per-icon bitmap storage by 75%; it targets runtime memory, not package size.

In three fresh, isolated hosts using the real Debug library and 500 warmed synthetic icons, median physical footprint with the palette open fell from 78.03 MiB to 61.30 MiB (about 16.7 MiB). These are synthetic measurements, not a whole-app memory-budget claim. Cache-size separation tests and launcher navigation/light-mode checks passed; broad real-app icon visual comparison and physical 1× coverage remain outstanding.

The current experiment hardcodes 48px for views no larger than 24pt. A PR could make the intended size/scale explicit so future larger views cannot accidentally reuse undersized bitmaps. No artwork replacement is needed. [Apple's Retina sizing guidance](https://developer.apple.com/library/archive/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Optimizing/Optimizing.html) describes the point/pixel distinction.

Would this optimization fit the project, and would you like me to prepare a focused PR for you to try?

caolonghao:
The row icon path is documented and tested to resize the cached 96px icon, but it fetched a second dynamic `NSWorkspace` image instead. Under parallel AppKit load and appearance changes, the second fetch can resolve to different bytes, making the final-size rendering inconsistent and causing `icon-cache-test` to fail intermittently.

Reuse the full-size cached source for row resizing. This also avoids a redundant workspace icon decode and makes the implementation match the existing quality invariant.

Closes #568

Validation:

- `./Scripts/run-tests.sh icon-cache-test` passed 5 consecutive runs
- `./Scripts/run-tests.sh`: all 73 harnesses passed in parallel
- `./Scripts/lint.sh`: `lint-clean` (existing warnings only)
- `xcodebuild build -project Tinycast.xcodeproj -scheme Tinycast -configuration Debug CODE_SIGNING_ALLOWED=NO -quiet`


abue-ammar:
@caolonghao i can not just simpley trust what your pr description says. and it  undoes #617 's memory saving. show me actual regression matrics that actually proves

PR #895, opened 2026-09-18
Show calendar colour, time range and countdown on meeting rows

### LINKED ISSUE #881: [Feature]: Show calendar events date/time for meetings/events
### What should it do?

Right now, it shows like this 

Image

It's fine, but I want to see the time of the event and for the first 3 upcoming(if they are in the next hour) - show a label and calculate the minutes, like: in 25 minutes.

The functionality is mostly there.

### Contribution

- [x] I'm willing to implement this myself **once this issue is labelled `approved`**

abue-ammar:
## Related issue

Closes #881

## What changed

Meeting rows in the launcher and in My Schedule now show when each event happens and which calendar it belongs to.

- **Calendar colour.** A thin bar in the calendar's own colour sits between the row icon and the title. `CalendarStore` converts `EKCalendar.cgColor` to sRGB and stores it as `MeetingEvent.CalendarColor`, so `Model/` stays free of AppKit.
- **Time range.** The row ends with the start and end time, for example `12:39 AM – 1:09 AM`. The two times are formatted separately, not with the interval style, for two reasons: the interval style drops a shared AM/PM, and it prints full dates for an event that crosses midnight. The 12/24-hour format follows the system setting.
- **Countdown pill.** `UpcomingWindow.rowCountdown` returns `Now` for a meeting under way and `in N min` for one starting within 60 minutes, and returns nothing otherwise. It reuses the existing `countdown` wording, so the card and the rows cannot drift apart. The pill sits in a slot at the far right that is always reserved, so pills form one straight column.
- **Join card.** The subtitle shows the range and a calendar-colour dot (`ColorDot`).
- **Launcher cap.** "Upcoming meetings in launcher" now defaults to **5 next** instead of 3. A value the user saved is kept.

Non-obvious: the bar and the pill on a launcher row are resolved live from `CalendarStore` and `MeetingClock` through `MeetingEntryContent`. They are not baked into `AppEntry`, so a countdown that changes every minute never republishes the search index. New tokens: `colorDot`, `calendarBarWidth`, `calendarBarHeight`.

## Memory footprint

Not measured for this change. It adds one small struct per event and no caches or long-lived objects.

| Idle after launch       | —      |
| Palette open            | —      |
| Feature in use (peak)   | —      |
| Palette closed, settled | —      |

Leak-tested: no.

## Demo

No video recorded. Tested by hand against real events in two calendars of different colours: events in progress, events inside and outside the hour, an event exactly 60 minutes away, events that cross midnight, and tomorrow's events. Checked in both 12-hour and 24-hour time.

## Drawbacks

- Rows with no countdown keep the pill slot empty, so all times end at the same x. This is on purpose, to keep the pills in a straight column.
- Launcher meeting rows no longer show the grey "Meeting" kind label. The timing takes that place, and the rows still sit under the **Meetings** section header.

## Tests & validation

- `./Scripts/run-tests.sh`: all 76 harnesses pass.
- `calendar-test`: new `rowCountdowns` cases cover exactly an hour out, one second past the hour, inside the hour, the start, under way, and finished.
- `interface-size-test`: pins the three new size tokens.
- `./Scripts/lint.sh` is clean, the Debug build has no new warnings, and the `Model/` import check is clean.


hackmajoris:
👏

PR #915, opened 2026-09-19
Record unmarked extension Clipboard.copy in history

### LINKED ISSUE #885: [Bug]: Extension Clipboard.copy never enters clipboard history
### Existing issues

### What happened?

An extension can `Clipboard.copy("…")` successfully — the text is on the system pasteboard and pastes into other apps — but Tinycast clipboard history never records it.

Raycast records a normal `Clipboard.copy` in history unless `{ concealed: true }` is passed. Tinycast's host does not: every extension text copy is stamped `com.tinycast.internal`, and the poller skips that marker.

Noticed with a local no-view command (`Clipboard.copy(generated.fullAddress)` with no concealed option). Same path is `Action.CopyToClipboard`.

**Expected:** a non-concealed `Clipboard.copy` appears in clipboard history, matching [Raycast's Clipboard.copy](https://developers.raycast.com/api-reference/clipboard).

**Actual:** the pasteboard updates, history does not.

**Cause:** `ExtensionHostBridge.clipboard` calls `Paster.copyString` for `copy`:

```swift
if method == "copy" {
  Paster.copyString(text)
}
```

`copyString` writes through `writeString`, which stamps `ClipboardManager.internalType`. The poller returns early on that type.

`Paster` already has the unmarked write this should use:

```swift
/// Put a string on the pasteboard unmarked, so it enters history like any other copy.
static func copyPlainText(_ text: String)
```

The host also never reads `arguments[1]`, so `concealed` is ignored. Today that happens to keep secrets out of history only because *every* extension copy is marked internal.

**Suggested fix:**

- `Clipboard.copy` without `concealed` → `Paster.copyPlainText`, so the poller records it
- `Clipboard.copy(..., { concealed: true })` → stamp `org.nspasteboard.ConcealedType` (and keep it out of history)

File copies from the same host already go through `writeFileToPasteboard` without the internal marker, so they can enter history while text copies cannot.

### Steps to reproduce

1. Enable extensions and install any command that copies plain text (a no-view `Clipboard.copy("hello")` is enough; I used a local DuckDuckGo Email "Generate and Copy Duck Address" command).
2. Copy something else first so history has a known latest item.
3. Run the command. Confirm the new text pastes in TextEdit / Notes.
4. Open Tinycast clipboard history.

The new text is not there. A manual ⌘C of the same string from TextEdit does appear.

### Screenshots or recording

_No response_

0.11.2-beta.97 Beta

### macOS version

27.0 (build 26A428)

### Contribution

- [x] I'm willing to implement this myself **once this issue is labelled `approved`**

FreeTeaspoon:
## Related issue

Closes #885

## What changed

`Clipboard.copy` from an extension wrote through `Paster.copyString`, which stamps `com.tinycast.internal`. The poller skips that marker, so the text landed on the system pasteboard and never in clipboard history.

The host now uses the unmarked write `Paster.copyPlainText` already existed for, so a normal copy is recorded like any other. `{ concealed: true }` (and the older `transient` option) stamps `org.nspasteboard.ConcealedType` instead, which the poller already treats as a secret. File copies take the same concealed flag.

Only `ExtensionHostBridge.swift` changes.

## Memory footprint

Not measured. A host dispatch change: no new state, observers or allocations beyond the copied string.

| Step | Memory |
| Idle after launch | n/a |
| Palette open | n/a |
| Feature in use (peak) | n/a |
| Palette closed, settled | n/a |

Leak-tested: no. No new owners, tasks or observers.

## Demo

Non-visual. Confirmed against `main` that `Paster.copyPlainText` writes only `public.utf8-plain-text`, and that a ConcealedType write matches the poller's existing skip set.

## Tests & validation

- `./Scripts/run-tests.sh`: 74 of 76 harnesses pass, including `ext-test`, `pasteboard-test` and `clipboard-test`. `custom-command-test` failed on "commands start in the user's home directory" and `codex-turn-test` failed talking to the stub app-server — neither compiles `ExtensionHostBridge`.
- `./Scripts/lint.sh` was not run: `swiftlint` is not installed in this environment.
- Debug `xcodebuild` did not complete here: this machine has no `Tinycast Self-Signed` identity.
- `grep -rln 'import AppKit\|import SwiftUI\|import Cocoa' Tinycast/Features/*/Model/` is empty.
- Docs were not touched; nothing in `docs/` stated that extension copies were skipped.

Made with [Cursor](https://cursor.com)


abue-ammar:
image
remove ai co authorship

PR #964, opened 2026-09-20
Open the palette without re-rendering it first

abue-ammar:
## Related issue

None.

## What changed

Opening the palette used to re-render the whole palette before it could show anything.

Closing it already resets it: `hide` schedules Pop to Root Search, which on the default
`immediately` setting resets the screen right there, while the window is off screen. Opening it then
called `prepare(mode:)` on that same fresh screen. `prepare` always bumps `focusToken` and
`resetToken`, `RootPaletteView` observes both, and each bump cascades — the `onChange` handlers write
`scroll` and `searchFocused`, which invalidate the body again. Measured with the SwiftUI instrument:
**four to five full evaluations of `RootPaletteView` per open**, each one rebuilding the launcher row
list, all of it in front of the first frame.

- `PaletteWindowController` records `isPoppedToRoot` when a pop to root resets the hidden palette, and
  spends it on the next `show()`.
- `PaletteCoordinator.showPalette` skips `navigate(to:)` when the hidden palette already holds that
  mode's fresh root. A seeded query, a different mode and a preserved screen all behave as before.
- The `vm.isVisible` observer moves into a small `ViewModifier`, so `noteVisible(true)` invalidates
  that modifier instead of the root view's body.

After this, an open evaluates `RootPaletteView` **once**.

## Memory footprint

No memory change is expected or claimed: this moves no allocation, it removes renders. The benchmark
below sampled the process group once a second for a minute after 50 opens, with the palette closed.

| Step | Memory |
| --- | --- |
| Idle after launch | not measured separately |
| Palette open | not measured separately |
| Feature in use (peak) | not measured separately |
| Palette closed, settled | 56.6 MiB footprint, 71.1 MiB RSS (median of 60 samples) |

Leak-tested: no. Nothing here changes ownership — one stored `Bool` on the controller, and an observer
that moves from one view body to another.

## Demo

NEW:
latency

memory

OLD: 

image
image

Nothing visual changed. Numbers instead, from Instruments, warm opens on an M3 Pro:

| | `main` | This branch |
| --- | ---: | ---: |
| `PaletteWindowController.show` | 34.5 ms | 7.6 ms |
| Shortcut → first CA commit | 49.0 ms | 20.8 ms |
| `RootPaletteView` body evaluations per open | 4–5 | 1 |

Medians of 15 opens each, `os_signpost` + Animation Hitches, same session. Instruments' own overhead
is in both columns.

End to end, against Raycast 2.4.1 through
[mortenjust's published protocol](https://gist.github.com/mortenjust/a0281894df2fdb8d006f6839bd8d23bc)
— 50 alternating opens per app, shortcut to the first frame the window server presents:

| | Tinycast (this branch) | Raycast 2.4.1 |
| --- | ---: | ---: |
| Median | 52.5 ms | **41.7 ms** |
| 95th percentile | 61.3 ms | **49.9 ms** |
| Memory footprint | **56.6 MiB** | 542.1 MiB |
| Resident memory | **71.1 MiB** | 575.5 MiB |

Raycast still wins the opening race by 10.8 ms at the median (paired-bootstrap 95% interval 6.4 to
12.6 ms; it was ahead in 46 of 50 pairs). Tinycast uses 9.6× less memory.

## Drawbacks

- **This does not beat Raycast, it closes about two thirds of the gap.** What is left is mostly one
  cost I could not remove: when the panel becomes the key window, AppKit hands SwiftUI a changed
  environment and every row's text is resolved and measured again, about 7 ms. I reproduced that in a
  40-row standalone panel (1.6 ms per open with no key change, 8.3 ms with one) and pinning
  `appearsActive` — or the deprecated `controlActiveState` — changed nothing. Removing it means
  reworking how the palette takes key, which is not this PR.
- Roughly 2 ms per open goes to LaunchServices, resolving the frontmost app and the paste target.
  Both are wanted before the first frame today, so I left them alone.
- `isPoppedToRoot` is a second piece of "what state is the hidden palette in", next to
  `consumePreservedState`. It is spent in `show()`, and a summon of a *different* mode still resets,
  so the two cannot both claim an open — but it is one more flag to keep honest.
- The benchmark above is one session on one Mac, on a 120 Hz built-in display rather than the 60 Hz
  one the protocol's author used. It measures warm opening only: not cold launch, not readiness to
  type. One Raycast reading of 5.0 ms is almost certainly a false detection by the pixel threshold;
  the median does not depend on it.

## Tests & validation

- `./Scripts/run-tests.sh` — 76 harnesses pass, including `palette-navigation-test`.
- `./Scripts/lint.sh` — clean.
- Release build, no new warnings.
- By hand on the Release build: open and close repeatedly with ⌘Space; open, type, Escape, reopen —
  the field is empty and focused, the list is at the top; open the clipboard hotkey over a hidden
  launcher and confirm it still resets; Escape from a pushed screen, reopen, confirm the root search;
  `Pop to Root Search` set to 90 seconds, reopen inside the window and confirm the query survives and
  is selected.
- Not run: the same end-to-end benchmark against `main`, so the two rows in the second table come from
  Instruments, not from that protocol.

## Reproducing the benchmark

One command, from a gist rather than this repo:
https://gist.github.com/abue-ammar/e84d75b2c3a4d389c6efb213bea1b1eb — it has the script and the full
setup. It fetches the runner
from the gist, builds Release under the shipping name (the runner matches the window owner `Tinycast`
exactly), preflights, runs the pilot, waits for you to confirm the four pilot screenshots, then
measures and opens the charts. Results land in `build/launcher-benchmark/runs/<timestamp>/`.

It needs: Raycast running on ⌥Space, Screen Recording and Accessibility for the terminal, Tinycast
Beta and Dev quit (the memory tool counts them as Tinycast), and about five minutes of an untouched
machine. The ⌘Space binding it writes to the `com.tinycast.app` domain is restored when it exits.


abue-ammar:
@kmplngj hey can you review this up and lemme know your feedback. if this has any bug or confirmed issue or break any existing feature

list-jonas (commented):
Didn't find any issues

Issue #962, opened 2026-09-20
[Bug]: Error: tls.TLSSocket is not supported in Tinycast extensions (no Node runtime). See docs/extensions.md.


durdevic:
### Existing issues

- [ ] I searched open and closed issues and didn't find this already reported

### What happened?

Is there a way to enable this. 

I tried using this extension

https://github.com/raycast/extensions/blob/3c654737b0d566d3103fcdf72221a9f34664bdf2/extensions/ip-geolocation/README.md

### Steps to reproduce

1. Install from Raycast store
2. try to use it 

### Tinycast version + channel

10.10.23 (93)

### macOS version

26.6.2

### Contribution

- [ ] I'm willing to implement this myself **once this issue is labelled `approved`**

abue-ammar:
this issue is partially solved, and to completely fix will open backdoor security issue

durdevic:
So how can I use it, or it's simply not usable at all?

abue-ammar:
> So how can I use it, or it's simply not usable at all?

it will partially work not fully you can not expect every raycast extention to work here as the disclaimer says already

<img width="817" height="467" alt="Image" src="https://github.com/user-attachments/assets/34abd915-9cea-4857-916f-33478dbeb30b" />

durdevic:
Yep, saw that, but is there a way to convert that IP extension to tinycast or no way at all, cuz I didn't find anything similar

PR #967, opened 2026-09-20
Run got-backed extensions by constructing tls.TLSSocket

list-jonas:
## Related issue

Partial fix for #962 — IP Geolocation fails with `tls.TLSSocket is not supported in Tinycast
extensions` before any command renders. Its data still comes back `undefined` after this; that's a
separate failure and I'll open its own PR.

## What changed

`http2-wrapper`, inside `got`, derives a class at import time from
`new tls.TLSSocket(stream)._handle._parentWrap.constructor`, so the bundle threw before the
extension ran a line. `TLSSocket` now constructs as an inert `Duplex` carrying that shape; the rest
of `tls` still throws.

## Tests & validation

All three IP Geolocation commands render through `node Scripts/raycast-runtime/test.mjs`. All 76
harnesses pass, lint is clean, Debug build has no new warnings.


SnaetWarre (approved):
Looks nice! thanks for also updating the right documentation!

PR #968, opened 2026-09-20
Let extensions reach cleartext endpoints

list-jonas:
## Related issue

The other half of #962. With #967 in place the extension loads, but every value still renders
`undefined`: its geolocation comes from `http://ip-api.com/json/`, and ATS fails that request with
`NSURLErrorAppTransportSecurityRequiresSecureConnection` (-1022) before it leaves the app.

## What changed

Extensions name their own endpoints and plenty are cleartext — ip-api.com, a Home Assistant on the
LAN, a localhost dev server. ATS is a process-wide switch with no per-session opt-out, so
`NSAppTransportSecurity.NSAllowsArbitraryLoads` in `Info.plist` is the only lever that reaches
`ExtensionFetcher`'s session. Tinycast's own features each name an https URL, which this cannot
downgrade.

## Tests & validation

IP Geolocation's My IP Geolocation command renders its full result — public IPv4, location,
coordinates, timezone, AS, ISP and organization — in a Debug build carrying both changes, where it
rendered `undefined` without this one. Verified the mechanism separately: the same URLSession
request fails with -1022 from a bundle without the key and returns 200 with it. All 76 harnesses
pass, lint is clean, and the Debug build has no new warnings.


list-jonas:
@abue-ammar not sure if you want to merge this since it touches Info.plist

Doesnt need any perms but idk about App Store review

abue-ammar:
@list-jonas  this is more of a giving aceess to the entire world have access to tinycast. more of a backdoor open and has major risk. is the other pr linked to it?

